PR #12, opened 2022-06-22
Enable running on already deployed Authorino

pehala:
* The operator test will be skipped if that's the case.
* You can set it up by settings `authorino.deploy` to false
   * You also need to specify the authorization URL for it through `authorino.url`
   * Details can be found in `settings.local.yamp.tpl`

### DIFF
--- a/testsuite/objects/__init__.py
+++ b/testsuite/objects/__init__.py
@@ -35,3 +35,24 @@ class Authorization(LifecycleObject):
     @abc.abstractmethod
     def add_oidc_identity(self, name, endpoint):
         """Adds OIDC identity provider"""
+
+
+class PreexistingAuthorino(Authorino):
+    """Authorino which is already deployed prior to the testrun"""
+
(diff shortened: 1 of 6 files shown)


testsuite/objects/__init__.py
```diff
@@ -35,3 +35,24 @@ class Authorization(LifecycleObject):
...
+
+class PreexistingAuthorino(Authorino):
+    """Authorino which is already deployed prior to the testrun"""
+
+    def __init__(self, authorization_url) -> None:
+        super().__init__()
+        self._authorization_url = authorization_url
+
+    def wait_for_ready(self):
+        return True
```
guicassolato, line 48: I suppose we don't care if the preexisting Authorino instance is NOT ready, right? Tests will fail of course, but that's all right. Any preexisting Authorino instance is supposed to be ready before using this option, I imagine.

dhlavac (approved):

PR #28, opened 2022-07-25
Add basic test for authorino sharding

### LINKED ISSUE #16: Add tests for sharding
Authorino can be set up to watch only some namespaces and/or labels. We should test that it really watches only the specified resources and what happens when multiple Authorinos watch the same resource.

https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/sharding.md

JaurbanRH:
Fixes #16

### DIFF
--- a/testsuite/openshift/objects/api_key.py
+++ b/testsuite/openshift/objects/api_key.py
@@ -8,7 +8,7 @@ class APIKey(OpenShiftObject):
     """Represents API Key Secret CR for Authorino"""
 
     @classmethod
-    def create_instance(cls, openshift: OpenShiftClient, name, label, api_key_string):
+    def create_instance(cls, openshift: OpenShiftClient, name, label, api_key):
         """Creates base instance"""
         model = {
             "apiVersion": "v1",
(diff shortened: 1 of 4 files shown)


testsuite/tests/kuadrant/authorino/sharding/test_sharding.py
```diff
@@ -0,0 +1,75 @@
...
+from testsuite.openshift.objects.authorino import AuthorinoCR
+
+
+@pytest.fixture(scope="session")
+def authorino(openshift, blame, request, testconfig, label) -> AuthorinoCR:
+    """Custom deployed Authorino instance"""
+    if not testconfig["authorino"]["deploy"]:
+        return pytest.skip("Operator tests don't work with already deployed Authorino")
+
+    authorino = AuthorinoCR.create_instance(openshift,
```
pehala, line 16: I would wait until #19 is merged and then used `authorino_parameters` fixture it includes. I would also migrate this test to operator dir as only there should we modify Authorino creation in any way

PR #29, opened 2022-08-02
Add tests for multiple hosts

### LINKED ISSUE #24: Add test for multiple hosts
AuthConfig can have multiple hosts, we should test that all of them work and if one of them is invalid (e.g. collision) that others still work as expected. https://github.com/Kuadrant/authorino/issues/323

pehala:
Simple test for AuthConfig with two hosts

Closes #24

### DIFF
--- a/testsuite/objects/__init__.py
+++ b/testsuite/objects/__init__.py
@@ -36,6 +36,18 @@ class Authorization(LifecycleObject):
     def add_oidc_identity(self, name, endpoint):
         """Adds OIDC identity provider"""
 
+    @abc.abstractmethod
+    def add_host(self, hostname):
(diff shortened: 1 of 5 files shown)


jsmolar (approved):

dhlavac:
@pehala shouldn't there be also a test that removes the first host/create invalid host and try that the second host still works?

pehala:
Regarding the removal of first host and reconciliation that is a good test case but I would implement it in subsequent PR.

As for the invalid hosts, the authorino doesn't care about host validity, it only uses it as a key for a map basically, so there is no need to test that.

dhlavac (approved):

PR #43, opened 2022-08-09
Add Clusterwide support

pehala:
Enable running Authorino in cluster-wide mode and add a simple test for it

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -18,10 +18,26 @@ def testconfig():
 
 @pytest.fixture(scope="session")
 def openshift(testconfig):
-    """Returns OpenShift client builder"""
-    client = OpenShiftClient(weakget(testconfig)["openshift"]["project"] % None)
+    """OpenShift client for the primary namespace"""
+    section = weakget(testconfig)["openshift"]
+    client = OpenShiftClient(
+        section["project"] % None,
+        section["api_url"] % None,
+        section["token"] % None)
     if not client.connected:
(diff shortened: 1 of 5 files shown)


testsuite/tests/conftest.py
```diff
@@ -18,13 +18,29 @@ def testconfig():
...
+        section["project"] % None,
+        section["api_url"] % None,
+        section["token"] % None)
     if not client.connected:
         pytest.fail("You are not logged into Openshift or the namespace doesn't exist")
     return client
 
 
+@pytest.fixture(scope="session")
+def openshift2(openshift, testconfig):
```
jsmolar, line 33: I wouldn't say I like creating a second openshift client fixture in the primary contest for some subset of tests. In addition, this is a fixed solution for 2 projects, is there a possibility, that we would need to work with 3 or more projects?
My humble suggestion is that we deal with openshift clients like login fixtures in 3scale. One parametrized (custom_openshift) fixture, that by default creates openshift with values from testconfig which will be called by openshift fixture with default values. This will provide modularity to create additional clients.

pehala: I do not currently see the possibility for 3 or more OpenShifts. The reason why I choose hardcoded instead of custom fixture is that we have requirement for relations between those clients, they need to be on the same server, meaning we cannot allow full customization of a client (e.g. api_url, token).

Second reason is that I would like to keep a static configuration with static amount fo sections, having custom_openshift, which would take values from config would mean a dynamic amount of sections, which I do not like.

jsmolar (approved):

JaurbanRH (approved):

PR #46, opened 2022-08-09
Add module_label

pehala:
Previously we had only one label, which lasted the entire test run, meaning different modules could share resources. This eliminates it.

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -109,6 +109,12 @@ def label(blame):
     return blame("testrun")
 
 
+@pytest.fixture(scope="module")
+def module_label(label):
+    """Module scope label for all resources"""
+    return randomize(label)
+
+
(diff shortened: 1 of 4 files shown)


testsuite/tests/conftest.py
```diff
@@ -93,6 +93,12 @@ def label(blame):
     return blame("testrun")
 
 
+@pytest.fixture(scope="module")
+def module_label(label):
```
dhlavac, line 97: Shouldn't be all labels randomized not only module scope?

pehala: Well, thats what label fixture is only for session scope. I did not yet add the change to every test that uses it but all the stuff deployed should use either module_label or label

jsmolar (approved):

JaurbanRH (dismissed):

dhlavac (approved):

PR #54, opened 2022-08-23
Change how are OpenShifts processed in the settings

pehala:
Change how the OpenShift is loaded in testsuite, now all the processing (e.g. converting it to OpenShiftClient) occurs in the new dynaconf loader and the respective sections in the testconfig are already processed OpenShiftClients.

### Breaking change
Config structure regarding the second OpenShift project was changed.

### DIFF
--- a/config/settings.local.yaml.tpl
+++ b/config/settings.local.yaml.tpl
@@ -1,8 +1,11 @@
 #default:
 #  skip_cleanup: false
 #  openshift:
-#    project: "authorino"
-#    second_project: "authorino2"
+#    project: "kuadrant"                       # Optional: namespace for tests to run, if None uses current project
+#    api_url: "https://api.openshift.com"      # Optional: OpenShift API URL, if None it will OpenShift that you are logged in
+#    token: "KUADRANT_RULEZ"                   # Optional: OpenShift Token, if None it will OpenShift that you are logged in
+#  openshift2:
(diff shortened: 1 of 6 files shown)


JaurbanRH (approved):

config/settings.local.yaml.tpl
```diff
@@ -1,8 +1,11 @@
 #default:
 #  skip_cleanup: false
 #  openshift:
-#    project: "authorino"
-#    second_project: "authorino2"
+#    project: "kuadrant"                       # Optional: namespace for tests to run, if None uses current project
+#    api_url: "https://api.openshift.com"      # Optional: API URL for OpenShift, if None it will OpenShift that you are logged in
+#    token: "KUADRANT_RULEZ"                   # Optional: API URL for OpenShift, if None it will OpenShift that you are logged in
```
dhlavac, line 6: IMHO wrong comment of token

dhlavac (changes requested):

config/settings.yaml
```diff
@@ -1,10 +1,9 @@
 default:
   skip_cleanup: false
   dynaconf_merge: true
-  openshift:
-    project: "authorino"
   cfssl: "cfssl"
   rhsso:
+    username: "admin"
```
dhlavac, line 6: where is used ["rhsso"]["username"] ?

dhlavac (changes requested):

dhlavac (approved):

PR #70, opened 2022-08-30
Separate envoy into its separate module

pehala:
* All envoy classes are not in one envoy module, instead of httpbin one
* Change naming of envoy instances, previously there was envoy mentioned twice
* Fix route() property of envoy to accurately select route even if it was called after another route was requested


dhlavac (approved):

dhlavac:
@pehala pls rebase

dhlavac (approved):

JaurbanRH (approved):

PR #71, opened 2022-08-30
Add tools dynaconf loader

### LINKED ISSUE #61: Enable fetching configuration from OpenShift
We could fetch RHSSO info and other tools which we might need in the future from OpenShift.

pehala:
- Lazily loads required info from OpenShift
- For now only loads RHSSO details, if present

Fixes #61


pehala:
Closed in favour of #78

PR #74, opened 2022-08-31
Rework RHSSO into a more general OIDCProvider interface

### LINKED ISSUE #68: Rethink RHSSO object structure
Currently we used objects migrated from 3scale testsuite with slight modifications, they are not ideal and contain a lot of leftover stuff, it would be good idea to rethink them and design a better interface.

pehala:
* Introduce `OIDCProvider` interface
   * This enables us to quickly swap between OIDCProviders (planned Auth0 and RHSSO for now) without too much of a hassle
   * Add Token class which wraps the Token and can refresh itself if needed
   * Add `oidc_provider` fixture, which should replace `rhsso_service_info` almost everywhere
* Cleanup of RHSSO interface
  * Remove unneeded methods
  * Remove `RHSSOAdmin` class
  * Move the initialization from fixture to `RHSSOServiceConfiguration` class
  * Move it into the new `oidc` module

Fix #68


testsuite/oidc/rhsso/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+from keycloak import KeycloakOpenID, KeycloakAdmin
+
+from testsuite.oidc import OIDCProvider, Token
+from testsuite.objects import LifecycleObject
+from .objects import Realm, Client
+
+
+# pylint: disable=too-many-instance-attributes
+class RHSSOServiceConfiguration(OIDCProvider, LifecycleObject):
```
jsmolar, line 13: Maybe keep the naming with `Auth0Provider`? 
Also reformat doc string to be one-liner

testsuite/oidc/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def refresh(self) -> "Token":
+        """Refreshes token"""
+        return self.refresh_function(self.refresh_token)
+
+    def __str__(self) -> str:
+        return self.access_token
+
+
+class OIDCProvider(ABC):
```
jsmolar, line 22: Add abstract methods `refresh_token` and `get_token`

jsmolar (approved):

Issue #9, opened 2022-06-15
Add tests for raw-HTTP authorization


pehala:
https://github.com/Kuadrant/authorino-operator/pull/61
https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/validating-webhook.md

PR #77, opened 2022-09-02
Add raw http test with response with adding extra header

mkudlej:
closes https://github.com/Kuadrant/testsuite/issues/9

### DIFF
--- a/testsuite/tests/kuadrant/authorino/operator/http/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/operator/http/conftest.py
@@ -7,7 +7,7 @@ from testsuite.httpx import HttpxBackoffClient
 
 # pylint: disable=unused-argument
 @pytest.fixture(scope="module")
-def authorization(authorization, wildcard_domain, blame, openshift, rhsso_service_info, module_label) -> Authorization:
+def authorization(authorization, wildcard_domain, openshift, rhsso_service_info, module_label) -> Authorization:
     """In case of Authorino, AuthConfig used for authorization"""
     authorization.remove_all_hosts()
     authorization.add_host(wildcard_domain)
@@ -21,7 +21,7 @@ def authorization(authorization, wildcard_domain, blame, openshift, rhsso_servic
 
 
 @pytest.fixture(scope="module")
-def client_http_auth(authorization, authorino_route):
+def client(authorization, authorino_route):
     """Returns httpx client to be used for requests, it also commits AuthConfig"""
     client = HttpxBackoffClient(base_url=f"http://{authorino_route.model.spec.host}", verify=False)
     yield client
@@ -31,8 +31,7 @@ def client_http_auth(authorization, authorino_route):
 @pytest.fixture(scope="module")
 def authorino_route(authorino, blame, openshift):
     """Add route for authorino http port to be able to access it."""
-    name = f"route-{authorino.name()}"
-    route = openshift.routes.expose(name, f"{authorino.name()}-authorino-authorization",
+    route = openshift.routes.expose(blame('route'), f"{authorino.name()}-authorino-authorization",
                                     port='http')
     yield route
     route.delete()
--- a/testsuite/tests/kuadrant/authorino/operator/http/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/operator/http/conftest.py
@@ -1,0 +1,38 @@
+"""Conftest for all tests requiring custom deployment of Authorino"""
+import pytest
+
+from testsuite.objects import Authorization
+from testsuite.httpx import HttpxBackoffClient
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(authorization, wildcard_domain, blame, openshift, rhsso_service_info, module_label) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+    authorization.remove_all_hosts()
+    authorization.add_host(wildcard_domain)
+    resp = {'name': 'another-json-returned-in-a-header',
+            'wrapperKey': 'x-ext-auth-other-json',
+            'json': {'properties': [
+                {'name': 'propX', 'value': 'valueX'}
+            ]}}
+    authorization.add_response(response=resp)
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def client_http_auth(authorization, authorino_route):
+    """Returns httpx client to be used for requests, it also commits AuthConfig"""
+    client = HttpxBackoffClient(base_url=f"http://{authorino_route.model.spec.host}", verify=False)
+    yield client
+    client.close()
+
+
+@pytest.fixture(scope="module")
+def authorino_route(authorino, blame, openshift):
+    """Add route for authorino http port to be able to access it."""
+    name = f"route-{authorino.name()}"
+    route = openshift.routes.expose(name, f"{authorino.name()}-authorino-authorization",
+                                    port='http')
+    yield route
+    route.delete()
--- a/testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
+++ b/testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
@@ -4,17 +4,17 @@ Test raw http authorization interface.
 
 
 # pylint: disable=unused-argument
-def test_authorized_via_http(authorization, client_http_auth, auth):
+def test_authorized_via_http(authorization, client, auth):
     """Test raw http authentization with Keycloak."""
-    response = client_http_auth.request("GET", "/check", auth=auth)
+    response = client.get("/check", auth=auth)
     assert response.status_code == 200
     assert response.text == ''
     assert response.headers.get('x-ext-auth-other-json', '') == '{"propX":"valueX"}'
 
 
 # pylint: disable=unused-argument
-def test_unauthorized_via_http(authorization, client_http_auth):
+def test_unauthorized_via_http(authorization, client):
     """Test raw http authentization with unauthorized request."""
-    response = client_http_auth.request("GET", "/check")
+    response = client.get("/check")
     assert response.status_code == 401
     assert response.text == ''
--- a/testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
+++ b/testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
@@ -1,0 +1,20 @@
+"""
+Test raw http authorization interface.
+"""
+
+
+# pylint: disable=unused-argument
+def test_authorized_via_http(authorization, client_http_auth, auth):
+    """Test raw http authentization with Keycloak."""
+    response = client_http_auth.request("GET", "/check", auth=auth)
+    assert response.status_code == 200
+    assert response.text == ''
+    assert response.headers.get('x-ext-auth-other-json', '') == '{"propX":"valueX"}'
+
+
+# pylint: disable=unused-argument
+def test_unauthorized_via_http(authorization, client_http_auth):
+    """Test raw http authentization with unauthorized request."""
+    response = client_http_auth.request("GET", "/check")
+    assert response.status_code == 401
+    assert response.text == ''
--- a/testsuite/openshift/objects/auth_config.py
+++ b/testsuite/openshift/objects/auth_config.py
@@ -153,5 +153,6 @@ class AuthConfig(OpenShiftObject, Authorization):
 
     @modify
     def add_response(self, response):
-        """Add response to AuthConfig"""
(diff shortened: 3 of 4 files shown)


testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+from testsuite.httpx import HttpxBackoffClient
+
+HTTP_DEFAULT_PORT=5001
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(authorization, authorino, wildcard_domain, blame, openshift, label, rhsso_service_info) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+    #if authorization is None:
+    authorization = AuthConfig.create_instance(openshift, blame("ac"), wildcard_domain, labels={"testRun": label})
```
pehala, line 17: Use module_label instead, the label is for session scoped.

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,48 @@
+"""Conftest for all tests requiring custom deployment of Authorino"""
+import pytest
+from weakget import weakget
+
+from testsuite.openshift.objects.authorino import AuthorinoCR
+from testsuite.objects import Authorino, Authorization
+from testsuite.openshift.objects.auth_config import AuthConfig
+from testsuite.httpx import HttpxBackoffClient
+
+HTTP_DEFAULT_PORT=5001
```
pehala, line 10: I believe the services have their ports named so you could expose the route and use a name instead an actual port

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    yield client
+    client.close()
+    authorization.delete()
+
+
+@pytest.fixture(scope="module")
+def authorino_route(authorino, blame, openshift):
+    name = f"route-{authorino.name()}"
+    # expose cannot be used, see https://github.com/openshift/origin/issues/8699
+    route = openshift.do_action('create', 'route', 'edge', name,
```
pehala, line 40: I do not understand why expose won't work, because that issue seems to be closed. I believe binding it to a specific service port should work

pehala: Also, you could try using `openshift.routes.expose`

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
```diff
@@ -0,0 +1,13 @@
+"""
+Test raw http authorization interface.
+"""
+
+import pytest
+
+def test_first(authorization, client_http_auth, auth):
+    response = client_http_auth.request("GET", "/check", auth=auth, json='{}')
+    # ako sa spozna, ze komunikoval s rhsso a overil to?
```
pehala, line 9: You can add tests with invalid token and without token at all

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
```diff
@@ -0,0 +1,13 @@
+"""
+Test raw http authorization interface.
+"""
+
+import pytest
+
+def test_first(authorization, client_http_auth, auth):
```
pehala, line 7: Please change testname to something more explicit

testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
```diff
@@ -1,5 +1,4 @@
 """Conftest for all TLS-enabled tests"""
-
```
pehala, line 2: Unnecessary change

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+HTTP_DEFAULT_PORT=5001
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(authorization, authorino, wildcard_domain, blame, openshift, label, rhsso_service_info) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+    #if authorization is None:
+    authorization = AuthConfig.create_instance(openshift, blame("ac"), wildcard_domain, labels={"testRun": label})
+    authorization.add_oidc_identity("rhsso", rhsso_service_info.issuer_url())
+    authorization.add_response(inject_json={'name': 'another-json-returned-in-a-header',
```
pehala, line 19: I might be missing something, but `add_response` doesn't seem to have a parameter called `inject_json`

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+                                'json': { 'properties': [{'name': 'propX', 'value': 'valueX'}]}
+                                })
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def client_http_auth(authorization, authorino_route):
+    """Returns httpx client to be used for requests, it also commits AuthConfig"""
+    authorization.commit()
+    client = HttpxBackoffClient(base_url=f"https://{authorino_route.as_dict()['spec']['host']}", verify=False)
```
pehala, line 30: `authorino_route.model.spec.host`

pehala:
The reason, why the acceptance checks fail, is that you have added `__init__.py` to the root directory, where it shouldn't be and you are missing it in your `tests/kuadrant/authorino/operator/http` dir, where it should be

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
mkudlej: It still doesn't work, see:
`$ oc expose service --name=route-authorin-mkudlej-mkdlej-kd-bzb23 -o json authorin-mkudlej-mkdlej-kd-bzb2-authorino-authorization --target-port=5001 --port=80

error: cannot use --target-port with --generator=route/v1`
I don't know what do you mean by `Also, you could try using openshift.routes.expose`, there is no such submodule of openshift client named `routes.expose` or maybe I haven't found it.

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
```diff
@@ -0,0 +1,18 @@
...
+
+def test_authorized_via_http(authorization, client_http_auth, auth):
+    response = client_http_auth.request("GET", "/check", auth=auth, json='{}')
+    assert response.is_success
+    assert response.status_code == 200
+    assert response.text == ''
+    assert response.headers.get('x-ext-auth-other-json', '') == '{"propX":"valueX"}'
+
+def test_unauthorized_via_http(authorization, client_http_auth):
+    response = client_http_auth.request("GET", "/check", json='{}')
```
pehala, line 15: `client_http_auth.get`

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
```diff
@@ -0,0 +1,18 @@
+"""
+Test raw http authorization interface.
+"""
+
+import pytest
+
+def test_authorized_via_http(authorization, client_http_auth, auth):
+    response = client_http_auth.request("GET", "/check", auth=auth, json='{}')
+    assert response.is_success
```
pehala, line 9: Redundant since you are checking the status code anyway

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
```diff
@@ -0,0 +1,18 @@
...
+    assert response.status_code == 200
+    assert response.text == ''
+    assert response.headers.get('x-ext-auth-other-json', '') == '{"propX":"valueX"}'
+
+def test_unauthorized_via_http(authorization, client_http_auth):
+    response = client_http_auth.request("GET", "/check", json='{}')
+    assert not response.is_success
```
pehala, line 16: Redundant

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
```diff
@@ -0,0 +1,18 @@
...
+    assert response.status_code == 401
+    assert response.text == ''
```
pehala, line 18: The text is important in this one, right?

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+from testsuite.openshift.objects.authorino import AuthorinoCR
+from testsuite.objects import Authorino, Authorization
+from testsuite.openshift.objects.auth_config import AuthConfig
+from testsuite.httpx import HttpxBackoffClient
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(wildcard_domain, blame, openshift, rhsso_service_info, module_label) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+    authorization = AuthConfig.create_instance(openshift, blame("ac"), wildcard_domain, labels={"testRun": module_label})
```
pehala, line 14: There is a different solution that doesn´t require creating new `AuthConfig` and using existing one. (I came up with that only after we talked, sorry)

```
authorization.remove_all_hosts()
authorization.add_host(wildcard_domain)
```

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
mkudlej: What do you mean exactly? I think it is needed to check as much as we can. On the other hand Authorino returns some explaining text in case of not successful call. Do you think that text is missing in this case?

pehala: Sorry, the question should be `The text is NOT important in this one, right?`. Authorino uses Header for extra information about checks not response text I believe, so this check is not needed.

testsuite/tests/kuadrant/authorino/operator/http/test_raw_http.py
mkudlej: `TypeError: Client.get() got an unexpected keyword argument 'json'`

pehala: Why are you sending empty JSON in a get request?

mkudlej: See this example:
```kubectl port-forward services/authorino-authorino-authorization 3000:3000 &;PID=$!

curl -H 'Content-type: application/json' -d '{}' http://wristband.127.0.0.1.nip.io:3000/check -i
# HTTP/1.1 200 OK

kill $PID``` from https://github.com/Kuadrant/authorino-operator/pull/61

File: testsuite/tests/kuadrant/authorino/operator/http/conftest.py [hunk 14]
```diff
@@ -0,0 +1,38 @@
...
+    """Returns httpx client to be used for requests, it also commits AuthConfig"""
+    client = HttpxBackoffClient(base_url=f"http://{authorino_route.model.spec.host}", verify=False)
+    yield client
+    client.close()
+
+
+@pytest.fixture(scope="module")
+def authorino_route(authorino, blame, openshift):
+    """Add route for authorino http port to be able to access it."""
+    name = f"route-{authorino.name()}"
```
pehala, line 34: Use blame instead, this could fail if multiple people are running this test

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+            'wrapperKey': 'x-ext-auth-other-json',
+            'json': {'properties': [
+                {'name': 'propX', 'value': 'valueX'}
+            ]}}
+    authorization.add_response(response=resp)
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def client_http_auth(authorization, authorino_route):
```
pehala, line 24: Can you rename this to simply `client`?

pehala: And that's exactly why I want this to be renamed. It acts exactly as a client and provides the same functionality. The original fixture doesnt work in this context as there is no envoy here, so we can safely replace it by this one which does work and is used for the same purpose.

mkudlej: ok, if you think this is a good idea, then I rename it.

pehala (approved):

JaurbanRH (approved):

PR #85, opened 2022-09-09
Tests for OPA external registry

jsmolar:


testsuite/openshift/mockserver.py
```diff
@@ -0,0 +1,38 @@
+"""Module for Mockserver integration"""
+import json
+from urllib.parse import urljoin
+
+import requests
+
+
+class Mockserver:
```
pehala, line 8: I would move this somewhere else than openshift dir, since it doesn't have anything to do with openshift

testsuite/tests/kuadrant/authorino/authorization/opa/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+    """Rego query that allows all requests that contain `key` and `value`"""
+    def _rego(key, value):
+        return f"allow {{ input.context.request.http.headers.{key} == \"{value}\" }}"
+    return _rego
+
+
+@pytest.fixture(scope="module")
+def mockserver(request, testconfig, module_label, rego_query, header):
+    """Returns mockserver and creates Expectation that returns Rego query"""
+    mockserver = Mockserver(testconfig["mockserver"]["url"])
```
pehala, line 24: Please skip tests if mockserver is not configured

testsuite/openshift/objects/auth_config.py
```diff
@@ -151,8 +151,30 @@ def add_opa_policy(self, name, rego_policy):
             }
         })
 
+    @modify
+    def add_external_registry(self, name, endpoint, ttl=0):
```
pehala, line 155: `add_external_opa` would probably be better

testsuite/tests/kuadrant/authorino/authorization/opa/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+def header():
+    """Header used by OPA policy"""
+    return "opa", "opa-test"
+
+
+@pytest.fixture(scope="module")
+def rego_query():
+    """Rego query that allows all requests that contain `key` and `value`"""
+    def _rego(key, value):
+        return f"allow {{ input.context.request.http.headers.{key} == \"{value}\" }}"
```
pehala, line 17: I know it is code repetition but I would probably use this explicitly instead so its more explicit when reading the test, or alternatively move this to utils and give it a better name

testsuite/openshift/objects/auth_config.py
```diff
@@ -151,8 +151,30 @@ def add_opa_policy(self, name, rego_policy):
...
+            "opa": {
+                "externalRegistry": {
+                    "endpoint": endpoint,
+                    "ttl": ttl
+                }
+            }
+        })
+
+    @modify
+    def remove_all_authorization(self):
```
pehala, line 171: Where is this used?

testsuite/openshift/objects/auth_config.py
pehala, line 155: Add this to the interface as well

testsuite/openshift/objects/auth_config.py
jsmolar: This was intended to use with test, where the external registry URL will be swapped to a different endpoint. But I scraped this test case later on.

testsuite/openshift/mockserver.py
jsmolar: Do you have something in mind?

pehala: If this will be a single file, I would simply put it in `testsuite/` dir

testsuite/openshift/mockserver.py
```diff
@@ -0,0 +1,38 @@
...
+        """Creates an Expectation - response with body that contains OPA policy (Rego query)"""
+        response = requests.put(
+            urljoin(self.url, "/mockserver/expectation"), verify=False, timeout=5, data=json.dumps(
+                {
+                    "id": expectation_id,
+                    "httpRequest": {
+                        "path": path
+                    },
+                    "httpResponse": {
+                        "body": opa_policy
```
guicassolato, line 24: The response must either be the raw Rego content of the policy (`Content-Type: text/plain`) or comply with OPA Policy API's [_Get a Policy_](https://www.openpolicyagent.org/docs/latest/rest-api/#get-a-policy) endpoint (`Content-Type: application/json`).

jsmolar: I've added `"Content-Type": ["plain/text"]`
We will create tests for content-type in the future.

pehala (approved):

JaurbanRH (approved):

guicassolato (approved):
I won't pretend I understand everything how this works, but at least from the perspective of the configs, expected outputs, etc, it all looks good to me.

PR #112, opened 2022-09-26
Make validators even more lazy

pehala:
* Current state is that all validators are executed immediately
  * This means all config options e.g RHSSO and Mockserver need to present and will be fetched if missing even for a single test which doesn't need it
  * This PR improves it by moving validation to the individual fixtures that load it
* Switched fixture loading order to ensure that config loading happens as soon as possible to fail quickly.


pehala:
> When authorino is not present in `settings.yaml` and `make test` is run all tests fail with error

Thanks for spotting this, it is addressed in https://github.com/Kuadrant/testsuite/pull/119

ficap (approved):

jsmolar (dismissed):

Issue #6, opened 2022-06-14
Add external service metadata test


pehala:
https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/external-metadata.md

pehala:
Closed by #138, #130 and #125

PR #125, opened 2022-10-03
Tests for External auth metadata features

jsmolar:
Based on documentation: https://github.com/Kuadrant/authorino/blob/main/docs/features.md#oidc-userinfo-metadatauserinfo
For issue: https://github.com/Kuadrant/testsuite/issues/6

### DIFF
--- a/testsuite/mockserver.py
+++ b/testsuite/mockserver.py
@@ -1,7 +1,10 @@
 """Module for Mockserver integration"""
+from typing import Union
 from urllib.parse import urljoin
 
 import httpx
+
+from testsuite.utils import ContentType
 
 
 class Mockserver:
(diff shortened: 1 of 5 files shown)


JaurbanRH (approved):

testsuite/mockserver.py
```diff
@@ -1,16 +1,23 @@
 """Module for Mockserver integration"""
+import enum
 from urllib.parse import urljoin
 
 import httpx
 
 
+class ContentType(enum.Enum):
```
pehala, line 8: We could possibly move this to utils and use this in more functions, but this does not need to be done in this PR

pehala (approved):

PR #131, opened 2022-10-14
Add User RHSSO object

pehala:
QoL improvement to RHSSO user manipulation

### DIFF
--- a/testsuite/oidc/rhsso/__init__.py
+++ b/testsuite/oidc/rhsso/__init__.py
@@ -6,7 +6,7 @@ from keycloak import KeycloakOpenID, KeycloakAdmin, KeycloakPostError
 
 from testsuite.oidc import OIDCProvider, Token
 from testsuite.objects import LifecycleObject
-from .objects import Realm, Client
+from .objects import Realm, Client, User
 
 
 # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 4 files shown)


testsuite/oidc/rhsso/objects.py
```diff
@@ -49,15 +49,6 @@ def create_realm_role(self, role_name: str):
         role_id = self.admin.get_realm_role(role_name)["id"]
         return {"name": role_name, "id": role_id}
 
-    def assign_realm_role(self, role, user_id: str):
```
averevki, line 52: there is one more usage of this method in the `rhsso.test_rhsso_roles.py::user_with_role` fixture

averevki (changes requested):

testsuite/oidc/rhsso/__init__.py
```diff
@@ -63,7 +63,8 @@ def commit(self):
             publicClient=False,
             protocol="openid-connect",
             standardFlowEnabled=False)
-        self.user = self.realm.create_user(self.test_username, self.test_password)
+        user_id = self.realm.create_user(self.test_username, self.test_password)
```
averevki, line 66: what if `create_user` method would return already initialized `User` class directly instead of his id?

averevki (changes requested):

azgabur (approved):

PR #133, opened 2022-10-14
Hierarchical AuthConfig

pehala:
Did you miss drastic changes to the Testsuite? Well, in that case, this PR is for you!

# Problem
AuthConfig is a hierarchical resource, you have multiple levels where you can define something, we currently do not account for this and use a completely flat interface and implementation, this results in a huuuge classes (not so now, but will so later)

# Solution
Separate AuthConfig into sections and move the abstract methods there.

# Do we need it?
Well, I am not sure, this is a proof of concept, and I wanted to start a discussion.


testsuite/objects/__init__.py
```diff
@@ -50,21 +65,25 @@ def authorization_url(self):
 class Authorization(LifecycleObject):
     """Object containing Authorization rules and configuration for either Authorino or Kuadrant"""
 
+    @cached_property
     @abc.abstractmethod
-    def add_oidc_identity(self, name, endpoint, credentials, selector):
-        """Adds OIDC identity provider"""
+    def authorization(self) -> Authorizations:
+        """Gives access to authorization settings"""
```
jsmolar, line 71: not setting configuration? Just small nitpicking

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -113,25 +91,62 @@ def add_api_key_identity(self, name, all_namespaces: bool = False,
...
+    def remove_all(self):
+        """Removes all identities from AuthConfig"""
+        self.section.clear()
+
+
+class MetadataSection(Section, Metadata):
+    """Section which contains metadata configuration"""
+    @modify
+    def add_http_metadata(self, name, endpoint, method: Literal["GET", "POST"]):
+        self.add_item(name, {
```
jsmolar, line 108: Is it OK not to have documentation here?

pehala: I believe it is, we have documentation in the interfaces

report-test.html
```diff
@@ -0,0 +1,788 @@
+<!DOCTYPE html>
```
mganisin, line 1: this?

test
```diff
@@ -0,0 +1,30 @@
+#128  Fix dynaconf_l...  Env was applie...  https://github...  2022-10-05T13:...
```
mganisin, line 1: And this?

mganisin: and what about check failures? :-)

mkudlej:
@pehala I think the important question is, would we have multiple identities in authconfig or not?

pehala:
1

> @pehala I think the important question is, would we have multiple identities in authconfig or not?

100% yes

JaurbanRH:
AttributeError: 'IdentitySection' object has no attribute 'committed'

jsmolar (dismissed):

pehala:
> AttributeError: 'IdentitySection' object has no attribute 'committed'

Fixed

jsmolar (approved):

dhlavac (approved):

PR #134, opened 2022-10-18
Added logLevel debug for authorino

jsmolar:
Temporarily adding debug log level for Authorino. 
When we add logs to the test results in the future, we can set this as an optional option for test runs.

### DIFF
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -11,5 +11,6 @@ default:
       password: "testPassword"
   authorino:
     deploy: true
+    log_level: "debug"
   envoy:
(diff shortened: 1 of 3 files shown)


dhlavac (changes requested):
why hardcoding it instead of setting the log level in the configuration file?

jsmolar:
> why hardcoding it instead of setting the log level in the configuration file?

we can add config file option as well if u insist

pehala (approved):

dhlavac (approved):

PR #138, opened 2022-10-21
Add test for metadata UMA resource-level restrictions

jsmolar:


testsuite/tests/kuadrant/authorino/metadata/test_user_info.py
```diff
@@ -13,7 +13,7 @@ def authorization(authorization, rhsso):
     Adds auth metadata OIDC UserInfo which fetches OIDC UserInfo in request-time.
     Adds a simple rule that accepts only when fetched UserInfo contains the email address of the default RHSSO user.
     """
-    user = rhsso.client.admin.get_user(rhsso.user)
+    user = rhsso.client.admin.get_user(rhsso.user.user_id)
```
pehala, line 16: Fixed in #140

testsuite/oidc/rhsso/objects.py
```diff
@@ -71,13 +72,31 @@ def assign_role(self, role_name):
         role = self.admin.get_client_role(realm_management, role_name)
         self.admin.assign_client_role(user["id"], realm_management, role)
 
+    @cached_property
+    def client_named_id(self):
```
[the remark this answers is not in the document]
jsmolar: then we would have `client_id` and `clientId`. But I agree that `client_name_id` is ****.

testsuite/oidc/rhsso/objects.py
```diff
@@ -71,13 +72,31 @@ def assign_role(self, role_name):
...
+        """Note This is different clientId (clientId) than self.client_id (Id), because RHSSO"""
+        return self.admin.get_client(self.client_id)["clientId"]
+
+    @cached_property
```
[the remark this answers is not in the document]
jsmolar: I'm changing this to `@property`

testsuite/openshift/objects/client_secret.py
```diff
@@ -0,0 +1,26 @@
+"""Client Secret CR object"""
+
+from testsuite.openshift.client import OpenShiftClient
+from testsuite.openshift.objects import OpenShiftObject
+
+
+class ClientSecret(OpenShiftObject):
```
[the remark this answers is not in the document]
jsmolar: My plan here is to do the same thing you did in https://github.com/Kuadrant/testsuite/pull/133
This was just temp solution

testsuite/httpx/__init__.py
```diff
@@ -53,6 +53,10 @@ def close(self) -> None:
             file.close()
         self.files = []
 
+    def add_retry_code(self, code):
+        """Add a new retry code to"""
+        self.RETRY_CODES.add(code)
```
pehala, line 58: `RETRY_CODES` is static so this will add it to every Client that exists and will so in the future

testsuite/httpx/__init__.py
```diff
@@ -27,10 +27,10 @@ def __init__(self, msg, response):
 
 class HttpxBackoffClient(Client):
     """Httpx client which retries unstable requests"""
-    RETRY_CODES = {503}
 
     def __init__(self, *, verify: Union[Certificate, bool] = True, cert: Certificate = None, **kwargs):
```
pehala, line 31: This is fine, but how about adding the retry_codes as an optional init parameter?

jsmolar: With this implementation u can easily add a new retry code in your test: `client.add_retry_code(404) and you don't have to define a new client fixture.

pehala (approved):

JaurbanRH (approved):

PR #141, opened 2022-11-03
Add webhooks test

mkudlej:
This PR adds testcase for testing k8s Dynamic Admission Control webhooks, see https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/ which are used for calling Authorino service for authenticating of Ingress object creation.

### DIFF
--- a/testsuite/tests/kuadrant/authorino/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/conftest.py
@@ -30,8 +30,9 @@ def authorino(authorino, openshift, blame, request, testconfig, module_label, au
     labels = authorino_parameters.setdefault("label_selectors", [])
     labels.append(f"testRun={module_label}")
 
+    authorino_parameters.setdefault('name', blame("authorino"))
+
     authorino = AuthorinoCR.create_instance(openshift,
-                                            blame("authorino"),
                                             image=weakget(testconfig)["authorino"]["image"] % None,
                                             log_level=weakget(testconfig)["authorino"]["log_level"] % None,
                                             **authorino_parameters)
--- a/testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
+++ b/testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
@@ -1,0 +1,177 @@
+"""
+Test raw http authorization used in Kubernetes Validating Webhooks.
+"""
+import base64
+from typing import Any, Dict
+import pytest
+
+import openshift as oc
+from openshift import OpenShiftPythonException
+
+from testsuite.objects import Authorization
+from testsuite.certificates import CertInfo
+from testsuite.utils import cert_builder
+from testsuite.openshift.objects.ingress import Ingress
+
+OPA_POLICY = """
+    request := json.unmarshal(input.context.request.http.body).request
+    verb := request.operation
+    ingress := request.object { verb == "CREATE" }
+    ingress := request.oldObject { verb == "DELETE" }
+    forbidden { count(object.get(ingress.spec, "rules", [])) == 0 }
+    rules { count(ingress.spec.rules) == 1; ingress.spec.rules[0] == {} }
+    allow { rules; not forbidden }
+"""
+
+
+@pytest.fixture(scope="session")
+def specific_authorino_name(blame):
+    """Define specific name for authorino which matches with name in authorino certificate"""
+    return blame('authorino')
+
+
+@pytest.fixture(scope="session")
+def authorino_domain(openshift, specific_authorino_name):
+    """
+    Hostname of the upstream certificate sent to be validated by APIcast
+    May be overwritten to configure different test cases
+    """
+    return f"{specific_authorino_name}-authorino-authorization.{openshift.project}.svc"
+
+
+@pytest.fixture(scope="session")
+def certificates(cfssl, authorino_domain, wildcard_domain):
+    """Certificate hierarchy used for the tests.
+    Authorino certificate has *hosts* set to *authorino_domain* value.
+    """
+    chain = {
+        "envoy_ca": CertInfo(children={
+            "envoy_cert": None,
+            "valid_cert": None
+        }),
+        "authorino_ca": CertInfo(children={
+            "authorino_cert": CertInfo(hosts=authorino_domain),
+        }),
+        "invalid_ca": CertInfo(children={
+            "invalid_cert": None
+        })
+    }
+    return cert_builder(cfssl, chain, wildcard_domain)
+
+
+@pytest.fixture(scope="module")
+def authorino_parameters(authorino_parameters, specific_authorino_name):
+    """Setup TLS with specific name for authorino."""
+    authorino_parameters['name'] = specific_authorino_name
+    return authorino_parameters
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(authorization, openshift, module_label, authorino_domain) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+
+    # Authorino should have specific url so it is accessible by k8s webhook
+    authorization.remove_all_hosts()
+    authorization.add_host(authorino_domain)
+
+    # get user info from admission webhook
+    authorization.identity.remove_all()
+    authorization.identity.kubernetes('k8s-userinfo',
+                                      'context.request.http.body.@fromstr|request.userInfo')
+
+    # add OPA policy to process admission webhook request
+    authorization.authorization.opa_policy('features', OPA_POLICY)
+
+    when = [{'selector': 'auth.authorization.features.allow', 'operator': 'eq', 'value': 'true'},
+            {'selector': 'auth.authorization.features.verb', 'operator': 'eq', 'value': 'CREATE'}]
+    kube_attrs = {
+                    'namespace': {'value': openshift.project},
+                    'group': {'value': 'networking.k8s.io'},
+                    'resources': {'value': 'Ingress'},
+                    'verb': {'value': 'create'}
+                 }
+    # add response for admission webhook for creating Ingress
+    authorization.authorization.kubernetes('ingress-authn-k8s-binding-create',
+                                           when, kube_attrs, 1)
+
+    when = [{'selector': 'auth.authorization.features.allow', 'operator': 'eq', 'value': 'true'},
+            {'selector': 'auth.authorization.features.verb', 'operator': 'eq', 'value': 'DELETE'}]
+    kube_attrs = {
+                    'namespace': {'value': openshift.project},
+                    'group': {'value': 'networking.k8s.io'},
+                    'resources': {'value': 'Ingress'},
+                    'verb': {'value': 'delete'}
+                 }
+    # add response for admission webhook for deleting Ingress
+    authorization.authorization.kubernetes('ingress-authn-k8s-binding-delete',
+                                           when, kube_attrs, 1)
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def validating_webhook(openshift, authorino_domain, certificates, blame):
+    """Create validating webhook."""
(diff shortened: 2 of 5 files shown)


Pipfile
```diff
@@ -4,9 +4,9 @@ pylint = "*"
 types-PyYAML = "*"
 # Have commented out python-language-server to make it available quickly
 # for the development
-#python-language-server = "*"
-#pyls-mypy = "*"
-#pyls-isort = "*"
+# python-language-server = "*"
+# pyls-mypy = "*"
+# pyls-isort = "*"
```
JaurbanRH, line 9: Unrelated change

testsuite/objects/__init__.py
```diff
@@ -67,7 +67,7 @@ def add_oidc_identity(self, name, endpoint, credentials, selector):
         """Adds OIDC identity provider"""
 
     @abc.abstractmethod
-    def add_api_key_identity(self, name, all_namespaces, match_label, match_expression,  credentials, selector):
+    def add_api_key_identity(self, name, all_namespaces, match_label, match_expression, credentials, selector):
```
JaurbanRH, line 70: Unrelated change

testsuite/openshift/objects/authorino.py
```diff
@@ -50,6 +51,8 @@ def create_instance(cls, openshift: OpenShiftClient, name, image=None,
                 "enabled": True,
                 "certSecretRef": {"name": listener_certificate_secret}
             }
+        if specific_name:
+            model["metadata"]["name"] = specific_name
```
JaurbanRH, line 55: You can use `name` attribute instead of creating a new one that is for same thing.

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,135 @@
...
+        })
+    }
+    return cert_builder(cfssl, chain, wildcard_domain)
+
+
+@pytest.fixture(scope="module")
+def authorino_parameters(authorino_parameters, specific_authorino_name):
+    """Setup TLS with specific name for authorino."""
+    authorino_parameters['specific_name'] = specific_authorino_name
+    yield authorino_parameters
```
JaurbanRH, line 52: Why `yield` instead of `return` ?

testsuite/openshift/objects/authorino.py
mkudlej: For 99% of testcases is OK to have Authorino with blamed name as it is generated here https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/kuadrant/authorino/conftest.py#L34
Only this testcase needs to have it. But you are right that this is not good solution, so I've found better place to set the Authorino name here testsuite/tests/kuadrant/authorino/conftest.py

pehala: You shouldnt change it in upper conftest either change it properl here and it cannot be done then we need to create a way how to do it here

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
mkudlej: `yield` was there before, see https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/kuadrant/authorino/operator/tls/conftest.py#L160 but I've changed both.

testsuite/openshift/objects/validating_webhook.py
```diff
@@ -0,0 +1,53 @@
...
+    def create_instance(cls, openshift: 'OpenShiftClient', name: str, service_name: str, ca_bundle: str):
+        """Creates base instance"""
+        model: Dict[str, Any] = {
+            'apiVersion': 'admissionregistration.k8s.io/v1',
+            'kind': 'ValidatingWebhookConfiguration',
+            'metadata': {
+                'name': name,
+                'namespace': openshift.project
+            },
+            'webhooks': [
```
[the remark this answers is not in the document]
mkudlej: it will not change I think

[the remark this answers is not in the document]
pehala: If it doesnt need to change and is only useful for one test it doesn't need a class

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Hostname of the upstream certificate sent to be validated by APIcast
+    May be overwritten to configure different test cases
+    """
+    return f"{specific_authorino_name}-authorino-authorization.{openshift.project}.svc"
+
+
+@pytest.fixture(scope="session")
+def certificates(cfssl, authorino_domain, wildcard_domain):
+    """Certificate hierarchy used for the tests"""
```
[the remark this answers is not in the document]
mkudlej: The big difference is that Authorino certificate has `hosts` value to value of `authorino_domain` fixture.

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,135 @@
...
+        ingress := request.oldObject { verb == "DELETE" }
+        forbidden { count(object.get(ingress.spec, "rules", [])) == 0 }
+        rules { count(ingress.spec.rules) == 1; ingress.spec.rules[0] == {} }
+        allow { rules; not forbidden }
+    """
+    authorization.add_opa_policy('features', opa_policy)
+
+    when = [{'selector': 'auth.authorization.features.allow', 'operator': 'eq', 'value': 'true'},
+            {'selector': 'auth.authorization.features.verb', 'operator': 'eq', 'value': 'CREATE'}]
+    kubernetes = {
```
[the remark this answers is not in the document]
mkudlej: I don't have the same feeling :) There are just 2 short dicts. When there is a need to have complex `when` condition we can create wrapper for it with special treatment.

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,135 @@
...
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(authorization, openshift, module_label, authorino_domain) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+    authorization.remove_all_hosts()
+    authorization.add_host(authorino_domain)
+
+    authorization.remove_all_identities()
+    plain = {'authJSON': 'context.request.http.body.@fromstr|request.userInfo'}
+    authorization.add_kubernetes_identity('k8s-userinfo', plain)
```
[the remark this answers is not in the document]
mkudlej: Maybe I don't know what you exactly mean. Validation webhook request contains various information which ca be used but for this test this one is the only one used. Should I move it to separated function then?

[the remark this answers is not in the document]
pehala: `authorization.add_kubernetes_identity('k8s-userinfo', 'context.request.http.body.@fromstr|request.userInfo')` are the input parameters I would like to see (you can use variables ofc)

testsuite/tests/kuadrant/authorino/conftest.py
```diff
@@ -30,8 +30,10 @@ def authorino(authorino, openshift, blame, request, testconfig, module_label, au
     labels = authorino_parameters.setdefault("label_selectors", [])
     labels.append(f"testRun={module_label}")
 
+    authorino_name = authorino_parameters.pop('specific_name', blame("authorino"))
```
pehala, line 33: `authorino_parameters.setdefault("name", blame("authorino"))`

testsuite/openshift/objects/authorino.py
```diff
@@ -14,8 +14,8 @@ class AuthorinoCR(OpenShiftObject, Authorino):
 
     @classmethod
     def create_instance(cls, openshift: OpenShiftClient, name, image=None,
-                        cluster_wide=False, label_selectors: List[str] = None, listener_certificate_secret=None,
-                        log_level=None):
```
pehala, line 18: Unrelated change

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,133 @@
...
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization(authorization, openshift, module_label, authorino_domain) -> Authorization:
+    """In case of Authorino, AuthConfig used for authorization"""
+    authorization.remove_all_hosts()
+    authorization.add_host(authorino_domain)
+
+    authorization.identity.remove_all()
+    plain = {'authJSON': 'context.request.http.body.@fromstr|request.userInfo'}
```
pehala, line 74: I would change the interface to just support `plain = 'context.request.http.body.@fromstr|request.userInfo'`

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,133 @@
...
+    authorization.identity.kubernetes('k8s-userinfo', plain)
+
+
+    authorization.authorization.opa_policy('features', OPA_POLICY)
+
+    when = [{'selector': 'auth.authorization.features.allow', 'operator': 'eq', 'value': 'true'},
+            {'selector': 'auth.authorization.features.verb', 'operator': 'eq', 'value': 'CREATE'}]
+    kube_user = {'valueFrom': {'authJSON': 'auth.identity.username'}}
```
pehala, line 82: Similar here, you were right that there is an option to use constant value but there will most likely be a future refactor that will unify the behaviour of `valueFrom` and `value` and if we cut it down to just `auth.identity.username` I think it will be more readable and easier to work with

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,133 @@
...
+                 }
+    authorization.authorization.kubernetes('ingress-authn-k8s-binding-delete',
+                                           when, kube_user, kube_attrs, 1)
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def validating_webhook(openshift, authorino_domain, certificates):
+    """Create validating webhook."""
+    name = 'check-ingress.authorino.kuadrant.io'
```
pehala, line 109: What will happen if two people run this test in parallel? Wouldnt it break because the name is not randomised?

testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -0,0 +1,133 @@
...
+                                                certificates['authorino_ca'])
+    webhook.commit()
+    yield webhook
+    webhook.delete()
+
+
+# pylint: disable=unused-argument
+def test_authorized_via_http(authorization, openshift, authorino, authorino_domain, validating_webhook):
+    """Test raw http authentization via webhooks."""
+    ingress = Ingress.create_instance(openshift, 'minimal-ingress', rules=[{}])
```
pehala, line 121: Similar problem (parallel runs) here

testsuite/openshift/objects/validating_webhook.py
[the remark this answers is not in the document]
pehala: > If it doesnt need to change and is only useful for one test it doesn't need a class

This is still valid, in current state ValidatingWebhook class is not needed as it is hardcoded to Ingresses only and doesn't add any additional value, this can be a simple fixture and everything would work the same

pehala (commented):
Thanks for the changes! It is much better now!

pehala (dismissed):

jsmolar:
@mkudlej PR looks good, but I would appreciate a short description for `test_webhook.py` that briefly summarizes the tests case (mainly setup)

mkudlej:
@jsmolar Is it better?

pehala (approved):

PR #144, opened 2022-11-08
Add implicit_optional to mypy

pehala:
* mypy 990 (https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html) make strict optional as default, meaning Types that have default value of None are no longer implicitly Optional
  * It is technically forbidden by PEP, but I wont change it now, it is a bigger change and it causes the type to be longer

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,5 @@
 [mypy]
+implicit_optional = True
 
 [mypy-dynaconf.*]
 ignore_missing_imports = True


JaurbanRH (approved):

mypy.ini
```diff
@@ -1,4 +1,5 @@
 [mypy]
+implicit_optional = True
```
dhlavac, line 2: Shouldn't it be  `no_implicit_optional = False` ? Also wouldn't it be better to refactor code?

averevki (approved):

mypy.ini
pehala: > Also wouldn't it be better to refactor code?

I could do that if you want, only reason why I didn't do it in the first place is that Optional is simply too long to be included in every parameter

dhlavac: no, shorter is better =D just found only that one

PR #145, opened 2022-11-09
Use black for formatting

pehala:
This is what black looks like with --line-length set to 120

This is an experimental PR which showcases what black would look like and I would like to hear what you think about code quality.

For Pycharm users, see `https://black.readthedocs.io/en/stable/integrations/editors.html` for configuring support for black.

### DIFF
--- a/testsuite/certificates/__init__.py
+++ b/testsuite/certificates/__init__.py
@@ -15,6 +15,7 @@ class CFSSLException(Exception):
 @dataclasses.dataclass
 class CertInfo:
     """Certificate configuration details"""
+
     hosts: Optional[Union[Collection[str], str]] = None
     ca: bool = False
     children: Optional[Dict[str, Optional["CertInfo"]]] = None
@@ -24,6 +25,7 @@ class CertInfo:
 @dataclasses.dataclass
 class Certificate:
     """Object representing Signed certificate"""
+
     key: str
     certificate: str
     chain: str
@@ -32,13 +34,14 @@ class Certificate:
 @dataclasses.dataclass
 class UnsignedKey:
     """Object representing generated key waiting to be signed"""
+
     key: str
     csr: str
 
 
-def build_cert_request_json(common_name: str,
-                            names: Optional[List[Dict[str, str]]] = None,
-                            hosts: Optional[Collection[str]] = None) -> dict:
+def build_cert_request_json(
+    common_name: str, names: Optional[List[Dict[str, str]]] = None, hosts: Optional[Collection[str]] = None
+) -> dict:
     """
     Build certificate request for the CFSSL client
     :param common_name: certificate identifier
@@ -50,15 +53,13 @@ def build_cert_request_json(common_name: str,
         "CN": common_name,
         "names": names,
         "hosts": hosts,
-        "key": {
-            "algo": "rsa",
-            "size": 4096
-        },
+        "key": {"algo": "rsa", "size": 4096},
     }
 
 
 class CFSSLClient:
     """Client for working with CFSSL library"""
+
     DEFAULT_NAMES = [
         {
             "O": "Red Hat Inc.",
@@ -73,20 +74,20 @@ class CFSSLClient:
         super().__init__()
         self.binary = binary
 
-    def _execute_command(self,
-                         command: str,
-                         *args: str,
-                         stdin: Optional[str] = None,
-                         env: Optional[Dict[str, str]] = None):
+    def _execute_command(
+        self, command: str, *args: str, stdin: Optional[str] = None, env: Optional[Dict[str, str]] = None
+    ):
         args = (self.binary, command, *args)
         try:
-            response = subprocess.run(args,
-                                      stderr=subprocess.PIPE,
-                                      stdout=subprocess.PIPE,
-                                      input=stdin,
-                                      universal_newlines=bool(stdin),
-                                      check=False,
-                                      env=env)
+            response = subprocess.run(
+                args,
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                input=stdin,
+                universal_newlines=bool(stdin),
+                check=False,
+                env=env,
+            )
             if response.returncode != 0:
                 raise CFSSLException(f"CFSSL command {args} returned non-zero response code, error {response.stderr}")
             return json.loads(response.stdout)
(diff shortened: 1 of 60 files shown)


pehala:
Rebased to the latest main, please let me **all** know what kind of changes you do not like so we can discuss them

testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
```diff
@@ -11,11 +11,7 @@
 @pytest.fixture(scope="session")
 def cert_attributes() -> Dict[str, str]:
     """Certificate attributes"""
-    return dict(O="Organization Test",
-                OU="Unit Test",
-                L="Location Test",
-                ST="State Test",
-                C="Country Test")
+    return dict(O="Organization Test", OU="Unit Test", L="Location Test", ST="State Test", C="Country Test")
```
JaurbanRH, line 14: This type of change is welcomed because it makes whole code shorter.

testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
```diff
@@ -129,24 +129,42 @@ def selector_params(module_label):
...
-    }
+    labels = {"authorino.kuadrant.io/managed-by": "authorino", selector_params[0]: selector_params[1]}
     return labels
 
 
 # pylint: disable-msg=too-many-locals
 @pytest.fixture(scope="module")
-def envoy(request, authorino, openshift, create_secret, blame, label, backend,
-          authorino_authority, envoy_authority, envoy_cert, testconfig, authorino_labels):
+def envoy(
```
JaurbanRH, line 138: This type of change is nightmare for every sane person.

testsuite/tests/kuadrant/authorino/operator/tls/mtls/test_mtls_identity.py
```diff
@@ -23,10 +23,19 @@ def test_mtls_success(envoy_authority, valid_cert, envoy):
...
+    "cert_authority, certificate, err, err_match",
+    [
+        pytest.param("envoy_authority", "invalid_cert", ReadError, "unknown ca", id="Invalid certificate"),
+        pytest.param(
+            "invalid_authority",
+            "valid_cert",
+            ConnectError,
+            "certificate verify failed",
+            id="Unknown authority",
+        ),
```
JaurbanRH, line 36: I dislike the inconsistency here.

testsuite/certificates/__init__.py
```diff
@@ -15,6 +15,7 @@ class CFSSLException(Exception):
 @dataclasses.dataclass
 class CertInfo:
     """Certificate configuration details"""
+
```
JaurbanRH, line 18: This type of change is welcome because it makes code more readable by my opinion.

testsuite/httpx/auth.py
```diff
@@ -23,12 +23,12 @@ def token(self):
         return self._token
 
     def _add_credentials(self, request: Request, token):
-        if self.location == 'authorization':
-            request.headers['Authorization'] = f"Bearer {token}"
-        elif self.location == 'headers':
-            request.headers['access_token'] = token
-        elif self.location == 'query':
-            request.url = URL(request.url, params={'access_token': token})
+        if self.location == "authorization":
```
JaurbanRH, line 26: Big plus of Black is that it is consistent on using quotation marks and apostrophes.

testsuite/tests/kuadrant/authorino/operator/tls/mtls/test_mtls_identity.py
mganisin: it is consistent, the length is either below max line length or above it. It's everywhere same

testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
mganisin: not at right enough?
interesting that it was "fine" here: ;-)
https://github.com/Kuadrant/testsuite/pull/145/files#diff-e242f4968001181ce82739996a5e823dc762bb2e988c8bf27d1c6e24240f7126L62

testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
mganisin: it should be welcome, because it brings consistency (as all the black formatting) and that's main benefit of formatter

testsuite/mockserver.py
```diff
@@ -13,30 +13,25 @@ class Mockserver:
...
-                            "Content-Type": [str(content_type)]
-                        },
-                        "body": opa_policy
-                    }
-                }
-            )
+            urljoin(self.url, "/mockserver/expectation"),
+            verify=False,
+            timeout=5,
+            json={
```
jsmolar, line 24: JSON one-line formatting is arguably worse readable.

jsmolar (dismissed):

averevki:
I like the overall consistency of the code. I believe long one-liners can be solved by setting a shorter line length (e.g. to black default 88 characters), if this is an option.

testsuite/mockserver.py
averevki: There is an option to leave dictionary formatting in the "traditional way" if black sees a trailing comma after the last element.

jsmolar: That would be great, but black can't be configured if I remember correctly. Its rules are set (that is what is good about black). But we can still use `ignore`

JaurbanRH (approved):

testsuite/mockserver.py
pehala: @averevki You are right! Thanks for the suggestions. https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma. It is not a config option, but if you have a comma after the last element in dict it won't get squashed to a oneliner. @jsmolar Could you show me all locations where you would like to get rid of one-line dicts?

pehala:
> I like the overall consistency of the code. I believe long one-liners can be solved by setting a shorter line length (e.g. to black default 88 characters), if this is an option.

I consider a length of 120 to be still reasonable also thanks to your suggestion with the magic trailing comma we have an option to "disable" one-liners.

jsmolar (dismissed):

JaurbanRH (approved):

averevki (dismissed):

PR #147, opened 2022-11-09
Use poetry instead of pipenv

pehala:
* Poetry is in general a better tool, so migrating sooner than later is not a bad idea
  * It has better env management, you can have multiple pythons setup per project and Pycahrm respect them and doesn't create its own
  * It is quicker to sync resources
* Remove `pipenv` management from Makefile
     * This means the user is responsible for creating their own poetry env, there are helper `make` targets but they do not ensure it is up-to-date 
* Migrate all configuration files into `pyproject.toml`


.gitignore
```diff
@@ -40,6 +40,7 @@ share/python-wheels/
 *.egg
 MANIFEST
 Pipfile.lock
```
jsmolar, line 42: Remove `Pipfile.lock`

jsmolar (dismissed):
Approved, but do we still need `Pipfile.lock` in gitignore?

pehala:
One additional change I made, I moved all configurations for our tools into `pyproject.toml`, which means we have less files we need to take care of.

JaurbanRH (approved):

averevki:
I would like to also point out the format in which `poetry.lock` is saved. Instead of plain json by Pipenv, Poetry uses TOML format for its lock file. Which is more readable in my opinion.

jsmolar (dismissed):

averevki (approved):

PR #148, opened 2022-11-15
Do not expect hosts in ingress

pehala:
With the recent change, the GLBC stopped adding a host into the Ingress, it can instead be read from the annotations. This is a hot fix to resolve that. It works but will be replaced by a better solution along the way.

### DIFF
--- a/testsuite/openshift/objects/ingress.py
+++ b/testsuite/openshift/objects/ingress.py
@@ -1,6 +1,8 @@
 """Kubernetes Ingress object"""
 from typing import Any, Dict, List, Optional, TYPE_CHECKING
 
+import openshift as oc
+
 from testsuite.objects import LifecycleObject
 from testsuite.openshift.objects import OpenShiftObject
 
@@ -73,12 +75,25 @@ class Ingress(OpenShiftObject, LifecycleObject):
         """Returns rules defined in the ingress"""
         return self.model.spec.rules
 
-    def wait_for_hosts(self, tolerate_failures: int = 5):
-        """Waits until all rules within the ingress have host fields filled"""
-        def _all_rules_have_host(obj):
-            return all("host" in r and len(r.get("host")) > 0 for r in obj.model.spec.rules)
+    @property
+    def host(self):
+        """Returns host of the ingress"""
+        return self.model.metadata.annotations["kuadrant.dev/host.generated"]
 
-        success, _, _ = self.self_selector().until_all(success_func=_all_rules_have_host,
-                                                       tolerate_failures=tolerate_failures)
+    def wait(self, timeout=360):
+        """
+        Waits until the Ingress is ready,
+        temporarily it just checks for certificate as that is almost certainly created afterwards
+        """
+        return self.wait_for_certificate(timeout=timeout)
+
+    def wait_for_certificate(self, tolerate_failures: int = 5, timeout=360):
+        """Waits until all rules within the ingress have host fields filled"""
+        def _certificate_ready(obj):
+            return obj.model.metadata.annotations["kuadrant.dev/certificate-status"] == "ready"
+
+        with oc.timeout(timeout):
+            success, _, _ = self.self_selector().until_all(success_func=_certificate_ready,
+                                                           tolerate_failures=tolerate_failures)
 
         return success
--- a/testsuite/tests/glbc/test_ingress_reconciliation.py
+++ b/testsuite/tests/glbc/test_ingress_reconciliation.py
@@ -1,6 +1,4 @@
 """Basic tests for ingress reconciliation"""
-import time
-
 import backoff
 import httpx
 import pytest
@@ -29,18 +27,15 @@ def backend_ingress(request, backend, blame):
 @backoff.on_exception(backoff.fibo, exception=httpx.ConnectError, max_time=600)
 def test_ingress_host_add(backend_ingress):
     """Creates ingress for a backend and checks for host field filled by glbc, checks host points to backend"""
-    rules = backend_ingress.rules
-    assert len(rules) == 1
-
-    backend_ingress.wait_for_hosts()
-    host = rules[0].get("host")
+    backend_ingress.wait_for_ready()
+    host = backend_ingress.host
 
     assert host
 
-    # needed because of negative dns caching
-    # once https://github.com/kcp-dev/kcp-glbc/issues/354 is implemented this can be removed and reimplemented
-    time.sleep(20)  # wait until dns record is propagated
+    # # needed because of negative dns caching
+    # # once https://github.com/kcp-dev/kcp-glbc/issues/354 is implemented this can be removed and reimplemented
+    # time.sleep(20)  # wait until dns record is propagated
 
-    response = httpx.get(f"http://{host}")
+    response = httpx.get(f"http://{host}", follow_redirects=True)
 
     assert response.status_code == 200


testsuite/openshift/objects/ingress.py
```diff
@@ -69,16 +71,17 @@ def create_service_ingress(cls, openshift: 'OpenShiftClient', name, service_name
         return cls.create_instance(openshift, name, rules=[rule])
 
     @property
-    def rules(self):
+    def host(self):
         """Returns rules defined in the ingress"""
```
ficap, line 75: fix this docstring as it no longer corresponds to property description

testsuite/openshift/objects/ingress.py
```diff
@@ -69,16 +71,17 @@ def create_service_ingress(cls, openshift: 'OpenShiftClient', name, service_name
         return cls.create_instance(openshift, name, rules=[rule])
 
     @property
-    def rules(self):
```
ficap, line 72: as it already has been implemented I would keep the rules property present. I think that it will find its use soon.

testsuite/tests/glbc/test_ingress_reconciliation.py
```diff
@@ -29,18 +27,15 @@ def backend_ingress(request, backend, blame):
 @backoff.on_exception(backoff.fibo, exception=httpx.ConnectError, max_time=600)
 def test_ingress_host_add(backend_ingress):
     """Creates ingress for a backend and checks for host field filled by glbc, checks host points to backend"""
-    rules = backend_ingress.rules
-    assert len(rules) == 1
-
-    backend_ingress.wait_for_hosts()
-    host = rules[0].get("host")
+    backend_ingress.wait_for_certificate()
```
ficap, line 30: I'd go for a solution where we first wait for host and then for certificate as it would be more robust. (certificate generation can be disabled). and also split this test into two separate ones as this one should test that dns record has been set up properly and the other one should test that the certificate is generated and used properly. these are two test cases IMO. (yes I know the certificate would not work without generated hostname but I would still vote for separate solution)

testsuite/tests/glbc/test_ingress_reconciliation.py
```diff
@@ -29,18 +27,15 @@ def backend_ingress(request, backend, blame):
...
 
-    # needed because of negative dns caching
-    # once https://github.com/kcp-dev/kcp-glbc/issues/354 is implemented this can be removed and reimplemented
-    time.sleep(20)  # wait until dns record is propagated
+    # # needed because of negative dns caching
+    # # once https://github.com/kcp-dev/kcp-glbc/issues/354 is implemented this can be removed and reimplemented
+    # time.sleep(20)  # wait until dns record is propagated
 
-    response = httpx.get(f"http://{host}")
+    response = httpx.get(f"http://{host}", follow_redirects=True)
```
ficap, line 39: why are `follow_redirects=True` needed here? what component does require this?

ficap (changes requested):

testsuite/tests/glbc/test_ingress_reconciliation.py
pehala: It goes through load balancer and returns 302

testsuite/tests/glbc/test_ingress_reconciliation.py
pehala: Host is there much sooner than certificate, almost instantly. To circumvent direct sleep I used certificate wait until the negative caching issue is fixed

testsuite/tests/glbc/test_ingress_reconciliation.py
ficap: hmm makes sense but it's weird how it could have worked in the past.

testsuite/tests/glbc/test_ingress_reconciliation.py
ficap: I don't like "almost instantly" since it can become flaky as this test is supposed to be the smoke one. and it is not clear that this wait should serve (even temporarily) as a replacement for negative caching issue, I'd appreciate some comment.

testsuite/tests/glbc/test_ingress_reconciliation.py
```diff
@@ -29,18 +27,15 @@ def backend_ingress(request, backend, blame):
...
-    host = rules[0].get("host")
+    backend_ingress.wait_for_ready()
+    host = backend_ingress.host
 
     assert host
 
-    # needed because of negative dns caching
-    # once https://github.com/kcp-dev/kcp-glbc/issues/354 is implemented this can be removed and reimplemented
-    time.sleep(20)  # wait until dns record is propagated
+    # # needed because of negative dns caching
```
mganisin, line 35: hihi

testsuite/openshift/objects/ingress.py
```diff
@@ -73,12 +75,25 @@ def rules(self):
...
+        return self.model.metadata.annotations["kuadrant.dev/host.generated"]
+
+    def wait_for_ready(self):
+        """
+        Waits until the Ingress is ready,
+        temporarily it just checks for certificate as that is almost certainly created afterwards
+        """
+        return self.wait_for_certificate()
+
+    def wait_for_certificate(self, tolerate_failures: int = 5):
```
mganisin, line 90: if this should exist at all, should it be rather `_private`?

testsuite/tests/glbc/test_ingress_reconciliation.py
```diff
@@ -29,18 +27,15 @@ def backend_ingress(request, backend, blame):
 @backoff.on_exception(backoff.fibo, exception=httpx.ConnectError, max_time=600)
 def test_ingress_host_add(backend_ingress):
     """Creates ingress for a backend and checks for host field filled by glbc, checks host points to backend"""
-    rules = backend_ingress.rules
-    assert len(rules) == 1
-
-    backend_ingress.wait_for_hosts()
-    host = rules[0].get("host")
+    backend_ingress.wait_for_ready()
```
mganisin, line 30: also if this should not be `wait()` why shouldnt it remain `wait_for_hosts()`? At the end, is `host` without waiting needed use case? (`host` is property now, can do much more than just return value)

testsuite/openshift/objects/ingress.py
```diff
@@ -73,12 +75,25 @@ def rules(self):
         """Returns rules defined in the ingress"""
         return self.model.spec.rules
 
-    def wait_for_hosts(self, tolerate_failures: int = 5):
+    @property
+    def host(self):
+        """Returns host of the ingress"""
+        return self.model.metadata.annotations["kuadrant.dev/host.generated"]
+
+    def wait_for_ready(self):
```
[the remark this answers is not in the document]
pehala: I think `wait(timeout)` is reasonable

testsuite/tests/glbc/test_ingress_reconciliation.py
pehala: I dont like mixing value access and waiting because it breaks use cases like "I have one ingress that I know is ready, and I want to access it in two different tests" so I would rather keep it separate

testsuite/openshift/objects/ingress.py
pehala: Could be, but at the same time, I do not see the value of hiding it

mganisin: Having the public/official interface minimal

testsuite/tests/glbc/test_ingress_reconciliation.py
mganisin: then maybe `def get_host(self, wait=False):` (together with `host` property that could or could not wait by default) sounds also good

pehala:
Closed as the GLBC is gonna change significantly anyway

PR #153, opened 2022-12-07
Add ability to create HttpxOidcClientAuth from User

pehala:
Should help with cases where you have multiple users

### DIFF
--- a/testsuite/tests/kuadrant/authorino/metadata/test_user_info.py
+++ b/testsuite/tests/kuadrant/authorino/metadata/test_user_info.py
@@ -34,6 +34,6 @@ def test_correct_auth(client, auth):
 
 def test_incorrect_auth(client, rhsso, user2):
     """Updates RHSSO user email address and tests incorrect auth"""
-    auth = HttpxOidcClientAuth(rhsso.get_token(user2.username, user2.password), "authorization")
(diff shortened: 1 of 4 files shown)


JaurbanRH (approved):

testsuite/tests/kuadrant/authorino/metadata/test_user_info.py
```diff
@@ -34,6 +34,6 @@ def test_correct_auth(client, auth):
 
 def test_incorrect_auth(client, rhsso, user2):
     """Updates RHSSO user email address and tests incorrect auth"""
-    auth = HttpxOidcClientAuth(rhsso.get_token(user2.username, user2.password), "authorization")
+    auth = HttpxOidcClientAuth.from_user(rhsso.get_token, user2, "authorization")
```
jsmolar, line 37: remove "authorization"

pehala: It is a default so it can be omitted but keep it it should do no harm

jsmolar: OK

jsmolar (approved):

PR #155, opened 2022-12-19
Fix scopes in tls conftest

averevki:
Correct scopes for the tls conftest, so there is actually an option to overwrite the default certificate hierarchy. Currently, conftest fixtures always return certificates from the first found `certificates` fixture in the testsuite, as they have `session` scope. With new `module` scopes they will return the certificate according to the overwritten fixture (if there is one).

Also, the secret creation scope changed from `session` to `module`, so these secrets aren't left until the end of the testsuite execution.


pehala:
I am not sure that is the problem:
1) According to what I read documentation, it should work: https://docs.pytest.org/en/6.2.x/fixture.html#overriding-fixtures-on-various-levels
2) The override is used abundantly in the test suite, and for `certificates` specifically, the `test_webhook.py` overrides it.

pehala:
Never mind, you are correct @averevki. The certificates are correctly overridden however the fixtures that depend on it are not recreated.

pehala (approved):

PR #168, opened 2023-01-24
Fix authconfig instance parameters

averevki:
Fix to the https://github.com/Kuadrant/testsuite/pull/158. In the AuthConfig instance creation, there was still parsed string instead of the route object or list with hostnames.

### DIFF
--- a/testsuite/openshift/objects/auth_config/__init__.py
+++ b/testsuite/openshift/objects/auth_config/__init__.py
@@ -1,6 +1,6 @@
 """AuthConfig CR object"""
 from functools import cached_property
-from typing import Dict
+from typing import Dict, List
 
 from testsuite.objects import Authorization, Responses, Metadata, Identities, Authorizations
 from testsuite.openshift.client import OpenShiftClient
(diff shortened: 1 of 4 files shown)


testsuite/openshift/objects/auth_config/__init__.py
```diff
@@ -50,7 +50,7 @@ def create_instance(cls, openshift: OpenShiftClient, name, route: Route,
                 "namespace": openshift.project
             },
             "spec": {
-                "hosts": hostnames or route.hostnames
+                "hosts": hostnames or (route.hostnames if route else None) or []
```
pehala, line 53: why this change?

averevki: there will be an error if hostnames and route are None. Because NoneType has no attributes

pehala: But you need either hostnames or route, otherwise it doesnt make any sense

averevki: yes, that's just an error treatment. there is an `add_host` function below to add hosts manually if you need it

pehala: I would add that only after we found usecase where that makes sense, I think for now the current way is alright

pehala (approved):

JaurbanRH (approved):

PR #169, opened 2023-01-26
Simplify Mockserver expectation

jsmolar:
This PR adds some functionality and simplifies Mockserver
- removes the unnecessary `path` argument for the `create_expectation` method. The `expectation_id` is used to create a path for expectation. It is possible to create expectations with scope function.
- adds a new method that creates an expectation for templated responses: https://www.mock-server.com/mock_server/response_templates.html
- an instance of mockserver now stores live expectations

### DIFF
--- a/testsuite/mockserver.py
+++ b/testsuite/mockserver.py
@@ -8,30 +8,63 @@ from testsuite.utils import ContentType
 
 
 class Mockserver:
-    """Mockserver deployed in Openshift (located in Tools or self-managed)"""
+    """
+    Mockserver deployed in Openshift (located in Tools or self-managed)
+    All existing expectations are stored in `self.expectations: dict[expectation_id, expectation_path]`
+    """
 
     def __init__(self, url):
         self.url = url
+        self.expectations = {}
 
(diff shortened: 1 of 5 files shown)


testsuite/mockserver.py
```diff
@@ -40,3 +73,12 @@ def clear_expectation(self, expectation_id):
                     "id": expectation_id
                 }
             ).raise_for_status()
+        del self.expectations[expectation_id]
+
+    def verify_expectation(self, path):
```
pehala, line 78: If you need just verification you can use https://www.mock-server.com/mock_server/verification.html which doesn't require a expectation at all (see https://github.com/3scale-qe/3scale-tests/pull/488)

jsmolar: The verification endpoint does not provide information about the number of requests received.

jsmolar: And I need expectation for random UUID

pehala (approved):
I would be happier if the new methods are added only when needed (=in the PR that requires them), but I will approve this nonetheless.

JaurbanRH (approved):

PR #175, opened 2023-02-08
Add tests for HTTPRoute reconciliation

### LINKED ISSUE #173: Add HTTPRoute reconciliation tests
**Summary**

In the initial Kuadrant release, there were multiple bugs related to the reconciliation of HTTPRoute

**Test cases**

- [x] Updating HTTPRoute hostnames is propagated (Already covered by a test_remove_hosts)
- [ ] Updating HTTPRoute matches are propagated

pehala:
Closes #173, depends on #160

### DIFF
--- a/testsuite/tests/kuadrant/reconciliation/conftest.py
+++ b/testsuite/tests/kuadrant/reconciliation/conftest.py
@@ -1,0 +1,24 @@
+"""Conftest for reconciliation tests"""
+import backoff
+import pytest
+
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization, kuadrant):
+    """Commits all important stuff before tests"""
+    if not kuadrant:
+        pytest.skip("Reconciliation test can only run on Kuadrant for now")
+    request.addfinalizer(authorization.delete)
+    authorization.commit()
+
+
+@pytest.fixture(scope="module")
+def resilient_request(client):
+    """Fixture which allows to send retrying requests until the expected status code is returned"""
+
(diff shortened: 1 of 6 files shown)


testsuite/tests/kuadrant/reconciliation/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+def commit(request, authorization, kuadrant):
+    """Commits all important stuff before tests"""
+    if not kuadrant:
+        pytest.skip("Reconciliation test can only run on Kuadrant for now")
+    request.addfinalizer(authorization.delete)
+    authorization.commit()
+
+
+@pytest.fixture(scope="module")
+def resilient_request(client):
```
jsmolar, line 16: `HttpxBackoffClient` is doing the same thing no?

pehala: Not exactly, `HttpxBackoffClient` is currently only for backoff for 503 (we backoff no and has much higher timeout period, technically the `HttpxBackoffClient` could be updated to support even these use case if you would like so

jsmolar: I added `add_retry_code` in the past, where you can adjust retry codes.

pehala: You can, but I reuse the client and I would have to remove it afterwards, which seems suboptimal to say the least. I also tried to add to Client constructor but that would mean I have to create new client for each request (basically copying the client fixtures), which is also not great.

jsmolar (approved):

azgabur (approved):

PR #180, opened 2023-03-14
Make envoy number of replicas configurable

ficap:


pehala:
Closed as it is not needed currently

PR #183, opened 2023-03-16
Add Edge Authentication tests

### LINKED ISSUE #15: Add tests for Edge Authentication
Authorino can be used as an Authentication provider which will issue its own wristband token.

https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/edge-authentication-architecture-festival-wristbands.md

- Test for the successful authorization with issued wristband token (200 OK)
- Test for the failed authorization with an incorrect token (401 Unauthorized)
- Test for the verification of information in the decoded wristband token

averevki:
- Move cfssl client to the root conftest
- Add Edge (Wristband) envoy template
- Add Edge Authentication tests
  - Test for the successful authorization with issued wristband token (200 OK)
  - Test for the failed authorization with an incorrect token (401 Unauthorized)
  - Test for the claims verification in the decoded wristband token

Closes #15

### DIFF
--- a/testsuite/openshift/client.py
+++ b/testsuite/openshift/client.py
@@ -3,7 +3,7 @@
 import enum
 import os
 from functools import cached_property
-from typing import Dict, Optional
+from typing import Dict, Optional, Literal
 from urllib.parse import urlparse
 
 import openshift as oc
@@ -132,14 +132,28 @@ class OpenShiftClient:
         success, _, _ = selector.until_all(success_func=lambda obj: "readyReplicas" in obj.model.status)
         return success
 
-    def create_tls_secret(self, name: str, certificate: Certificate, labels: Optional[Dict[str, str]] = None):
-        """Creates a TLS secret"""
+    def create_tls_secret(
+        self,
+        name: str,
+        certificate: Certificate,
+        cert_name: str = "tls.crt",
+        key_name: str = "tls.key",
+        type_: Literal["kubernetes.io/tls", "Opaque"] = "kubernetes.io/tls",
+        labels: Optional[Dict[str, str]] = None,
+    ):
+        """Creates a TLS or signing Opaque secret"""
         model: Dict = {
             "kind": "Secret",
             "apiVersion": "v1",
-            "metadata": {"name": name, "labels": labels},
-            "stringData": {"tls.crt": certificate.chain, "tls.key": certificate.key},
-            "type": "kubernetes.io/tls",
+            "metadata": {
+                "name": name,
+                "labels": labels,
+            },
+            "stringData": {
+                cert_name: certificate.chain,
+                key_name: certificate.key,
+            },
+            "type": type_,
         }
 
         with self.context:
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -9,6 +9,7 @@ from weakget import weakget
 from testsuite.mockserver import Mockserver
 from testsuite.oidc import OIDCProvider
 from testsuite.config import settings
+from testsuite.certificates import CFSSLClient
 from testsuite.oidc.auth0 import Auth0Provider
 from testsuite.openshift.httpbin import Httpbin
 from testsuite.openshift.envoy import Envoy
@@ -141,6 +142,15 @@ def auth0(testconfig):
         return Auth0Provider(section["url"], section["client_id"], section["client_secret"])
     except KeyError as exc:
         return pytest.skip(f"Auth0 configuration item is missing: {exc}")
+
+
+@pytest.fixture(scope="session")
+def cfssl(testconfig):
+    """CFSSL client library"""
+    client = CFSSLClient(binary=testconfig["cfssl"])
+    if not client.exists:
+        pytest.skip("Skipping CFSSL tests as CFSSL binary path is not properly configured")
+    return client
 
 
 @pytest.fixture(scope="module")
--- a/testsuite/tests/kuadrant/authorino/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/conftest.py
@@ -25,7 +25,7 @@ def authorino(authorino, openshift, blame, request, testconfig, module_label, au
     if not testconfig["authorino"]["deploy"]:
         if len(authorino_parameters) > 0:
             return pytest.skip("Can't change parameters of already deployed Authorino")
-        return PreexistingAuthorino(testconfig["authorino"]["url"])
+        return PreexistingAuthorino(testconfig["authorino"]["auth_url"], testconfig["authorino"]["oidc_url"])
 
     labels = authorino_parameters.setdefault("label_selectors", [])
     labels.append(f"testRun={module_label}")
--- a/testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
@@ -3,7 +3,7 @@ from typing import Optional, Dict
 
 import pytest
 
-from testsuite.certificates import CFSSLClient, Certificate, CertInfo
+from testsuite.certificates import Certificate, CertInfo
 from testsuite.openshift.envoy import TLSEnvoy
 from testsuite.utils import cert_builder
 
@@ -67,15 +67,6 @@ def create_secret(blame, request, openshift):
         return secret_name
 
     return _create_secret
-
-
-@pytest.fixture(scope="session")
-def cfssl(testconfig):
-    """CFSSL client library"""
-    client = CFSSLClient(binary=testconfig["cfssl"])
-    if not client.exists:
-        pytest.skip("Skipping CFSSL tests as CFSSL binary path is not properly configured")
-    return client
 
 
 @pytest.fixture(scope="session")
--- a/testsuite/tests/kuadrant/authorino/wristband/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/wristband/conftest.py
@@ -1,0 +1,129 @@
+"""Conftest for Edge Authentication tests"""
+from importlib import resources
+
+import pytest
+
+from testsuite.openshift.objects.auth_config import AuthConfig
+from testsuite.openshift.envoy import Envoy
+from testsuite.certificates import CertInfo
+from testsuite.objects import Authorization
+from testsuite.utils import cert_builder
+
+
+@pytest.fixture(scope="module")
+def run_on_kuadrant():
+    """Kuadrant doesn't allow customization of Authorino parameters"""
+    return False
+
+
+@pytest.fixture(scope="session")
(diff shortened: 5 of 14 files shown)


testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+            },
+        }
+    )
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def wristband_token(client, auth):
+    """Test token acquirement from oidc endpoint"""
+    client.headers = {"Authorization": "Bearer " + auth.token.access_token}
```
pehala, line 65: if you are using auth, you dont need to add headers manually, you can use `client.get("/auth", auth=auth)`

testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+    """Client for second AuthConfig"""
+    client = envoy.client()
+    client.base_url = f"http://{hostname2}"
+    yield client
+    client.close()
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, commit, certificates, authorization2):
```
pehala, line 107: I am not sure certificates need to be there, they should be requested automatically either way

testsuite/tests/kuadrant/authorino/wristband/test_wristband.py
```diff
@@ -0,0 +1,30 @@
...
+def test_wristband_token_claims(wristband_token, wristband_endpoint, certificates):
+    """Verify acquired jwt token claims"""
+    # extract public key from the certificate
+    cert_encoded = certificates["signing_ca"].certificate.encode("utf-8")
+    cert_object = load_pem_x509_certificate(cert_encoded)
+    public_key = cert_object.public_key()
+    # decode jwt to examine claims
+    wristband_decoded = jwt.decode(wristband_token, key=public_key, algorithms=["RS256"])
+    assert wristband_decoded["exp"] - wristband_decoded["iat"] == 300
+    assert wristband_decoded["iss"] == wristband_endpoint
```
pehala, line 16: Could you also test that the token contains less information than the original one from RHSSO?

testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+@pytest.fixture(scope="module")
+def envoy_template():
+    """Envoy with additional route match"""
+    return "envoy_wristband.yaml"
+
+
+@pytest.fixture(scope="module")
+def wristband_endpoint(openshift, authorino, authorization_name):
+    """Authorino oidc wristband endpoint"""
+    return f"http://{authorino.oidc_url}:8083/{openshift.project}/{authorization_name}/wristband"
```
pehala, line 37: Do you need to specify the openshift project?

testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,110 @@
+"""Conftest for Edge Authentication tests"""
+import pytest
+
+from testsuite.openshift.objects.auth_config import AuthConfig
+from testsuite.certificates import CertInfo
+from testsuite.objects import Authorization
+from testsuite.utils import cert_builder
+
+
```
pehala, line 9: There is a missing `run_on_kuadrant` fixture (should return False), as you need custom envoy

testsuite/tests/conftest.py
```diff
@@ -235,12 +235,20 @@ def backend(request, openshift, blame, label):
 
 
 @pytest.fixture(scope="module")
-def envoy(request, kuadrant, authorino, openshift, blame, backend, module_label, testconfig) -> Proxy:
+def envoy_template():
```
pehala, line 238: I would not do this in root conftest, we are trying to avoid using envoy modification as much as possible, I would override envoy fixture in wristband conftest instead.

testsuite/tests/kuadrant/authorino/conftest.py
```diff
@@ -64,6 +65,15 @@ def auth(oidc_provider):
     return HttpxOidcClientAuth(oidc_provider.get_token, "authorization")
 
 
+@pytest.fixture(scope="session")
+def cfssl(testconfig):
```
pehala, line 69: You can move it up to the root conftest in this particular case

testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+    assert response.headers.get("wristband-token") is not None
+    return response.headers["wristband-token"]
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def authorization2_name(blame, oidc_provider):
+    """Second AuthConfig unique name"""
+    return blame("authz2")
```
pehala, line 77: Do you need the name in multiple fixtures? If not you can generate it only where it is needed.

testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@pytest.fixture(scope="module")
+def hostname2(envoy, blame):
+    """Second route for the envoy"""
```
pehala, line 82: Which route is it? It would be great if it was easy to see if this is the wristband route or the other one.

testsuite/openshift/client.py
```diff
@@ -149,6 +149,22 @@ def create_tls_secret(self, name: str, certificate: Certificate, labels: Optiona
         with self.context:
             return oc.create(model, ["--save-config=true"])
 
+    def create_signing_secret(self, name: str, key: str, labels: Optional[Dict[str, str]] = None):
```
pehala, line 152: Would it make sense to update the existing create_tls_secret method? e.g. ability to rename the keys?

testsuite/tests/kuadrant/authorino/wristband/conftest.py
pehala: What does "core" mean? I would appreciate a better docstring

testsuite/tests/kuadrant/authorino/wristband/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+    authorization2 = AuthConfig.create_instance(
+        openshift, blame("auth-core"), None, hostnames=[hostname_core], labels={"testRun": module_label}
+    )
+    authorization2.identity.oidc("edge-authenticated", wristband_endpoint)
+    return authorization2
+
+
+@pytest.fixture(scope="module")
+def client2(hostname_core, envoy):
+    """Client for second AuthConfig"""
```
pehala, line 105: I would again opt for better naming that reflects more its purpose, in both name and docstring

testsuite/tests/kuadrant/authorino/wristband/test_wristband.py
```diff
@@ -0,0 +1,27 @@
...
+from jose import jwt
+
+
+def test_wristband_token_claims(oidc_provider, auth, wristband_token, wristband_endpoint, certificates):
+    """Verify acquired jwt token claims"""
+    wristband_decoded = jwt.decode(wristband_token, certificates["signing_ca"].certificate)
+    assert wristband_decoded["exp"] - wristband_decoded["iat"] == 300
+    assert wristband_decoded["iss"] == wristband_endpoint
+    # check differences in claims between rhsso token and acquired wristband token
+    access_token_decoded = jwt.decode(auth.token.access_token, oidc_provider.get_public_key(), audience="account")
```
pehala, line 11: There is a `decode_token` method in the python-keycloak library

pehala, line 11: By using `get_public_key()` without specifically requesting RHSSO, this test will fail if we switch to auth0, as the method is not there

averevki: This method uses a hard-coded audience claim check as client id, which does not match the actual audience value (it is a known issue https://stackoverflow.com/questions/53550321/keycloak-gatekeeper-aud-claim-and-client-id-do-not-match). There is an option to not verify audience and use this build-in method, but I would prefer to decode it by using the jwt library (as it is now).

testsuite/openshift/client.py
averevki: I would prefer to leave these secrets as they are for now. Because, in my opinion, they have a lot of differences, such as different types, key names, count of secret pairs.

pehala: Would it work with both key and cert in the secret? If yes, then I think adding the option to rename the key and change the type would be enough and it wouldn't be that different

averevki:
rebase

testsuite/openshift/client.py
```diff
@@ -132,14 +132,29 @@ def is_ready(self, selector: Selector):
...
 
-    def create_tls_secret(self, name: str, certificate: Certificate, labels: Optional[Dict[str, str]] = None):
-        """Creates a TLS secret"""
+    def create_key_secret(
+        self,
+        name: str,
+        certificate: Certificate,
+        type_: Literal["kubernetes.io/tls", "Opaque"],
+        labels: Optional[Dict[str, str]] = None,
+    ):
```
pehala, line 141: I would add two parameters: key_name, cert_name which we would use as keys for the secret and wouldn't make it bound to the specific type

testsuite/tests/kuadrant/authorino/wristband/test_wristband.py
```diff
@@ -0,0 +1,27 @@
...
+    for claim in ["preferred_username", "email", "realm_access", "resource_access"]:
+        assert claim in access_token_decoded
+        assert claim not in wristband_decoded
+
+
+def test_wristband_success(client_authenticated, wristband_token):
+    """Test successful Edge Authentication"""
```
pehala, line 18: Please explain what the test tests and why it should succeed

testsuite/tests/kuadrant/authorino/operator/tls/conftest.py
```diff
@@ -62,22 +62,15 @@ def create_secret(blame, request, openshift):
 
     def _create_secret(certificate: Certificate, name: str, labels: Optional[Dict[str, str]] = None):
         secret_name = blame(name)
-        secret = openshift.create_tls_secret(secret_name, certificate, labels=labels)
+        secret = openshift.create_tls_secret(
+            secret_name, certificate, "tls.crt", "tls.key", "kubernetes.io/tls", labels=labels
```
pehala, line 66: You can have tls.crt, tls.key and "kubernetes.io/tls" as defaults

jsmolar (approved):

PR #190, opened 2023-04-06
Remove auto_refresh_token as it is deprecated and enabled by default

pehala:


averevki (commented):
One more occurrence in `testsuite/oidc/rhsso/objects.py`

pehala:
Fixed

jsmolar (approved):

Issue #76, opened 2022-09-02
Add test for caching 


pehala:
You can set up caching for individual sections of AuthConfig.

https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/caching.md

PR #192, opened 2023-04-11
Add tests for Common feature - Caching

jsmolar:
Add simple tests for caching of metadata. I think we should discuss if more tests for caching are required. Examples of operations where caching may be desired: are OAuth2 token introspection, fetching of metadata from external sources (via HTTP request), and complex OPA policies.

Closes https://github.com/Kuadrant/testsuite/issues/76


testsuite/tests/kuadrant/authorino/common_feature/caching/metadata/conftest.py
```diff
@@ -0,0 +1,50 @@
...
+        module_label, mockserver.get_expectation_endpoint(module_label), "GET", cache=meta_cache
+    )
+    authorization.responses.add(
+        {"name": "auth-json", "json": {"properties": [{"name": "auth", "valueFrom": {"authJSON": "auth.metadata"}}]}}
+    )
+    return authorization
+
+
+@pytest.fixture(autouse=True)
+def commit(request, authorization):
```
pehala, line 47: Should not be necessary

testsuite/tests/kuadrant/authorino/common_feature/__init__.py
pehala, line 1: `common_feature` directory is not needed

testsuite/tests/kuadrant/authorino/common_feature/caching/metadata/conftest.py
pehala, line 1: Theoretically, you could also include a testcase where you would check if not specifying cache will really not cache the result

testsuite/tests/kuadrant/authorino/common_feature/caching/metadata/conftest.py
jsmolar: This is function scope. I need to reset the cached value after every test.

testsuite/openshift/objects/auth_config/sections.py
pehala, line 1: Why was this file changed? I do not see any of the new functions used in this PR nor do I reason to change add_item

testsuite/utils.py
```diff
@@ -136,3 +137,11 @@ def fire_requests(client, max_requests, period, grace_requests=0, iterations=1,
         assert httpx.get(url).status_code == 429, f"Iteration {iteration + 1} failed to start limiting"
         sleep(period)
         assert httpx.get(url).status_code == 200, f"Iteration {iteration + 1} failed to reset limits"
+
+
+def extract_from_response(response, *args):
+    """Extracts the UUID expectation from the request metadata"""
+    result = json.loads(response.json()["headers"]["Auth-Json"])["auth"]
+    for arg in args:
+        result = result.get(arg)
```
averevki, line 146: I think `assert` or plain `result[arg]` here will produce more comprehensive error message, if any key does not exist

testsuite/utils.py
```diff
@@ -136,3 +137,11 @@ def fire_requests(client, max_requests, period, grace_requests=0, iterations=1,
         assert httpx.get(url).status_code == 429, f"Iteration {iteration + 1} failed to start limiting"
         sleep(period)
         assert httpx.get(url).status_code == 200, f"Iteration {iteration + 1} failed to reset limits"
+
+
+def extract_from_response(response, *args):
+    """Extracts the UUID expectation from the request metadata"""
```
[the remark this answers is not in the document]
jsmolar: No, you are right. I have to change this.

testsuite/utils.py
```diff
@@ -136,3 +137,11 @@ def fire_requests(client, max_requests, period, grace_requests=0, iterations=1,
...
+    result = json.loads(response.json()["headers"]["Auth-Json"])["auth"]
```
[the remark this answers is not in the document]
jsmolar: I found 4 potential usages for this method in the current code. All of them are getting the "auth" key. Therefore I would leave this as it is.

testsuite/utils.py
jsmolar: agreed, I will use `result[arg]`

pehala (approved):

averevki (approved):

PR #193, opened 2023-04-13
Add back poetry management to Makefile

pehala:
- Only if there is no other env installed
- I am terrible at shell


mganisin:
Merge it and update later or update now, but you should warn that some env exists therefore nothing is updated according to recent changes in pyproject.toml
Personally I wouldn't be shy to implement some kind of `force=1` flag (env exists & no force -> report error; forced -> warn)

mganisin (approved):

PR #198, opened 2023-04-21
Stability fixes

pehala:
Should fix a few failures

### DIFF
--- a/testsuite/tests/kuadrant/reconciliation/conftest.py
+++ b/testsuite/tests/kuadrant/reconciliation/conftest.py
@@ -4,6 +4,20 @@ import pytest
 
 
 @pytest.fixture(scope="module")
+def authorization(authorization):
+    """Add anonymous identity"""
+    authorization.identity.anonymous("anonymous")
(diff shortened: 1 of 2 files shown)


testsuite/tests/kuadrant/reconciliation/conftest.py
```diff
@@ -3,6 +3,20 @@
 import pytest
 
 
+@pytest.fixture(scope="module")
+def authorization(authorization):
+    """Add RHSSO identity"""
+    authorization.identity.anonymous("anonymous")
```
averevki, line 9: Is this rhsso identity?

pehala: It is not, it is anonymous identity, meaning anyone can access it

averevki: above is incorrect docstring then

averevki (approved):

jsmolar (approved):

PR #208, opened 2023-05-18
Add tests for Common feature - Metrics

### LINKED ISSUE #106: Add tests for Common feature: Metrics
**Summary**
Implement tests for Metrics. Consider test cases for identity, metadata, authorization, and response evaluator types.

**Documentation**
https://github.com/Kuadrant/authorino/blob/main/docs/features.md#common-feature-metrics-metrics
https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/observability.md#metrics

averevki:
Fix routes iteration method (**\_\_iter\_\_**).
Add real Openshift token acquiring possibility.
Add Authorino metrics endpoint getter (Preexisting and new Authorino).
Add Prometheus and Metrics clients for easier metrics fetching.
Add Authorino metrics and server-metrics tests.

Test cases:
- [X] Tests for integrity of /metrics endpoint
- [X] Tests for integrity of /server-metrics endpoint
- [X] Test for deep metrics:
    - [X] Identity evaluator metrics
    - [X] Authorization evaluator metrics
    - [X] Metadata evaluator metrics
    - [X] Response evaluator metrics

Closes #106

### DIFF
--- a/testsuite/objects/__init__.py
+++ b/testsuite/objects/__init__.py
@@ -135,6 +135,11 @@ class Authorino(LifecycleObject):
 
     @property
     @abc.abstractmethod
+    def metrics_service(self):
+        """Authorino metrics service name"""
+
+    @property
+    @abc.abstractmethod
     def authorization_url(self):
         """Authorization URL that can be plugged into envoy"""
 
@@ -147,13 +152,18 @@ class Authorino(LifecycleObject):
 class PreexistingAuthorino(Authorino):
     """Authorino which is already deployed prior to the testrun"""
 
-    def __init__(self, authorization_url, oidc_url) -> None:
+    def __init__(self, authorization_url, oidc_url, metrics_service) -> None:
         super().__init__()
         self._authorization_url = authorization_url
         self._oidc_url = oidc_url
+        self._metrics_service = metrics_service
 
     def wait_for_ready(self):
         return True
+
+    @property
+    def metrics_service(self):
+        return self._metrics_service
 
     @property
     def authorization_url(self):
--- a/testsuite/openshift/client.py
+++ b/testsuite/openshift/client.py
@@ -32,12 +32,12 @@ class OpenShiftClient:
     def __init__(self, project: str, api_url: str = None, token: str = None, kubeconfig_path: str = None):
         self._project = project
         self._api_url = api_url
-        self.token = token
+        self._token = token
         self._kubeconfig_path = kubeconfig_path
 
     def change_project(self, project) -> "OpenShiftClient":
         """Return new OpenShiftClient with a different project"""
-        return OpenShiftClient(project, self._api_url, self.token, self._kubeconfig_path)
+        return OpenShiftClient(project, self._api_url, self._token, self._kubeconfig_path)
 
     @cached_property
     def context(self):
@@ -46,7 +46,7 @@ class OpenShiftClient:
 
         context.project_name = self._project
         context.api_url = self._api_url
-        context.token = self.token
+        context.token = self._token
         context.kubeconfig_path = self._kubeconfig_path
 
         return context
@@ -56,6 +56,12 @@ class OpenShiftClient:
         """Returns real API url"""
         with self.context:
             return oc.whoami("--show-server=true")
+
+    @property
+    def token(self):
+        """Returns real OpenShift token"""
+        with self.context:
+            return oc.whoami("-t")
 
     @cached_property
     def apps_url(self):
--- a/testsuite/openshift/objects/metrics/prometheus.py
+++ b/testsuite/openshift/objects/metrics/prometheus.py
@@ -1,0 +1,91 @@
+"""Simple client for the OpenShift metrics"""
+from typing import Callable
+from datetime import datetime
+
+import httpx
+import backoff
+from apyproxy import ApyProxy
+
+
+def _params(key: str = "", labels: dict[str, str] = None) -> dict[str, str]:
+    """Generate metrics query parameters based on key and labels"""
+    if not labels:
+        return {"query": key}
+    # pylint: disable=consider-using-f-string
+    return {"query": "%s{%s}" % (key, ",".join(f"{k}='{v}'" for k, v in labels.items()))}
+
+
+class Metrics:
+    """Interface to the returned OpenShift metrics"""
+
+    def __init__(self, metrics):
+        self.metrics = metrics
+
+    def filter(self, func: Callable[[dict], bool]) -> "Metrics":
+        """Filter method accept function `func` as a single argument.
+        Given function will be used as a filter on metrics structure.
+        E.g. func = lambda x: x["metric"]["evaluator_name"] == 'json'
+        After the filtering, new Metrics object will be returned."""
+        return Metrics([m for m in self.metrics if func(m)])
+
+    @property
+    def names(self) -> list[str]:
+        """Return list of metrics names"""
+        return [m["metric"]["__name__"] for m in self.metrics]
+
+    @property
+    def values(self) -> list[float]:
+        """Return list of metrics values as floats"""
+        return [float(m["value"][1]) for m in self.metrics]
+
+
+class Prometheus:
+    """Interface to the OpenShift Prometheus client"""
+
+    def __init__(self, url: str, token: str, namespace: str = None):
+        self.token = token
+        self.headers = {"Authorization": f"Bearer {self.token}"}
+        self.namespace = namespace
+
+        self._client = httpx.Client(headers=self.headers, verify=False)
+        self.client = ApyProxy(url, self._client).api.v1
+
+    def get_targets(self) -> dict:
+        """Get active metric targets information"""
+        params = {"state": "active"}
+        response = self.client.targets.get(params=params)
+
+        return response.json()["data"]["activeTargets"]
+
+    def get_metrics(self, key: str = "", labels: dict[str, str] = None) -> Metrics:
+        """Get metrics by key or labels"""
+        if self.namespace:
+            labels = labels or {}
+            labels.setdefault("namespace", self.namespace)
+
+        params = _params(key, labels)
+        response = self.client.query.get(params=params)
+
+        return Metrics(response.json()["data"]["result"])
+
+    def wait_for_scrape(self, target_service: str, metrics_path: str = "/metrics"):
+        """Wait before next metrics scrape on service is finished"""
+        call_time = datetime.utcnow()
+
+        @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=10)
+        def _wait_for_scrape():
+            """Wait for new scrape after the function call time"""
+            for target in self.get_targets():
+                if (
+                    "service" in target["labels"].keys()
+                    and target["labels"]["service"] == target_service
+                    and target["discoveredLabels"]["__metrics_path__"] == metrics_path
+                ):
+                    return call_time < datetime.fromisoformat(target["lastScrape"][:26])
+            return False
+
+        _wait_for_scrape()
+
+    def close(self):
+        """Close httpx client connection"""
+        self._client.close()
--- a/testsuite/openshift/types/__init__.py
+++ b/testsuite/openshift/types/__init__.py
@@ -22,8 +22,15 @@ class RemoteMapping:
 
     def __iter__(self):
         """Return iterator for requested resource"""
-        data = self._client.do_action("get", self._resource_name, parse_output=True)
-        return iter(data["items"])
+        data = self._client.do_action(
+            "get",
+            self._resource_name,
+            "-o",
+            "json",
+            "--ignore-not-found=true",
+            parse_output=True,
+        )
+        return iter(data.as_dict()["items"])
 
     def __getitem__(self, name):
         """Return requested resource as APIObject"""
--- a/testsuite/tests/kuadrant/authorino/metrics/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/metrics/conftest.py
@@ -1,0 +1,56 @@
+"""Conftest for the Authorino metrics tests"""
+import pytest
+
+from testsuite.openshift.objects.metrics import ServiceMonitor, MetricsEndpoint, Prometheus
+
+
+@pytest.fixture(scope="module")
+def run_on_kuadrant():
+    """Kuadrant doesn't allow customization of Authorino parameters"""
+    return False
+
+
+@pytest.fixture(scope="session")
+def prometheus(request, openshift):
+    """
+    Return an instance of OpenShift metrics client
+    Skip tests if query route is not properly configured
+    """
+    # find thanos-querier route in the openshift-monitoring project
+    # this route allows to query metrics
+    openshift_monitoring = openshift.change_project("openshift-monitoring")
+    routes = openshift_monitoring.routes.for_service("thanos-querier")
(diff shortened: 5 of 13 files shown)


pehala (commented):
I think we should focus not on the endpoints but we should export it to Prometheus and check data there. @jsmolar WDYT?

jsmolar:
> I think we should focus not on the endpoints but we should export it to Prometheus and check data there. @jsmolar WDYT?

@pehala Why? Is there any specific reason for that? Information extracted from the endpoint should be the same as represented by Prometheus. Few tests for Prometheus certainly, but I don't see the reason why we should prioritize it.

pehala:
> > I think we should focus not on the endpoints but we should export it to Prometheus and check data there. @jsmolar WDYT?
> 
> @pehala Why? Is there any specific reason for that? Information extracted from the endpoint should be the same as represented by Prometheus. Few tests for Prometheus certainly, but I don't see the reason why we should prioritize it.

This endpoint contains data specific for Prometheus (if it wouldn't we wouldn't need prometheus-client) and as such is unlikely to be consumed directly.

jsmolar:
> > > I think we should focus not on the endpoints but we should export it to Prometheus and check data there. @jsmolar WDYT?
> > 
> > 
> > @pehala Why? Is there any specific reason for that? Information extracted from the endpoint should be the same as represented by Prometheus. Few tests for Prometheus certainly, but I don't see the reason why we should prioritize it.
> 
> This endpoint contains data specific for Prometheus (if it wouldn't we wouldn't need prometheus-client) and as such is unlikely to be consumed directly.

If the data is exclusively generated for Prometheus consumption and that is the intended use case.. Then yes, we should test it that way and leave only some relevant tests for the endpoint testing.

pehala:
@guicassolato To verify, do I understand it correctly that the metrics endpoint is meant to be consumed by Prometheus or another metric collector?

averevki:
I can confirm that the metrics are in the Prometheus metric format and probably are intended to be exported back to Prometheus as I found them hard to read in plain text.

testsuite/objects/__init__.py
```diff
@@ -150,14 +155,19 @@ def add_rule(self, when: list[Rule]):
...
         self._authorization_url = authorization_url
         self._oidc_url = oidc_url
+        self._metrics_url = metrics_url
 
     def wait_for_ready(self):
         return True
 
+    @property
+    def metrics_service(self):
+        return self._metrics_url.split(".", 1)[0]
```
pehala, line 169: if you need metrics url in specific format, why dont you request that format in the settings directly?

testsuite/openshift/client.py
```diff
@@ -32,12 +32,12 @@ class OpenShiftClient:
     def __init__(self, project: str, api_url: str = None, token: str = None, kubeconfig_path: str = None):
         self._project = project
         self._api_url = api_url
-        self.token = token
+        self._token = token
```
pehala, line 35: why?

testsuite/openshift/objects/metrics.py
```diff
@@ -0,0 +1,92 @@
...
+        response = self._request("/api/v1/query", params=params)
+        response.raise_for_status()
+        return response.json()["data"]["result"]
+
+    @staticmethod
+    def get_metrics_keys(metrics: list):
+        """Return set of given metrics names"""
+        return {m["metric"]["__name__"] for m in metrics}
+
+    def wait_for_scrape(self, target_service: str, metrics_path: str = "/metrics", scrape_after: datetime = None):
```
pehala, line 62: This method seems too long. Isn't there a better way how to way for scrape? My suggestion (not sure if it will work) is to see when the last scrape was made and wait until the time updates, through something like this https://groups.google.com/g/prometheus-users/c/vzbKGoJvzA8 ?

testsuite/openshift/types/__init__.py
```diff
@@ -22,8 +22,15 @@ def fetch_resource(self, name, *cmd_args: str, auto_raise: bool = True):
 
     def __iter__(self):
         """Return iterator for requested resource"""
-        data = self._client.do_action("get", self._resource_name, parse_output=True)
-        return iter(data["items"])
+        data = self._client.do_action(
+            "get",
+            self._resource_name,
+            "-o",
+            "json",
```
pehala, line 29: Why do you need it in json, it can be parsed in the yaml as well

pehala, line 29: Why do you need it in json, it can be parsed in the yaml as well

testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+            path,
+            {
+                "NAME": blame("sm"),
+                "LABEL": module_label,
+            },
+        )
+    request.addfinalizer(service_monitor_object.delete)
+
+    # wait for the new configured endpoints to became available for metrics scraping
+    metrics.wait_for_scrape(authorino.metrics_service.name(), scrape_after=datetime.utcnow() + timedelta(seconds=90))
```
pehala, line 48: you are using 90secs here, but the constant is for 30sec, why?

pehala (changes requested):
I am not sold on how the metrics tests are written. They seem overly complicated when looked on by someone who doesn't know prometheus.

testsuite/tests/kuadrant/authorino/metrics/conftest.py
averevki: There is some time before Prometheus starts to follow configured endpoints after creation of a new ServiceMonitor object. It is approximately 75 seconds. After, every scrape is performed every 30 seconds by default.

testsuite/openshift/types/__init__.py
averevki: It doesn't matter in the context. Anyway in the next line output will be converted to a python dictionary.

testsuite/openshift/types/__init__.py
```diff
@@ -22,8 +22,15 @@ def fetch_resource(self, name, *cmd_args: str, auto_raise: bool = True):
 
     def __iter__(self):
         """Return iterator for requested resource"""
-        data = self._client.do_action("get", self._resource_name, parse_output=True)
-        return iter(data["items"])
+        data = self._client.do_action(
```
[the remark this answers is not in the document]
averevki: I'll look, but I don't think this method is used anywhere as it was not working. Response wasn't converted to data structure and method was showing a `data` key error.

testsuite/objects/__init__.py
averevki: It is an option. The point was to request all services in the same format so that it won't be a little confusing... https://github.com/Kuadrant/testsuite/blob/c76fcea6d8501c0843e1b23e562dfbca7531b08d/config/settings.local.yaml.tpl#L35-L37

pehala: What is the expected format?

averevki: Service URL look like this `<service>.<pod_namespace>.svc.cluster.local`. I need only a service name to label this service properly so that *ServiceMonitor* can discover it.

testsuite/openshift/client.py
averevki: This token is `NoneType` most of the time. An openshift token is needed to request Prometheus metrics. New property was introduced to fetch the real token. https://github.com/Kuadrant/testsuite/blob/c76fcea6d8501c0843e1b23e562dfbca7531b08d/testsuite/openshift/client.py#L60-L64

testsuite/openshift/objects/metrics.py
averevki: I'll look into this. But I don't think that method is that long when you think about all input parameters that should be respected. It is an optimized way. At the beginning this method was just a `time.wait` one-liner, but in this case it will be waiting 30 seconds for every scrape (which won't be that time consuming for now, as method for scrape waiting is called only 2 times).

testsuite/openshift/objects/metrics.py
```diff
@@ -0,0 +1,92 @@
...
+            labels.setdefault("namespace", self.namespace)
+
+        params = _params(key, labels)
+
+        response = self._request("/api/v1/query", params=params)
+        response.raise_for_status()
+        return response.json()["data"]["result"]
+
+    @staticmethod
+    def get_metrics_keys(metrics: list):
```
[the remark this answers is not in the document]
averevki: Simple utility to get a set of names of the returned from `get_metrics` method metrics. Those names are later used to check the integrity of the metrics endpoints.

testsuite/openshift/objects/metrics.py
```diff
@@ -0,0 +1,92 @@
...
+    def get_targets(self) -> dict:
+        """Get active metric targets information"""
+
+        params = {"state": "active"}
+
+        response = self._request("/api/v1/targets", params=params)
+        response.raise_for_status()
+        return response.json()["data"]["activeTargets"]
+
+    def get_metrics(self, key: str = "", labels: dict[str, str] = None) -> list:
```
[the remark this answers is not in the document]
averevki: A list with dictionary of metric parameters. Example:  `{'metric': {'__name__': 'auth_server_authconfig_duration_seconds_bucket', 'authconfig': 'authconfig-authorino-testing', 'endpoint': 'http', 'exported_namespace': 'authorino-testing', 'job': 'authorino-testing-job-controller-metrics', 'le': '+Inf', 'namespace': 'authorino-testing', 'pod': 'authorino-testing-pod', 'prometheus': 'openshift-user-workload-monitoring/user-workload', 'service': 'authorino-testing-controller-metrics'}, 'value': [1688048221.512, '1']}`

testsuite/objects/__init__.py
pehala: Why not ask only for service_name and name the variable appropriately, not naming it "URL"

testsuite/openshift/objects/metrics.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Prometheus:
+    """Interface to the OpenShift Prometheus client"""
+
+    def __init__(self, url: str, token: str, namespace: str = None):
+        self.client = ApyProxy(url, httpx.Client(verify=False)).api.v1
+
+        self.token = token
+        self.headers = {"Authorization": f"Bearer {self.token}"}
```
pehala, line 50: you can pass this directly into the httpx.Client

testsuite/openshift/objects/metrics.py
```diff
@@ -0,0 +1,87 @@
...
+    def values(self) -> list[float]:
+        """Return list of metrics values as floats"""
+        return [float(m["value"][1]) for m in self.metrics]
+
+
+class Prometheus:
+    """Interface to the OpenShift Prometheus client"""
+
+    def __init__(self, url: str, token: str, namespace: str = None):
+        self.client = ApyProxy(url, httpx.Client(verify=False)).api.v1
```
pehala, line 47: In an ideal scenario, `httpx.Client` should be closed

testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+def label_metrics_service(authorino, module_label):
+    """Label Authorino controller-metrics service for the proper discovery"""
+    authorino.metrics_service.label({"app": module_label})
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module", autouse=True)
+def create_service_monitor(request, openshift, authorino, blame, module_label, prometheus, label_metrics_service):
+    """Create ServiceMonitor object to follow Authorino /metrics and /server-metrics endpoints"""
+    with resources.path("testsuite.resources", "service-monitor.yaml") as path:
```
pehala, line 37: While this is not neccessarily wrong. You could create an OpenshiftObject descendant for ServiceMonitor instead

testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+            path,
+            {
+                "NAME": blame("sm"),
+                "LABEL": module_label,
+            },
+        )
+    request.addfinalizer(service_monitor_object.delete)
+
+    # wait for the new configured endpoints to became available for metrics scraping
+    time.sleep(90)
```
pehala, line 48: Shouldnt just wait for scrape suffice?

averevki: After a few tests, the wait here appears to be unnecessary at all, since every test now uses its own wait on scrape for the tested metric using backoff. Removed.

testsuite/tests/kuadrant/authorino/metrics/conftest.py
averevki: The new Service Monitor object resides under `testsuite/openshift/objects/metrics/service_monitor.py`

testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module", autouse=True)
+def create_service_monitor(request, openshift, authorino, blame, module_label, prometheus, label_metrics_service):
+    """Create ServiceMonitor object to follow Authorino /metrics and /server-metrics endpoints"""
+    endpoints = [MetricsEndpoint("/metrics", "http"), MetricsEndpoint("/server-metrics", "http")]
+    service_monitor = ServiceMonitor.create_instance(openshift, blame("sm"), endpoints, labels={"app": module_label})
+    service_monitor.commit()
+
+    request.addfinalizer(service_monitor.delete)
```
pehala, line 40: finalizer should be before the commit, in this case probably yield suffices

testsuite/openshift/objects/metrics/prometheus.py
```diff
@@ -0,0 +1,91 @@
...
+                    "service" in target["labels"].keys()
+                    and target["labels"]["service"] == target_service
+                    and target["discoveredLabels"]["__metrics_path__"] == metrics_path
+                ):
+                    return call_time < datetime.fromisoformat(target["lastScrape"][:26])
+            return False
+
+        _wait_for_scrape()
+
+    def close(self):
```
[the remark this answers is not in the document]
averevki: I don't think that LifecycleObject is suitable for this case. I think that LifecycleObject is intended to be used with Openshift objects as it specifies `commit` and `delete` methods. Delete name for the `close client` method is not right, in my opinion. (in addition `commit` method would be empty)

pehala (dismissed):

testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.fixture(scope="module")
+def label_metrics_service(authorino, module_label):
+    """Label Authorino controller-metrics service for the proper discovery"""
+    authorino.metrics_service.label({"app": module_label})
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module", autouse=True)
+def create_service_monitor(openshift, blame, module_label, prometheus, label_metrics_service):
```
jsmolar, line 34: Do you need `promethus` here?

jsmolar:
@averevki please add skip for Kuadrant

jsmolar (approved):

pehala (approved):

PR #209, opened 2023-05-22
Add RLP tests with authz

jsmolar:


testsuite/openshift/objects/rate_limit.py
```diff
@@ -19,20 +20,43 @@ def create_instance(cls, openshift: OpenShiftClient, name, route: Referencable,
...
+                            "path": [{"segment": {"key": "ext_auth_data"}}, {"segment": {"key": user_id}}],
+                        },
+                    }
+                }
+            ]
+        }
+
     @modify
-    def add_limit(self, max_value, seconds, conditions: list[str] = None):
+    def add_limit(self, max_value, seconds, conditions: list[str] = None, variable=None):
```
pehala, line 48: I would change variable for something more verbose, like `enable_auth`

jsmolar: Do you want to use `enable_auth` like bool value here?

mganisin: I see that `variable` is element of `variables` list, but then it is "called" also `user_id` or `descriptor_key`. Moreover from the code I understand that "userID" is currently only possible value, I can't imagine current code would work with different value. So without deep knowledge I would say there should be better name and maybe for future "simple" value (str, bool) doesn't have to be enough.

testsuite/objects/sections.py
```diff
@@ -38,6 +38,10 @@ class Identities(abc.ABC):
     def oidc(self, name, endpoint, credentials, selector, **common_features):
         """Adds OIDC identity provider"""
 
+    @abc.abstractmethod
+    def kubernetes(self, name: str, *audiences):
```
averevki, line 42: Is this somehow relevant to this PR?

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -64,6 +64,15 @@ def mtls(self, name: str, selector_key: str, selector_value: str, **common_featu
         """
         self.add_item(name, {"mtls": {"selector": {"matchLabels": {selector_key: selector_value}}}}, **common_features)
 
+    @modify
+    def kubernetes(self, name: str, *audiences):
```
averevki, line 68: Is this somehow relevant to this PR?

testsuite/tests/kuadrant/test_rate_limit_authz.py
```diff
@@ -0,0 +1,52 @@
+"""
```
averevki, line 1: I believe this test file should be under the limitador package

testsuite/tests/kuadrant/test_rate_limit_authz.py
```diff
@@ -0,0 +1,52 @@
...
+@pytest.fixture(scope="module")
+def auth2(rhsso):
+    """Creates new RHSSO user and returns its authentication object for HTTPX"""
+    user = rhsso.realm.create_user("user2", "password", email="test@test.com")
+    return HttpxOidcClientAuth.from_user(rhsso.get_token, user=user)
+
+
+def test_multiple_iterations(client, auth, auth2):
+    """Tests that simple limit is applied successfully and works for multiple iterations"""
+    assert client.get("/get", auth=auth).status_code == 200
```
averevki, line 49: Can these requests be sent through the `utilities.fire_requests` utility instead?

testsuite/objects/sections.py
jsmolar: No, this should be in a different PR.

testsuite/tests/kuadrant/test_rate_limit_authz.py
jsmolar: It is correctly placed under the Kuadrant package. The reason is that it uses both Limitador (RLP) and Authorino (OIDC identity) features.

pehala (dismissed):

averevki (approved):
Can suggest to squash commits :+1:

testsuite/tests/kuadrant/test_rate_limit_authz.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.fixture(scope="module")
+def auth(oidc_provider):
+    """Returns RHSSO authentication object for HTTPX"""
+    return HttpxOidcClientAuth(oidc_provider.get_token, "authorization")
+
+
+@pytest.fixture(scope="module")
+def auth2(rhsso):
+    """Creates new RHSSO user and returns its authentication object for HTTPX"""
+    user = rhsso.realm.create_user("user2", "password", email="test@test.com")
```
averevki, line 35: In my understanding second user's email should somehow derive from the first user's mail because this test requires 2 different emails for identification e.g. `f"second-{first_user_mail}"`. `test@test.com` easily can be the first email also and test would fail

averevki (dismissed):

testsuite/openshift/objects/rate_limit.py
```diff
@@ -39,13 +39,15 @@ def create_instance(cls, openshift: OpenShiftClient, name, route: Referencable,
         return cls(model, context=openshift.context)
 
     @modify
-    def add_limit(self, name, limits: Iterable[Limit], when: Iterable[Rule] = None):
+    def add_limit(self, name, limits: Iterable[Limit], when: Iterable[Rule] = None, counters=None):
```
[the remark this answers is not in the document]
jsmolar: https://docs.kuadrant.io/kuadrant-operator/doc/reference/ratelimitpolicy/ and see "counters"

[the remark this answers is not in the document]
averevki: more about this:
```suggestion
    def add_limit(self, name, limits: Iterable[Limit], when: Iterable[Rule] = None, counters: list[str] = None):
```

averevki (approved):

pehala (approved):

PR #211, opened 2023-05-23
Remove Authorization interface

pehala:
- AuthPolicy will always contain the entire AuthConfig, so there is no need to maintain both interface and implementation.


testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -65,7 +65,7 @@ def authorino_parameters(authorino_parameters, specific_authorino_name):
 
 # pylint: disable=unused-argument
 @pytest.fixture(scope="module")
-def authorization(authorization, openshift, module_label, authorino_domain) -> Authorization:
+def authorization(authorization, openshift, module_label, authorino_domain):
```
mganisin, line 68: is this (and the other one) loss of type annotation of return intentional?

testsuite/objects/sections.py
```diff
@@ -1,91 +0,0 @@
-"""Contains implementation for all AuthConfig sections"""
```
averevki, line 1: This is an empty file now

averevki (commented):
Seems like a good change 👍

mganisin (approved):

averevki (approved):

PR #214, opened 2023-06-02
Add simple response

pehala:
- Add simple debugging response - `response.add_simple`
  - Used when you need to return some authjson section and do not care about details 
- Refactor extract_from_response
  - Uses weakget
  - same default values as `response.add_simple`

### DIFF
--- a/testsuite/openshift/objects/auth_config/sections.py
+++ b/testsuite/openshift/objects/auth_config/sections.py
@@ -161,6 +161,9 @@ class MetadataSection(Section, Metadata):
 class ResponsesSection(Section, Responses):
     """Section which contains response configuration"""
 
+    def add_simple(self, auth_json, name="simple", key="data", **common_features):
+        self.add({"name": name, "json": {"properties": [{"name": key, "valueFrom": {"authJSON": auth_json}}]}})
+
     @modify
     def add(self, response, **common_features):
         """Adds response section to AuthConfig."""
--- a/testsuite/tests/kuadrant/authorino/caching/metadata/test_caching.py
+++ b/testsuite/tests/kuadrant/authorino/caching/metadata/test_caching.py
@@ -4,7 +4,7 @@ from time import sleep
 import pytest
 
 from testsuite.objects import Cache, Value
-from testsuite.utils import extract_from_response
+from testsuite.utils import extract_response
 
 
 @pytest.fixture(scope="module")
@@ -28,10 +28,12 @@ def test_cached(client, auth, module_label, mockserver):
         - only single external value evaluation occurs. The second response contains cached (in-memory) value
     """
     response = client.get("/get", auth=auth)
-    data = extract_from_response(response, module_label, "uuid")
+    data = extract_response(response)[module_label]["uuid"] % None
     response = client.get("/get", auth=auth)
-    cached_data = extract_from_response(response, module_label, "uuid")
+    cached_data = extract_response(response)[module_label]["uuid"] % None
 
+    assert cached_data is not None
+    assert data is not None
     assert data == cached_data
     assert len(mockserver.retrieve_requests(module_label)) == 1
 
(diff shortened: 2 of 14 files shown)


testsuite/tests/kuadrant/authorino/caching/metadata/test_caching.py
```diff
@@ -28,21 +28,23 @@ def test_cached(client, auth, module_label, mockserver):
         - only single external value evaluation occurs. The second response contains cached (in-memory) value
     """
     response = client.get("/get", auth=auth)
-    data = extract_from_response(response, module_label, "uuid")
+    data = extract_response(response)[module_label]["uuid"] % None
     response = client.get("/get", auth=auth)
-    cached_data = extract_from_response(response, module_label, "uuid")
+    cached_data = extract_response(response)[module_label]["uuid"] % None
 
+    assert data is not None
```
azgabur, line 35: Why this is the only place with `is not None` check?

pehala: Because two values are compared to each other, so I added None check to each of them

mganisin (approved):

testsuite/tests/kuadrant/authorino/caching/metadata/test_caching.py
```diff
@@ -28,21 +28,23 @@ def test_cached(client, auth, module_label, mockserver):
...
     """Tests that cached value expires after ttl"""
     response = client.get("/get", auth=auth)
-    data = extract_from_response(response, module_label, "uuid")
+    data = extract_response(response)[module_label]["uuid"] % None
     sleep(cache_ttl)
     response = client.get("/get", auth=auth)
-    cached_data = extract_from_response(response, module_label, "uuid")
+    cached_data = extract_response(response)[module_label]["uuid"] % None
 
     assert data != cached_data
```
averevki, line 49: Assert if `data` and `cached_data` are `not None`, same as in the previous test-case

testsuite/tests/kuadrant/authorino/caching/metadata/test_not_cached.py
```diff
@@ -14,8 +14,8 @@ def authorization(authorization, module_label, expectation_path):
 def test_no_caching(client, auth, module_label, mockserver):
     """Tests value is not cached for metadata without caching feature"""
     response = client.get("/get", auth=auth)
-    data = extract_from_response(response, module_label, "uuid")
+    data = extract_response(response)[module_label]["uuid"] % None
     response = client.get("/get", auth=auth)
 
-    assert extract_from_response(response, module_label, "uuid") != data
+    assert extract_response(response)[module_label]["uuid"] % None != data
```
averevki, line 20: Assert if returned uuid are `not None`

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -161,6 +161,9 @@ def uma_metadata(self, name, endpoint, credentials, **common_features):
 class ResponsesSection(Section, Responses):
     """Section which contains response configuration"""
 
+    def add_simple(self, auth_json, name="simple", key="data", **common_features):
```
averevki, line 164: Is it possible to also change response to simple in the `test_response_condition.py` authorization fixture? As simple response also supports conditional common feature?

pehala: Response module has custom fixture for manipulating responses. The method is mesnt more for tests that use response as a tool instead of a actual thing to test

testsuite/tests/kuadrant/authorino/caching/metadata/test_caching.py
mganisin:
```
>>> 1 == 1 != None
True
>>> 1 == 2 != None
False
>>> 1 == None != None
False
>>> None == None != None
False
```
Maybe style checkers will complain (I don't know), anyway worth method

pehala: Yep, it won't pass pylint. I should be using `is None` or I am using bad-chained-comparison

pehala: I added null check to everything, they could be probably squished to less lines but I think this is the most readable option

averevki (approved):

testsuite/tests/kuadrant/authorino/conditions/section_conditions/test_response_condition.py
```diff
@@ -31,10 +31,10 @@ def test_skip_response(client, auth):
     assert response.status_code == 200
 
     # verify that response was not returned on a GET request
-    with pytest.raises(KeyError, match="Auth-Json"):
-        extract_from_response(response)
+    assert "simple" not in response.json()["headers"]
```
azgabur, line 34: Found a bug. This string should be `Simple` because HTTP headers capitalize first letter of header name. As this PR is already merged, can you create a new small one with this fix @pehala ?

PR #219, opened 2023-06-21
Add clean target deleting poetry env

mdujava:


averevki (commented):
Clean target and some others were removed during the migration(https://github.com/Kuadrant/testsuite/pull/147) to leave the user to manage his own environments. Is this needed now? Thoughts @pehala?

pehala:
Not sure about this. Would it be possible for us to create and/or delete a specific env that we know was created by the makefile? Meaning delete would only delete the correct env and not mess up users own envs.

mdujava:
I think yes, if  `make poetry[-dev]` would install the env and save the name into the `.make-poetry-sync` then clean could only delete that env. WDYT?

mganisin:
And what if make created and used one specific env name? (unless env= said otherwise)

pehala:
Closing this, as Poetry env management in Makefile deserves a rework as whole

PR #213, opened 2023-06-02
Add ResponseProperties object and example refactor

azgabur:
Based on commit b042b30f63e8c1040de7138bc13469217fa0493d in #212 

Usage of dynamic response function is quite common in tests to expose and check Authorino internal variables. I propose creating new object for easier definition of this function without the need to construct complicated dictionaries in tests.

**Contains fix of Value class to dataclass.**

This PR also contains example refactor of https://github.com/Kuadrant/testsuite/tree/main/testsuite/tests/kuadrant/authorino/response tests.

PR #212, opened 2023-05-30
Add tests for extended_properties feature

azgabur:
- [x] `extended_properties` functionality implementation (contains fix of `Value` class)
- [x] Test cases:
    - [x] Token normalization
    - [x] Other extended_properties tests
    - [x] Overwrite tests
 
Contains also:

-  bugfix of "Add plain identity" function https://github.com/Kuadrant/testsuite/pull/212#discussion_r1272674382
- Replace Union with more native type marking https://github.com/Kuadrant/testsuite/pull/212#discussion_r1281998698
- ~Replace custom `asdict()` in `ValueFrom` with dict transformation https://github.com/Kuadrant/testsuite/pull/212#discussion_r1282710508~ 
- Fix typo in docstring for `asdict()` https://github.com/Kuadrant/testsuite/pull/212#discussion_r1285922191
- Change `add_item` function to force keyword-only common_features arguments https://github.com/Kuadrant/testsuite/pull/212#discussion_r1305881457

Closes #108

PR #220, opened 2023-07-12
Reformat AuthConfig responses section

averevki:
Reformat AuthConfig responses section so that response can be added same as the elements of other sections.
New dataclass `Property` was introduced. Property is a `Value` with name.

Example of adding different types of reponses:
```
authorization.responses.json("header", [Property("anything", Value("one"))])
authorization.responses.plain("header", ValueFrom("context.request.http.method"))
authorization.responses.wristband("wristband", wristband_endpoint, wristband_secret, wrapper="envoyDynamicMetadata")
```


testsuite/objects/__init__.py
azgabur, line 1: I recommend rebasing on c7b495b8ad21636e5423927f4add94fd8f58fa54 (still in PR #213) as it already implements Responses object and many improvements/additions to other objects.

testsuite/tests/kuadrant/authorino/conditions/section_conditions/test_response_condition.py
```diff
@@ -1,23 +1,16 @@
 """Test condition to skip the response section of AuthConfig"""
 import pytest
 
+from testsuite.objects import Property
```
azgabur, line 4: `from testsuite.objects import Rule, Property`

testsuite/tests/kuadrant/authorino/operator/http/conftest.py
```diff
@@ -1,6 +1,7 @@
 """Conftest for all tests requiring custom deployment of Authorino"""
 import pytest
 
+from testsuite.objects import Property
```
azgabur, line 4: `from testsuite.objects import Authorization, Property`

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -170,13 +180,59 @@ def uma_metadata(self, name, endpoint, credentials, **common_features):
 class ResponsesSection(Section, Responses):
     """Section which contains response configuration"""
 
+    def add(
+        self,
+        name: str,
+        value: dict,
+        wrapper_key: str = None,
+        wrapper: Literal["httpHeader", "envoyDynamicMetadata"] = None,
+        **common_features
```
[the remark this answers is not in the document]
averevki: Keyword arguments here can be placed anywhere after the positional arguments. They are not dependent on their order as the new private method `add` is introduced in the `ResponsesSection` class. This method is used in every* other response section method instead of direct `add_item` inheritance.

averevki:
rebase^

averevki:
rebase^

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -175,20 +177,59 @@ def uma_metadata(self, name, endpoint, credentials, **common_features):
...
+                },
+            },
+            **common_features
         )
 
     @modify
-    def add(self, response, **common_features):
-        """Adds response section to AuthConfig."""
-        self.add_item(response.pop("name"), response, **common_features)
+    def remove_all(self):
```
pehala, line 230: I would stick with the clear() method

testsuite/objects/__init__.py
```diff
@@ -97,6 +97,18 @@ def asdict(self):
...
+@dataclass
+class Property:
+    """Dataclass for static and dynamic values. Property is a Value with name."""
+
+    name: str
+    value: ABCValue
+
+    def asdict(self):
+        """Override `asdict` function"""
+        return {"name": self.name, **asdict(self.value)}
```
mganisin, line 109: this is not necessary with custom asdict function anymore, is it?

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -175,20 +177,59 @@ def uma_metadata(self, name, endpoint, credentials, **common_features):
...
+        self,
+        name: str,
+        value: dict,
+        wrapper_key: str = None,
+        wrapper: Literal["httpHeader", "envoyDynamicMetadata"] = None,
+        **common_features
+    ):
+        """Add response to AuthConfig"""
+        value["wrapper"] = wrapper
+        value["wrapperKey"] = wrapper_key
```
mganisin, line 190: is this supposed to be or safe to set it everywhere? Every object will have it set to None (at least). I see that this is the purpose of this function, anyway it looks odd.

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -175,20 +177,59 @@ def uma_metadata(self, name, endpoint, credentials, **common_features):
...
+
+        self.add_item(name, value, **common_features)
+
+    def add_simple(self, auth_json: str, name="simple", key="data", **common_features):
+        """Add simple response to AuthConfig"""
+        self.json(name, [Property(key, ValueFrom(auth_json))], **common_features)
+
+    @modify
+    def json(self, name: str, properties: list[Property], **common_features):
```
mganisin, line 199: My recommendation is to use `add_` in the names of these new functions. I see that this proposed form follows existing convention, in my opinion this short prefix has clear benefits. This is for discussion and broader agreement though (I am sorry if that already happened)

testsuite/openshift/objects/__init__.py
```diff
@@ -46,4 +46,6 @@ def commit(self):
 
     def delete(self, ignore_not_found=True, cmd_args=None):
         """Deletes the resource, by default ignored not found"""
-        return super().delete(ignore_not_found, cmd_args)
+        deleted = super().delete(ignore_not_found, cmd_args)
+        self.committed = False
```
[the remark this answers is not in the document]
averevki: I'll add a check, but I don't think it is necessary here, since the function will raise an error on incorrect deletion and stop the program execution either way.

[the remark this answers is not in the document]
pehala: You are not wrong, but that is only true if you delete it and then try to commit it again and that seems to be a weird use case

testsuite/objects/__init__.py
averevki: @mganisin another possibility how to implement these properties - https://github.com/Kuadrant/testsuite/pull/212/files#diff-25923b13ccdf4676496d71da370f7c9ccd254043ebe12c08c0faa40e7e32be2cR74-R86

testsuite/openshift/objects/auth_config/sections.py
averevki: I believe we decided to change the names of these functions in another PR. You are right that for now it just follows an existing convention

testsuite/openshift/objects/auth_config/sections.py
averevki: Name refactoring should be a part of the feature PR 👍

testsuite/openshift/objects/auth_config/sections.py
pehala: I would rather see the if, just to be sure

pehala (approved):

jsmolar (approved):

PR #221, opened 2023-07-25
Dataclasses refactor and add new to_dict function

azgabur:
This PR includes:
 - Refactor of Value class to three dataclasses removing redundant test of exclusivity.
 - Replacing usage of `dataclasses.asdict()` and direct `object.to_dict()` method calls in favor of new `to_dict()` function

### DIFF
--- a/testsuite/objects/__init__.py
+++ b/testsuite/objects/__init__.py
@@ -1,10 +1,49 @@
 """Module containing base classes for common objects"""
 import abc
-from dataclasses import dataclass
+from dataclasses import dataclass, is_dataclass, fields
+from copy import deepcopy
 from functools import cached_property
-from typing import Literal, List
+from typing import Literal, Union
 
 from testsuite.objects.sections import Metadata, Identities, Authorizations, Responses
+
+JSONValues = Union[None, str, int, bool, list["JSONValues"], dict[str, "JSONValues"]]
+
+
+def asdict(obj) -> dict[str, JSONValues]:
+    """
+    This function converts dataclass object to dictionary.
+    While it works similar to `dataclasses.asdict` a notable change is usage of
+    overriding `to_dict()` function if dataclass contains it.
+    This function works recursively in lists, tuples and dicts. All other values are passed to copy.deepcopy function.
+    """
+    if not is_dataclass(obj):
+        raise TypeError("asdict() should be called on dataclass instances")
+    return _asdict_recurse(obj)
+
+
+def _asdict_recurse(obj):
+    if hasattr(obj, "asdict"):
+        return obj.asdict()
+
+    if not is_dataclass(obj):
+        return deepcopy(obj)
+
+    result = {}
+    for field in fields(obj):
+        value = getattr(obj, field.name)
+        if value is None:
+            continue  # do not include None values
+
+        if is_dataclass(value):
+            result[field.name] = _asdict_recurse(value)
+        elif isinstance(value, (list, tuple)):
+            result[field.name] = type(value)(_asdict_recurse(i) for i in value)
+        elif isinstance(value, dict):
+            result[field.name] = type(value)((_asdict_recurse(k), _asdict_recurse(v)) for k, v in value.items())
+        else:
+            result[field.name] = deepcopy(value)
+    return result
 
 
 @dataclass
@@ -15,7 +54,7 @@ class MatchExpression:
     """
 
     operator: Literal["In", "NotIn", "Exists", "DoesNotExist"]
-    values: List[str]
+    values: list[str]
     key: str = "group"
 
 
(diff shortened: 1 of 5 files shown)


testsuite/tests/kuadrant/authorino/caching/metadata/test_caching.py
```diff
@@ -16,7 +16,7 @@ def cache_ttl():
 @pytest.fixture(scope="module")
 def authorization(authorization, module_label, expectation_path, cache_ttl):
     """Adds Cached Metadata to the AuthConfig"""
-    meta_cache = Cache(cache_ttl, Value(jsonPath="context.request.http.path"))
+    meta_cache = Cache(cache_ttl, ValueDynamic(valueFrom="context.request.http.path"))
```
averevki, line 19: As these `Value` dataclasses require only one value and class names are self-explanatory, I'd use positional arguments on instances.

testsuite/objects/__init__.py
```diff
@@ -1,12 +1,44 @@
...
+            if is_dataclass(value):
+                result[field.name] = to_dict(value)
+            elif isinstance(value, (list, tuple)):
+                result[field.name] = type(value)(to_dict(i) for i in value)
+            elif isinstance(value, dict):
+                result[field.name] = type(value)((to_dict(k), to_dict(v)) for k, v in value.items())
+            else:
+                result[field.name] = to_dict(value)
+    else:
+        result = deepcopy(obj)
```
averevki, line 37: I don't think that `to_dict` function should return an object deepcopy in any case. @mganisin WYT?

testsuite/objects/__init__.py
```diff
@@ -35,40 +67,45 @@ class Rule:
...
-        if not (value is None) ^ (jsonPath is None):
-            raise AttributeError("Exactly one of the `value` and `jsonPath` argument must be specified")
-        self.value = value
-        self.jsonPath = jsonPath
+
+@dataclass
+class ValueStatic(ValueBase):
+    """Dataclass for static Value. Can be anything, eg. string, integer, list, dict..."""
+
+    value: Any
```
averevki, line 82: `Any` type seems too broad. I'd rather `Union`'d all possible types.

testsuite/objects/__init__.py
```diff
@@ -1,12 +1,44 @@
...
+    result = {}
+    if is_dataclass(obj):
+        for field in fields(obj):
+            value = getattr(obj, field.name)
+            if value is None:
+                continue  # do not include None values
+
+            if is_dataclass(value):
+                result[field.name] = to_dict(value)
+            elif isinstance(value, (list, tuple)):
```
averevki, line 30:
```suggestion
            elif isinstance(value, Sequence):
```
Is this possible?

testsuite/objects/__init__.py
mganisin: I see the point. Anyway `Union` would be a) HUUUUGE b) something would be forgotten (I bet on `bytes`)

testsuite/objects/__init__.py
```diff
@@ -1,12 +1,44 @@
 """Module containing base classes for common objects"""
 import abc
-from dataclasses import dataclass
+from dataclasses import dataclass, is_dataclass, fields
+from copy import deepcopy
 from functools import cached_property
-from typing import Literal, List
+from typing import Literal, List, Any
 
 from testsuite.objects.sections import Metadata, Identities, Authorizations, Responses
```
mganisin, line 9: I would also consider whether is is necessary require docstrings here. It is good idea to have reference to struct from go code in docstring, however I feel quite a few classes and functions/methods won't require any documentation in this module.

maybe couple of extra global pylint disables (invalid-name, missing-docstrings) make sense

testsuite/objects/__init__.py
mganisin: As mentioned in other comment all this code is unnecessary in my opinion. Generally speaking this function resembles `dataclasses.asdict` therefore it should have same behavior. So `help(dataclasses.asdict)` would be authoritative answer if this should exist

testsuite/objects/__init__.py
mganisin: I see the point. Anyway `Union` would be a) HUUUUGE b) something would be forgotten (I bet on `bytes`)

testsuite/objects/__init__.py
```diff
@@ -35,40 +67,45 @@ class Rule:
...
+
+@dataclass
+class ValueStatic(ValueBase):
+    """Dataclass for static Value. Can be anything, eg. string, integer, list, dict..."""
+
+    value: Any
+
+
+@dataclass
+class ValueDynamic(ValueBase):
```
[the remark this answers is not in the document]
azgabur: I am not convinced creating too many classes especially when AuthJson is always (afaik) used with ValueFrom in structure: `{"valueFrom": {"authJSON": <str>}}` 
Maybe in future if it was used also in other places I would be convinced to make it a separate class. For now I dont think its needed and so it would reduce number of imports in creating Value classes and future classes using the Value.

Renaming of variable and class names is a good idea and I did it.

testsuite/objects/__init__.py
```diff
@@ -1,12 +1,44 @@
...
+    """
+    This function converts dataclass object to dictionary.
+    While it works similar to `dataclasses.asdict` a notable change is usage of
+    overriding `to_dict()` function if dataclass contains it.
+    This function works recursively in lists, tuples and dicts. All other values are passed to copy.deepcopy function.
+    """
+    if hasattr(obj, "to_dict"):
+        return obj.to_dict()
+
+    result = {}
```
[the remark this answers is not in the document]
mganisin: True, true. I forgot the recursion is needed

testsuite/objects/__init__.py
pehala: I am ok with `Any`, but if you want to have all types in Union I would prefer custom type called `JSONValues`, which would contain the Union and then only reference this new type in Value

azgabur: I created new type Union

pehala (approved):

testsuite/objects/__init__.py
```diff
@@ -1,11 +1,50 @@
 """Module containing base classes for common objects"""
 import abc
-from dataclasses import dataclass
+from dataclasses import dataclass, is_dataclass, fields
+from copy import deepcopy
 from functools import cached_property
-from typing import Literal, List
+from typing import Literal, List, Dict, Union
```
averevki, line 6: You can use generic types for `list` and `dict` instead (https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections).

testsuite/objects/__init__.py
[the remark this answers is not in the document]
averevki: > I am not convinced creating too many classes especially when AuthJson is always (afaik) used with ValueFrom in structure: `{"valueFrom": {"authJSON": <str>}}` Maybe in future if it was used also in other places I would be convinced to make it a separate class. For now I dont think its needed and so it would reduce number of imports in creating Value classes and future classes using the Value.

Agreed

pehala (approved):

testsuite/objects/__init__.py
azgabur: Removed types `List, Dict `and replaced by `list, dict`

mganisin (approved):

PR #224, opened 2023-08-02
Bump python to 3.11 in acceptance.yml

azgabur:
To fix failing test in #212

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -14,7 +14,7 @@ jobs:
         id: setup-python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.11'
           cache: "poetry"
           cache-dependency-path: "pyproject.toml"
       - name: Install dependencies
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ Bind kubeconfig to /opt/kubeconfig \
 Bind a dir to /test-run-results to get reports "
 
 RUN useradd --no-log-init -u 1001 -g root -m testsuite
-RUN dnf install -y python3 pip make git && dnf clean all
+RUN dnf install -y python311 pip make git && dnf clean all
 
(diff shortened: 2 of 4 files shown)


pehala:
IF you want to bump the minimum required python version to 3.10 you would need to bump it in more places like README or Dockerfile

azgabur:
I updated README and Dockerfile. Bumping one more version to 3.11 as previous version was not available in centos9 repositories.

pehala (dismissed):

averevki:
You probably should also change it in `pyproject.toml`. 3.11 is Python's lastest version and still some packages may not support it. Although there aren't many now, I think you should still check for testsuite packages compatibility. I found at least one package, `backoff`, that is not supporting Python 3.11 explicitly and is frequently used in tests (https://pypi.org/project/backoff/). I also found a Python 3.11 compatibility table with most popular packages https://pyreadiness.org/3.11/

pehala:
> You probably should also change it in `pyproject.toml`. 3.11 is Python's lastest version and still some packages may not support it. Although there aren't many now, I think you should still check for testsuite packages compatibility. I found at least one package, `backoff`, that is not supporting Python 3.11 explicitly and is frequently used in tests (https://pypi.org/project/backoff/). I also found a Python 3.11 compatibility table with most popular packages https://pyreadiness.org/3.11/

Does not support or does not work?

averevki:
> > You probably should also change it in `pyproject.toml`. 3.11 is Python's lastest version and still some packages may not support it. Although there aren't many now, I think you should still check for testsuite packages compatibility. I found at least one package, `backoff`, that is not supporting Python 3.11 explicitly and is frequently used in tests (https://pypi.org/project/backoff/). I also found a Python 3.11 compatibility table with most popular packages https://pyreadiness.org/3.11/
> 
> Does not support or does not work?

3.11 is not listed on pypi package page. So I think it should at least be tested.

pehala:
I have been using 3.11 for some time, so I do not expect any problems, feel free to run the entire testsuite on it though

azgabur:
Updated `pyproject.toml` and run `make test` and everything passed with occasional fails that after re-run pass.

pehala (approved):

jsmolar (approved):

PR #225, opened 2023-08-03
Add basic MGC tests

ficap:


ficap:
I assume the `GatewayLight` will be merged with `Gateway` depending on the potential impact on kuadrant tests.

testsuite/openshift/objects/gateway_api/__init__.py
```diff
@@ -145,3 +146,67 @@ def delete(self):
     @property
     def reference(self):
         return {"group": "gateway.networking.k8s.io", "kind": "Gateway", "name": self.name, "namespace": self.namespace}
+
+
+class GatewayLight(OpenShiftObject, Referencable):
```
pehala, line 151: I would call it MGCGateway.

testsuite/openshift/objects/gateway_api/__init__.py
```diff
@@ -145,3 +146,67 @@ def delete(self):
...
+            ]
+            return len(multi_cluster_addresses) > 0
+        except (KeyError, ModelError):
+            return False
+
+    @backoff.on_predicate(backoff.fibo, predicate=lambda x: not x.is_ready(), max_tries=20, jitter=None)
+    def wait_for_ready(self):
+        """Waits for the gateway to be ready in the sense of is_ready(self)"""
+        self.refresh()
+        return self
```
pehala, line 203: You can use until_all method from openshift library instead https://github.com/Kuadrant/testsuite/blob/5a67e31174be1aa6ace7fa3ba7c5922e067637a1/testsuite/openshift/objects/authorino.py#L50

testsuite/tests/mgc/test_basic.py
```diff
@@ -0,0 +1,105 @@
...
+# placement named "local-cluster" is created in openshift["project"] and bound to clusterset
+# gateway class "kuadrant-multi-cluster-gateway-instance-per-cluster" is created
+# openshift2["project"] is set
+#
+# dnspolicies are created and bound to gateways automatically by mgc operator
+# dnspolicies leak at this moment
+
+
+@pytest.fixture(scope="module")
+def mgc_apex_domain(openshift):
```
pehala, line 25: what is apex?

testsuite/tests/mgc/test_basic.py
```diff
@@ -0,0 +1,105 @@
...
+    """Returns preconfigured apex domain"""
+    managed_zone = openshift.do_action("get", ["managedzone", "mgc-dev-mz", "-o", "yaml"], parse_output=True)
+    return managed_zone.model["spec"]["domainName"]
+
+
+@pytest.fixture(scope="module")
+def mgc_hostname(blame, mgc_apex_domain):
```
pehala, line 32: `hostname` suffices

testsuite/tests/mgc/test_basic.py
```diff
@@ -0,0 +1,105 @@
...
+        hostname=mgc_hostname,
+        placement="local-gateway",
+    )
+    upstream_gateway.commit()
+    request.addfinalizer(upstream_gateway.delete)
+
+    return upstream_gateway.wait_for_ready()
+
+
+def _downstream_gateway_of_gateway(gateway, openshift):
```
pehala, line 53: this deserves a better name and some explanation around it

testsuite/tests/mgc/test_basic.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.fixture(scope="module")
+def http_route(gateway, backend, blame, openshift2):
+    """Creates and returns HTTPRoute bound to the backend and the downstream Gateway"""
+    downstream_gateway = _downstream_gateway_of_gateway(gateway, openshift2)
+
+    route = HTTPRoute.create_instance(
+        openshift2,
+        blame("route"),
+        parent=downstream_gateway,
+        hostname=downstream_gateway.model["spec"]["listeners"][0]["hostname"],
```
pehala, line 80: maybe the hostname could be a property of the gateway class?

testsuite/tests/mgc/test_basic.py
mganisin: > what is apex?
free to play battle royale? :-)

ficap: An apex domain is a root domain that does not contain a subdomain part. Something we would call a base domain I believe. If required we can change the naming to base domain. In this context it is used to describe a domain from which mgc creates subdomains.

pehala: Yeah, I would definitely change it to `base domain` or

testsuite/tests/mgc/test_basic.py
ficap: It depends since a gateway can have multiple listeners thus possibly multiple hostnames. It would simplify the implementation but at a cost of generality. I would rather vote for listeners property if desired. Or a full abstraction of listeners according to the gateway api spec -- `Each listener in a Gateway must have a unique combination of Hostname, Port, and Protocol`

pehala: Right, but currently, you only request the first one. We can add support to multiple ones later

pehala (dismissed):

testsuite/tests/mgc/test_basic.py
```diff
@@ -0,0 +1,110 @@
...
+This module contains the very basic tests and their dependencies for MGC
+"""
+from time import sleep
+import httpx
+import pytest
+from testsuite.openshift.httpbin import Httpbin
+from testsuite.openshift.objects.gateway_api import MGCGateway, HTTPRoute
+
+
+# prerequisites:
```
jsmolar, line 11: Move this to the module doc string and add a bullet list toa the actual prerequisites

jsmolar (approved):

pehala (approved):

PR #226, opened 2023-08-08
Clean GLBC mentions and replace them with MGC

pehala:
This includes docstrings, module and the test suite flag, which is not called `--mgc`. I also added a skip to the MGC test.


ficap (dismissed):

Makefile
azgabur, line 1: - `.PHONY` variable should be edited as well
- `testsuite/%:` rule should be edited too

azgabur (changes requested):

ficap (approved):

azgabur (approved):

PR #228, opened 2023-08-09
Update AuthConfig naming

pehala:
- Add explicit add_ prefix
- Add additional dataclasses for customization
- Change parameter naming to better match names in the AuthConfig
- APIKey tests now work with APIKey objects instead of a string
- auth_credentials test now works with AuthPolicy

### DIFF
--- a/testsuite/openshift/objects/auth_config/sections.py
+++ b/testsuite/openshift/objects/auth_config/sections.py
@@ -60,7 +60,7 @@ class Section:
 
     @modify
     def clear_all(self):
-        """Removes all identities from AuthConfig"""
+        """Clears content of a Section"""
         self.section.clear()
 
 
@@ -96,9 +96,9 @@ class IdentitySection(Section):
     def add_api_key(
         self,
         name,
+        selector: Selector,
         *,
         all_namespaces: bool = False,
-        selector: Selector = None,
         credentials: Credentials = None,
         **common_features
     ):
--- a/testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
@@ -85,8 +85,9 @@ def resource_info(request, mockserver, module_label):
     return _resource_info
 
 
+# pylint: disable=unused-argument
 @pytest.fixture()
-def commit():
+def commit(authorization):
     """Ensure no default resources are created"""
 
 
--- a/testsuite/tests/kuadrant/authorino/identity/api_key/test_reconciliation.py
+++ b/testsuite/tests/kuadrant/authorino/identity/api_key/test_reconciliation.py
@@ -2,13 +2,13 @@
 import pytest
 
 from testsuite.httpx.auth import HeaderApiKeyAuth
+from testsuite.objects import Selector
 
 
 @pytest.fixture(scope="function")
 def api_key(create_api_key, module_label):
     """Creates API key Secret"""
-    api_key = "api_key_value"
-    return create_api_key("api-key", module_label, api_key)
+    return create_api_key("api-key", module_label, "api_key_value")
 
 
 @pytest.fixture(scope="function")
@@ -20,7 +20,7 @@ def auth(api_key):
 @pytest.fixture(scope="module")
 def authorization(authorization, module_label):
     """Creates AuthConfig with API key identity"""
-    authorization.identity.api_key("api_key", match_label=module_label)
+    authorization.identity.add_api_key("api_key", selector=Selector(matchLabels={"group": module_label}))
     return authorization
 
 
--- a/testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
+++ b/testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
@@ -74,11 +74,11 @@ def authorization(authorization, openshift, module_label, authorino_domain) -> A
     authorization.add_host(authorino_domain)
 
     # get user info from admission webhook
-    authorization.identity.remove_all()
-    authorization.identity.plain("k8s-userinfo", "context.request.http.body.@fromstr|request.userInfo")
+    authorization.identity.clear_all()
+    authorization.identity.add_plain("k8s-userinfo", "context.request.http.body.@fromstr|request.userInfo")
 
     # add OPA policy to process admission webhook request
-    authorization.authorization.opa_policy("features", OPA_POLICY)
+    authorization.authorization.add_opa_policy("features", OPA_POLICY)
     user_value = ValueFrom("auth.identity.username")
 
     when = [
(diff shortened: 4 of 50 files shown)


testsuite/tests/kuadrant/authorino/operator/tls/test_webhook.py
```diff
@@ -74,8 +74,8 @@ def authorization(authorization, openshift, module_label, authorino_domain) -> A
     authorization.add_host(authorino_domain)
 
     # get user info from admission webhook
-    authorization.identity.remove_all()
-    authorization.identity.plain("k8s-userinfo", "context.request.http.body.@fromstr|request.userInfo")
+    authorization.identity.clear_all()
+    authorization.identity.add_plain("k8s-userinfo", "context.request.http.body.@fromstr|request.userInfo")
 
     # add OPA policy to process admission webhook request
     authorization.authorization.opa_policy("features", OPA_POLICY)
```
averevki, line 81:
```suggestion
    authorization.authorization.add_opa_policy("features", OPA_POLICY)
```

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -57,99 +58,85 @@ def add_item(
...
         Args:
             :param name: name of the identity
             :param audiences: token audiences
         """
-        self.add_item(name, {"kubernetes": {"audiences": [*audiences]}}, **common_features)
+        self.add_item(name, {"kubernetes": {"audiences": audiences}}, **common_features)
 
     @modify
-    def oidc(self, name, endpoint, credentials="authorization_header", selector="Bearer", **common_features):
+    def add_oidc(self, name, endpoint, *, credentials: Credentials = None, **common_features):
```
averevki, line 89: why do you need a star here?

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -57,99 +58,85 @@ def add_item(
...
+        self.section.clear()
 
-class Identities(Section):
+
+class IdentitySection(Section):
     """Section which contains identity configuration"""
 
     @modify
-    def mtls(self, name: str, selector_key: str, selector_value: str, **common_features):
+    def add_mtls(self, name: str, selector: Selector, **common_features):
```
averevki, line 71: changing number of params here is a breaking change

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -57,99 +58,85 @@ def add_item(
...
+        if credentials is None:
+            credentials = Credentials("authorization_header", "Bearer")
+        self.add_item(name, {"oidc": {"endpoint": endpoint}, "credentials": asdict(credentials)}, **common_features)
 
     @modify
-    def api_key(
+    def add_api_key(
         self,
         name,
+        *,
```
averevki, line 99: why do you need a star here?

averevki (changes requested):

testsuite/openshift/objects/auth_config/sections.py
pehala: everything in this PR is a breaking change :)

testsuite/openshift/objects/auth_config/sections.py
pehala: I do not need it, I think it would make sense to require some arguments to be passed only as a keywords

testsuite/tests/kuadrant/authorino/identity/api_key/test_reconciliation.py
```diff
@@ -20,7 +20,7 @@ def auth(api_key):
 @pytest.fixture(scope="module")
 def authorization(authorization, module_label):
     """Creates AuthConfig with API key identity"""
-    authorization.identity.api_key("api_key", match_label=module_label)
+    authorization.identity.add_api_key("api_key", selector=Selector(matchLabels={"group": module_label}))
```
azgabur, line 23: Wouldn't this be correct? 
Change `api_key` for `module_label` and:
```suggestion
    authorization.identity.add_api_key("api_key", selector=api_key.selector)
```

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -57,99 +58,85 @@ def add_item(
...
-    def oidc(self, name, endpoint, credentials="authorization_header", selector="Bearer", **common_features):
+    def add_oidc(self, name, endpoint, *, credentials: Credentials = None, **common_features):
         """Adds OIDC identity"""
-        self.add_item(
-            name,
-            {"oidc": {"endpoint": endpoint}, "credentials": {"in": credentials, "keySelector": selector}},
-            **common_features
-        )
+        if credentials is None:
+            credentials = Credentials("authorization_header", "Bearer")
```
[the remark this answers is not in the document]
pehala: Dataclasses can be made immutable (frozen=true) but are not immutable by default.

testsuite/tests/kuadrant/authorino/identity/api_key/test_reconciliation.py
pehala: This is because that test requires configuration for an API Key that does not exist yet (and will be created in the actual test)

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -57,99 +58,85 @@ def add_item(
...
-            matcher.update({"matchExpressions": [asdict(match_expression)]})
-
+        if credentials is None:
+            credentials = Credentials("authorization_header", "APIKEY")
         self.add_item(
             name,
             {
-                "apiKey": {"selector": matcher, "allNamespaces": all_namespaces},
-                "credentials": {"in": credentials, "keySelector": selector},
+                "apiKey": {"selector": asdict(selector), "allNamespaces": all_namespaces},
```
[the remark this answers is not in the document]
pehala: Made Selector required

testsuite/openshift/objects/auth_config/sections.py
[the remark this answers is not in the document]
azgabur: I think we use dataclasses as immutable, would it be beneficial to mark every one as immutable/frozen? But it maybe will raise a  problem with mutable values in dataclass instances such as list if they would be frozen.

Still seems better as to just using `None` in function declaration because inspecting just the declaration (as its done with helpful IDE's) will not convey the real default value which is `Credentials("authorization_header", "Bearer")`. Also can be problematic for typing, expecting `None` in places it should not. Maybe mention in docstring would be good.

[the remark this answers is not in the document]
pehala: We use them like it, but it is not enforced. The point of having mutable contents is one of the reasons why I would not like to introduce this change here and instead dedicate entirely new PR for it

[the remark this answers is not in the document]
azgabur: Okay, lets keep this part of code in mind for possible future change :+1:

azgabur (approved):

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -85,8 +85,9 @@ def _resource_info(org_id, owner):
     return _resource_info
 
 
+# pylint: disable=unused-argument
 @pytest.fixture()
-def commit():
+def commit(authorization):
```
averevki, line 90: 👍

averevki (approved):

PR #229, opened 2023-08-09
Add automatic builds

pehala:
Currently, the image is only built with each release, which requires manual steps to do. This PR proposes automatic builds to the `unstable` tag with occasional promotes to `latest` without doing a full release to create more frequent builds.

In the future we need to revise how we do releases and this should suffice until we figure it out.

### DIFF
--- a/.github/workflows/build_unstable_image.yml
+++ b/.github/workflows/build_unstable_image.yml
@@ -2,12 +2,14 @@ name: Build and push image
 
 on:
   push:
-    tags:
-      - v*
+    branches:
+      - main
 
 env:
   IMG_REGISTRY_HOST: quay.io
   IMG_REGISTRY_ORG: kuadrant
+  IMG_NAME: testsuite
+  IMG_TAGS: unstable
 
 jobs:
   build:
@@ -16,24 +18,11 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
-      - name: Add tag name
-        id: tag-name
-        run: |
-          echo "IMG_TAGS=$(echo ${{ github.ref_name }} | sed -r 's/(v)(.*)/\2/')" >> $GITHUB_ENV
-      - name: Format version
-        id: format-version
-        run: |
-          echo "IMG_TAGS=${{ env.IMG_TAGS }} $(echo ${{ env.IMG_TAGS }} | sed -r 's/(.*)(\.[0-9]*)(.*)/\1/')" >> $GITHUB_ENV
-      - name: Add latest tag
-        if: ${{ !contains(github.ref_name, 'rc') }}
-        id: add-latest-tag
-        run: |
-          echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
       - name: Build Image
         id: build-image
         uses: redhat-actions/buildah-build@v2
         with:
-          image: testsuite
+          image: ${{ env.IMG_NAME }}
           tags: ${{ env.IMG_TAGS }}
           layers: true
           platforms: linux/amd64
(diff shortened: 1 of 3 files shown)


.github/workflows/build_unstable_image.yml
```diff
@@ -0,0 +1,44 @@
...
+  build:
+    name: Build and push image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Add latest tag
+        id: add-latest-tag
+        run: |
+          echo "IMG_TAGS=unstable" >> $GITHUB_ENV
```
mganisin, line 22: why it is not in `env:`?

.github/workflows/promote_to_latest.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Log in to Quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          registry: ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}
+          username: ${{ secrets.IMG_REGISTRY_USERNAME }}
+          password: ${{ secrets.IMG_REGISTRY_TOKEN }}
+      - name: Create repository name
+        run: |
+          echo "REPOSITORY=${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/testsuite" >> $GITHUB_ENV
```
mganisin, line 23: why it is not in `env:`?

mganisin (approved):

jsmolar (dismissed):

.github/workflows/build_unstable_image.yml
```diff
@@ -0,0 +1,41 @@
...
+    name: Build and push image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Build Image
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: testsuite
```
ficap, line 24: I see REGISTRY_HOST and REGISTRY_ORG even TAG are env vars, I'd personally turn also image name to envvar.

.github/workflows/build_unstable_image.yml
```diff
@@ -0,0 +1,42 @@
...
+          platforms: linux/amd64
+          containerfiles: |
+            ./Dockerfile
+      - name: Push Image
+        if: ${{ !env.ACT }}
+        id: push-to-quay
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }}
```
ficap, line 37: I really don't like the idea of pushing directly to the floating tag. I'd rather see pushing to tag named after the commit hash and then pushing to IMG_TAGS to have a tag handle for each image instead of ambiguous floating tag. The same applies also for promote_to_latest.

pehala: Why?

ficap: To have unambiguous handle for each unstable build e.g. (in which image version the issue has appeared, what image version was used for your testing) this could be somehow resolved by using image digests BUT all images in a registry without a tag may be deleted at any time by a garbage collector. This might not matter much for unstable images but it would definitively matter for "promoted" images.

pehala: We have always used floating tags, even in the 3scale days and I cannot remember any problem we ever had with them :) Besides the "unchanging" images will still exist, and releases are not going away, they are just not automatic and require more work and I would like them to be aligned with the project releases. 

This just releases builds a releases some images with minimal work required for anyone to try.

ficap: Discussed in person and decided to implement it later together with aligning building process with engineering's one.

ficap (approved):

jsmolar (approved):

PR #230, opened 2023-08-16
asdict_compose function for composed dataclasses

azgabur:
Just an idea. Currently used just in one dataclass but I predict it would be useful to have it in future.
Also contains documentation fixes and small rename of overriding asdict functions.

### DIFF
--- a/testsuite/objects/__init__.py
+++ b/testsuite/objects/__init__.py
@@ -1,27 +1,27 @@
 """Module containing base classes for common objects"""
 import abc
-from dataclasses import dataclass, is_dataclass, fields
+from dataclasses import dataclass, is_dataclass, fields, replace
 from copy import deepcopy
 from typing import Literal, Union
 
 JSONValues = Union[None, str, int, bool, list["JSONValues"], dict[str, "JSONValues"]]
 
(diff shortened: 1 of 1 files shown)


testsuite/objects/__init__.py
```diff
@@ -104,9 +121,9 @@ class Property:
     name: str
     value: ABCValue
 
-    def asdict(self):
+    def asdict_override(self):
         """Override `asdict` function"""
-        return {"name": self.name, **asdict(self.value)}
+        return asdict_compose(self)
```
pehala, line 126: To be honest, I dont think this is an improvement as the override function is still there

testsuite/objects/__init__.py
```diff
@@ -92,7 +109,7 @@ class ValueFrom(ABCValue):
 
     authJSON: str  # pylint: disable=invalid-name
 
-    def asdict(self):
+    def asdict_override(self):
```
pehala, line 112: I do not think renaming is needed

azgabur:
Due to currently not being very beneficial I close this. Maybe in the future I will revisit this topic.

PR #238, opened 2023-09-26
Add kuadrant namespace annotation to gateway

azgabur:
This will fix the Reconciliation error of RateLimitPolicy in Limitador tests but the tests themselves are still failing and need more investigation to fix.
Related issue: https://github.com/Kuadrant/kuadrant-operator/issues/259


azgabur:
I tested it and it was fixed. This workaround is no longer needed.

PR #240, opened 2023-10-12
Make mgc test_smoke use TLSPolicy

ficap:


testsuite/tests/mgc/test_basic.py
```diff
@@ -30,7 +32,9 @@ def test_smoke(route):
     Tests whether the backend, exposed using the HTTPRoute and Gateway, was exposed correctly,
     having a tls secured endpoint with a hostname managed by MGC
     """
-    backend_client = route.client(verify=False)  # self-signed certificate; TBD
+    backend_client = HttpxBackoffClient(
```
pehala, line 35: can we verify that certificate is correct somehow? I think additional asserts, or HTTPX verify would be great

pehala (approved):

testsuite/tests/mgc/test_basic.py
```diff
@@ -25,12 +27,16 @@ def test_gateway_readiness(gateway):
...
     """
     Tests whether the backend, exposed using the HTTPRoute and Gateway, was exposed correctly,
     having a tls secured endpoint with a hostname managed by MGC
     """
-    backend_client = route.client(verify=False)  # self-signed certificate; TBD
+
+    tls_cert = upstream_gateway.get_tls_cert()
+
+    # assert that tls_cert is used by the server
+    backend_client = HttpxBackoffClient(base_url=f"https://{route.hostnames[0]}", verify=tls_cert)
```
averevki, line 39: would prefer "http or https" logic moved to the upper level, in HTTPRoute class / abstract it to fixture. And just client.get() here

pehala: Will be part of the next PR with the rework

averevki (approved):

testsuite/openshift/objects/gateway_api/__init__.py
```diff
@@ -12,3 +16,22 @@ def reference(self) -> dict[str, str]:
...
+    Manually creates Reference object.
+    https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.ParentReference
+    """
+
+    @property
+    def reference(self) -> dict[str, Any]:
+        return asdict(self)
+
+    group: str
+    kind: str
```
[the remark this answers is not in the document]
pehala: From our usage, it is better to have those explicit as we use it for both Gateways, Secrets and Issuers.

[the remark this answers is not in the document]
averevki: fair point 👍

PR #241, opened 2023-10-13
Fix metrics tests broken by reformat

averevki:
Were broken by #235 and #228

### DIFF
--- a/testsuite/openshift/client.py
+++ b/testsuite/openshift/client.py
@@ -98,6 +98,11 @@ class OpenShiftClient:
         with self.context:
             return oc.selector(f"route/{name}").object(cls=OpenshiftRoute)
 
+    def get_routes_for_service(self, service_name: str) -> list[OpenshiftRoute]:
+        """Returns list of routes for given service"""
+        with self.context:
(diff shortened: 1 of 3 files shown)


jsmolar (approved):
The only thing that I am scared of, is over-extending `OpenShiftClient`.

azgabur (approved):

pehala:
> The only thing that I am scared of, is over-extending `OpenShiftClient`.

Agreed, I am fine with this particular method, but the alternative might be to move it to be a class method in OpenShiftRoute. @jsmolar WDYT?

jsmolar:
> > The only thing that I am scared of, is over-extending `OpenShiftClient`.
> 
> Agreed, I am fine with this particular method, but the alternative might be to move it to be a class method in OpenShiftRoute. @jsmolar WDYT?

I don't like `OpenShiftRoute` either. The best place would be to get it straight from the service, but it is pointless to create such an object just for this. I am afraid I don't have an answer for this.

pehala:
Merging for now, we might resolve this when we add a Service object.

PR #244, opened 2023-10-18
Enhance metrics tests skip

pehala:
- Now it should skip if the userWorkloadMonitoring is off
- Add ConfigMap class for ease of use


testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -16,9 +19,18 @@ def prometheus(request, openshift):
     Return an instance of OpenShift metrics client
     Skip tests if query route is not properly configured
     """
+    openshift_monitoring = openshift.change_project("openshift-monitoring")
+    # Check if metrics are enabled
+    try:
+        with openshift_monitoring.context:
+            cm = selector("cm/cluster-monitoring-config").object(cls=ConfigMap)
+            assert yaml.safe_load(cm["config.yaml"])["enableUserWorkload"]
+    except:  # pylint: disable=bare-except
```
averevki, line 28:
```suggestion
    except Exception:
```

testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -1,6 +1,9 @@
 """Conftest for the Authorino metrics tests"""
 import pytest
+import yaml
```
averevki, line 3: nit: should be a top-level import

testsuite/tests/kuadrant/authorino/metrics/conftest.py
pehala: That would cause `too-broad-except` instead

testsuite/tests/kuadrant/authorino/metrics/conftest.py
pehala: If it is ok with Black, I will probably not change it

averevki (approved):

jsmolar (approved):

PR #246, opened 2023-10-18
Fix test_auth_credentials.py

pehala:
- Run it also on Kuadrant
- Fix wrong asserts

### DIFF
--- a/testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
+++ b/testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
@@ -1,7 +1,6 @@
 """Test for RHSSO auth credentials"""
 import pytest
 
-from testsuite.openshift.objects.auth_config import AuthConfig
 from testsuite.objects import Credentials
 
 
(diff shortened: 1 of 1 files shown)


testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
```diff
@@ -12,33 +11,45 @@ def credentials(request):
...
+    """Add RHSSO identity to Authorization"""
+    authorization.identity.clear_all()
     authorization.identity.add_oidc("rhsso", rhsso.well_known["issuer"], credentials=Credentials(credentials, "Token"))
     return authorization
 
 
 def test_custom_selector(client, auth, credentials):
     """Test if auth credentials are stored in right place"""
     response = client.get("/get", headers={"authorization": "Token " + auth.token.access_token})
-    assert response.status_code == 200 if credentials == "authorization_header" else 401
```
averevki, line 25: Wouldn't it be sufficient to just put brackets like this? 
`assert response.status_code == (200 if credentials == "authorization_header" else 401)`

testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
```diff
@@ -12,33 +11,45 @@ def credentials(request):
 
 
 @pytest.fixture(scope="module")
-def authorization(rhsso, openshift, blame, route, module_label, credentials):
-    """Add RHSSO identity to AuthConfig"""
-    authorization = AuthConfig.create_instance(openshift, blame("ac"), route, labels={"testRun": module_label})
+def authorization(authorization, rhsso, credentials):
+    """Add RHSSO identity to Authorization"""
+    authorization.identity.clear_all()
```
jsmolar, line 16: You don't need to clear identities here. `authorization` from rhsso conftest should be enough

pehala: I do, due to parametrization

testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
pehala: It would, I think it is more readable with separate if

averevki (approved):

jsmolar (approved):

PR #247, opened 2023-10-18
Fix tls secret deletion

averevki:


testsuite/openshift/objects/gateway_api/gateway.py
```diff
@@ -122,6 +122,11 @@ def get_tls_cert(self) -> Certificate:
         )
         return tls_cert
 
+    def delete_tls_secret(self):
+        """Deletes secret with TLS certificate used by the gateway"""
+        tls_cert_secret_name = self.model.spec.listeners[0].tls.certificateRefs[0].name
```
pehala, line 127: since we have this exact logic/line elsewhere I would probably prefer for it to be extracted into a property

pehala (approved):

jsmolar (approved):

PR #248, opened 2023-10-23
Authpolicy v1beta2 reformat

azgabur:
This PR makes this testsuite compatible with newest AuthConfig/AuthPolicy v1beta2

Few notes:
- Envoy Dynamic Metadata function needs to be added in Responses section when tests for it will be created
    - **Wristband tests are for now skipped!** 
- Test for Custom deny code/body/headers need to be created
- testsuite/tests/kuadrant/authorino/response/test_headers.py is maybe not needed as wrapperKey no longer exists or needs some rewrite

### DIFF
--- a/testsuite/objects/__init__.py
+++ b/testsuite/objects/__init__.py
@@ -73,12 +73,14 @@ class Selector:
 class Credentials:
     """Dataclass for Credentials structure"""
 
-    in_location: str
+    in_location: Literal["authorizationHeader", "customHeader", "queryString", "cookie"]
     keySelector: str
 
     def asdict(self):
-        """Custom asdict, because I cannot use 'in' as a name"""
-        return {"in": self.in_location, "keySelector": self.keySelector}
+        """Custom asdict because of needing to put location as parent dict key for inner dict"""
+        if self.in_location == "authorizationHeader":
+            return {self.in_location: {"prefix": self.keySelector}}
+        return {self.in_location: {"name": self.keySelector}}
 
 
 @dataclass
@@ -116,39 +118,7 @@ class Value(ABCValue):
 class ValueFrom(ABCValue):
     """Dataclass for dynamic Value. It contains reference path to existing value in AuthJson."""
 
-    authJSON: str
-
-    def asdict(self):
-        """Override `asdict` function"""
-        return {"valueFrom": {"authJSON": self.authJSON}}
-
-
-@dataclass
-class Property:
-    """Dataclass for static and dynamic values. Property is a Value with name."""
-
-    name: str
-    value: ABCValue
-
-    def asdict(self):
-        """Override `asdict` function"""
-        return {"name": self.name, **asdict(self.value)}
-
-
-@dataclass
-class ExtendedProperty(Property):
-    """
-    Dataclass extending Property class adding optional `overwrite` feature
-    used in extended_properties functionality in Identity section.
-    """
-
-    overwrite: Optional[bool] = None
-
-    def asdict(self):
-        """Extend inherited `asdict` function to include new attributes."""
-        if self.overwrite is not None:
-            return {**super().asdict(), "overwrite": self.overwrite}
-        return super().asdict()
+    selector: str
 
 
 @dataclass
--- a/testsuite/openshift/objects/auth_config/__init__.py
+++ b/testsuite/openshift/objects/auth_config/__init__.py
@@ -25,7 +25,7 @@ class AuthConfig(OpenShiftObject):
     @cached_property
     def identity(self) -> IdentitySection:
         """Gives access to identity settings"""
-        return IdentitySection(self, "identity")
+        return IdentitySection(self, "authentication")
 
     @cached_property
     def metadata(self) -> MetadataSection:
@@ -48,7 +48,7 @@ class AuthConfig(OpenShiftObject):
     ):
         """Creates base instance"""
         model: Dict = {
-            "apiVersion": "authorino.kuadrant.io/v1beta1",
+            "apiVersion": "authorino.kuadrant.io/v1beta2",
             "kind": "AuthConfig",
             "metadata": {"name": name, "namespace": openshift.project, "labels": labels},
             "spec": {"hosts": hostnames or [route.hostname]},  # type: ignore
@@ -70,13 +70,6 @@ class AuthConfig(OpenShiftObject):
     def remove_all_hosts(self):
         """Remove all hosts"""
         self.model.spec.hosts = []
-
-    @modify
-    def set_deny_with(self, code, value):
-        """Set denyWith"""
-        self.auth_section["denyWith"] = {
-            "unauthenticated": {"code": code, "headers": [{"name": "Location", "valueFrom": {"authJSON": value}}]}
-        }
 
     @modify
     def add_rule(self, when: list[Rule]):
--- a/testsuite/openshift/objects/auth_config/__init__.py
+++ b/testsuite/openshift/objects/auth_config/__init__.py
@@ -2,7 +2,7 @@
 from functools import cached_property
 from typing import Dict, List, Optional
 
-from testsuite.objects import Rule
+from testsuite.objects import Rule, asdict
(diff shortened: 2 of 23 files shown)


testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -208,7 +230,9 @@ def add_simple(self, auth_json: str, name="simple", key="data", **common_feature
     @modify
     def add_json(self, name: str, properties: list[Property], **common_features):
         """Adds json response to AuthConfig"""
-        asdict_properties = [asdict(p) for p in properties]
+        asdict_properties = {}
+        for prop in properties:
+            asdict_properties.update(asdict(prop))
```
pehala, line 235: I do not follow this change, does this mean properties can override each other?

averevki: I think it's just a way to populate a dict with other dicts in the loop.

*You probably can use something like this to disable overwriting if you want.
```suggestion
            asdict_properties = {**prop, **asdict_properties}
```

pehala: I meant that before it was a list, meaning all properties were separated, but now it is a one-level dict so that they can override each other. That change is fundamental on the AuthPolicy/AuthConfig level.

Previously we used List because in the CR it was a list, now it is changed to a dict, but we kept the list, which can cause confusion. IF I specify two properties with the same name, I would expect both of them to be propagated to the CR, but now they wont, only one will be.

Basically, I think maybe we should change to dict as well to keep it consistent.

averevki: Makes sense about possible confusion. Although, the behavior should be the same. @azgabur what would happen if you specify 2 properties with the same name? Would it be overridden / raise an error during creation?

testsuite/tests/kuadrant/authorino/response/test_headers.py
```diff
@@ -16,11 +16,11 @@ def header_name(request):
 def authorization(authorization, header_name):
     """Add response to Authorization"""
     authorization.responses.clear_all()  # delete previous responses due to the parametrization
-    authorization.responses.add_json("header", [Property("anything", Value("one"))], wrapper_key=header_name)
```
averevki, line 19: `operator.http.conftest` also requires a change with wrapper_key to work properly

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -180,23 +198,27 @@ def add_uma(self, name, endpoint, credentials_secret, **common_features):
 
 
 class ResponseSection(Section):
-    """Section which contains response configuration"""
+    """Section which contains response configuration. todo envoyDynamicMetadata"""
```
averevki, line 201: If we are leaving dynamic metadata implementation for the future PR, we should somehow mark wristband tests as they are currently raising an error

testsuite/openshift/objects/auth_config/__init__.py
```diff
@@ -25,7 +25,7 @@ def authorization(self) -> AuthorizationSection:
     @cached_property
     def identity(self) -> IdentitySection:
         """Gives access to identity settings"""
-        return IdentitySection(self, "identity")
+        return IdentitySection(self, "authentication")
```
averevki, line 28: should we also rename `identity` tests package to `authentication` to avoid confusion?

averevki (commented):
rhsso tests don't work on my machine (I don't know if they were working before 🙂)

testsuite/openshift/objects/auth_config/sections.py
pehala: Also sidenote: this does not have to solved in this PR specifically, follow-up PR is sufficient

testsuite/openshift/objects/auth_config/__init__.py
pehala: I agree, but can be done later

testsuite/openshift/objects/auth_config/sections.py
azgabur: I think two properties with same name are no longer possible with the new version - will raise OpenShiftError. Same with other places that it was changed from list to dict.  
Changing the function signature to dict would make object `Property` redundant as the name will now be a key of the dict. I can do the change.

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -42,29 +61,12 @@ def committed(self):
...
-        if when:
-            item["when"] = [asdict(x) for x in when]
-        if metrics:
-            item["metrics"] = metrics
-        if cache:
-            item["cache"] = asdict(cache)
-        if priority:
-            item["priority"] = priority
-        self.section.append(item)
+        add_common_features(value, **common_features)
```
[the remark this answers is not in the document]
azgabur: I extracted it to static function as I needed a way to add common features for new Responses add function. I am not sure if its really needed - meaning I am not sure if Responses section uses some common features. I need to research this more.

testsuite/openshift/objects/auth_config/sections.py
azgabur: Added pytest skip

testsuite/tests/kuadrant/authorino/response/test_headers.py
azgabur: Fixed

testsuite/objects/__init__.py
```diff
@@ -73,12 +73,14 @@ def __post_init__(self):
...
 
-    in_location: str
+    in_location: Literal["authorizationHeader", "customHeader", "queryString", "cookie"]
     keySelector: str
 
     def asdict(self):
-        """Custom asdict, because I cannot use 'in' as a name"""
-        return {"in": self.in_location, "keySelector": self.keySelector}
+        """Custom asdict because of needing to put location as parent dict key for inner dict"""
+        if self.in_location == "authorizationHeader":
```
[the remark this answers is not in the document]
pehala: In this case I do not think Abstract class is required but yes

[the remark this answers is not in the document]
azgabur: I was thinking about this for some time but cant find any solution that fits better as just custom `asdict` as its in commit https://github.com/Kuadrant/testsuite/pull/248/commits/88fed67d671a04712841c950522b540914ab8408. Please take a look at my new commit https://github.com/Kuadrant/testsuite/pull/248/commits/4673136b28eac5af9b0e3c294397e841fa299eb4 as it is my best idea and if you think its okay I will do tests refactor. Otherwise please advise how would you like this to look.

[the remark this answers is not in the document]
pehala: I do not think the assert is needed, I would simply remove the value from in_header Literal to indicate what values are accepted in the main class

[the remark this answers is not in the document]
azgabur: I reverted, please make your refactor PR as we discussed. Thanks!

pehala (dismissed):

azgabur:
^ Additional place for renamed Credentials locations I forgot about

testsuite/openshift/objects/auth_config/sections.py
```diff
@@ -180,35 +197,41 @@ def add_uma(self, name, endpoint, credentials_secret, **common_features):
...
     ):
         """Add response to AuthConfig"""
-        if wrapper:
-            value["wrapper"] = wrapper
-        if wrapper_key:
-            value["wrapperKey"] = wrapper_key
-
-        self.add_item(name, value, **common_features)
+        add_common_features(value, **common_features)
+        self.success_headers.update({name: value})
```
averevki, line 220: are these `success_headers` used somewhere in the testsuite?

pehala (approved):

averevki (approved):

testsuite/openshift/objects/auth_config/sections.py
azgabur: Nope. Just in the `_add` function.

PR #249, opened 2023-10-23
Add tests for TLSPolicy certificate parameters

averevki:
Add tests for introspection of the certificate generated by TLSPolicy
- Add explicit cryptography package dependency
- Add introspection methods to the testsuite Certificate dataclass
- Add tests


testsuite/tests/mgc/tlspolicy/test_cert_parameters.py
```diff
@@ -0,0 +1,67 @@
...
+def test_tls_cert_duration(tls_cert, cert_params):
+    """Test certificate duration"""
+    duration_time = datetime.strptime(cert_params["duration"], "%Hh")
+    assert tls_cert.duration == timedelta(hours=duration_time.hour)
+
+
+def test_tls_cert_usages(tls_cert):
+    """Test certificate usages"""
+    assert tls_cert.usages.digital_signature
+    assert tls_cert.usages.key_cert_sign
```
pehala, line 61: I would add at least one negative assert for usage that was not specified

testsuite/tests/mgc/tlspolicy/test_cert_parameters.py
```diff
@@ -0,0 +1,67 @@
...
+
+@pytest.fixture(scope="module")
+def tls_cert(upstream_gateway, gateway):  # pylint: disable=unused-argument
+    """Return certificate generated by TLSPolicy"""
+    return upstream_gateway.get_tls_cert()
+
+
+def test_tls_cert_common_name(tls_cert, cert_params):
+    """Test certificate Common Name"""
+    assert tls_cert.common_names[0].value == cert_params["commonName"]
```
pehala, line 49: I would prefer explicit string

testsuite/tests/mgc/tlspolicy/test_cert_parameters.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.fixture(scope="module")
+def dns_policy():
+    """Don't need DNSPolicy because only testing certificate generated by TLSPolicy"""
+    return None
+
+
+@pytest.fixture(scope="module")
+def tls_policy(tls_policy, cert_params):
+    """Update TLSPolicy with certificate parameters"""
+    tls_policy.model.spec.update(cert_params)
```
pehala, line 37: I would prefer if you specified all the cert values here instead of just applying dict

testsuite/certificates/__init__.py
```diff
@@ -30,6 +33,30 @@ class Certificate:
     certificate: str
     chain: str
 
+    @cached_property
+    def _decoded(self) -> x509.Certificate:
```
pehala, line 37: I think this can be public

testsuite/tests/mgc/tlspolicy/test_cert_parameters.py
pehala: This version is slightly better, but I would still prefer updating the existing policy, than creating a new one

pehala: One of the options how to achieve this, would be to make TLSPolicy dict-like and then use `tls_policy["commonName"] = commonName`

averevki: good suggestion 👍

pehala (dismissed):

testsuite/certificates/__init__.py
```diff
@@ -30,6 +33,31 @@ class Certificate:
     certificate: str
     chain: str
 
+    @cached_property
+    def decoded(self) -> x509.Certificate:
```
ficap, line 37: since the Certificate dataclass is not frozen, these cached_properties may become outdated and invalid when underlying certificate property is changed

ficap (approved):

PR #259, opened 2023-11-02
Add has_dns_error() and has_tls_error() assert to mgc tests

pehala:
* Add two new asserts to MGC tests, which ensure that in case of incorrect configuration or a bug, tests will Fail instead throwing and Exception
* Rename our HttpX client to `KuadrantClient`

### DIFF
--- a/testsuite/httpx/__init__.py
+++ b/testsuite/httpx/__init__.py
@@ -1,9 +1,11 @@
 """Common classes for Httpx"""
+# I change return type of HTTPX client to Kuadrant Result
+# mypy: disable-error-code="override, return-value"
 from tempfile import NamedTemporaryFile
 from typing import Union
 
 import backoff
-from httpx import Client, Response, ConnectError
+from httpx import Client, ConnectError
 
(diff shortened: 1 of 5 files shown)


testsuite/httpx/__init__.py
```diff
@@ -18,15 +20,50 @@ def create_tmp_file(content: str):
     return file
 
 
-class UnexpectedResponse(Exception):
-    """Slightly different response attributes were expected or no response was given"""
+class KuadrantResult:
```
jsmolar, line 23: Just an idea for discussion. Change this to `KuadrantResponse` and extend `Response`.

pehala: I can change the name, but sadly there is no way how I can extend the class, because in case of succes response I would have to do basically a copy (pass all params from response I got into a new object) and in case of error I wouldnt have all the required info to create a response

jsmolar (dismissed):

jsmolar (dismissed):

ficap (approved):

PR #260, opened 2023-11-02
Add Authorino make target

pehala:
- Update README.md


averevki (approved):

Makefile
azgabur, line 1: Please add `authorino` to .PHONY variable at line 1 as it is not a build target.
Also please add `performance` to the variable too, was probably missed in earlier PR.

azgabur (approved):

averevki (approved):

PR #263, opened 2023-11-03
Fix wrong credentials param test_auth_credentials.py

azgabur:
Fix wrong credentials parameters that were missed in #248 
It is interesting that it was not caught in automated commit test via `mypy` tool because the variables are behind `pytest` parametrize feature.

### DIFF
--- a/testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
+++ b/testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py
@@ -21,7 +21,7 @@ def authorization(authorization, rhsso, credentials):
 def test_custom_selector(client, auth, credentials):
     """Test if auth credentials are stored in right place"""
     response = client.get("/get", headers={"authorization": "Token " + auth.token.access_token})
-    if credentials == "authorization_header":
+    if credentials == "authorizationHeader":
         assert response.status_code == 200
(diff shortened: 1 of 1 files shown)


pehala (approved):

averevki (commented):
Good find! With it, I think you should fix it also in another missed place 🫤: 
https://github.com/Kuadrant/testsuite/blob/37a26c4cb74aeb35875a4181c8a7c7012f6d84d4/testsuite/tests/kuadrant/authorino/identity/api_key/test_auth_credentials.py#L21-L24

Would be great to also fix the tests there, like the fix from https://github.com/Kuadrant/testsuite/pull/246 😊

pehala:
> Good find! With it, I think you should fix it also in another missed place 🫤:
> 
> https://github.com/Kuadrant/testsuite/blob/37a26c4cb74aeb35875a4181c8a7c7012f6d84d4/testsuite/tests/kuadrant/authorino/identity/api_key/test_auth_credentials.py#L21-L24
> 
> Would be great to also fix the tests there, like the fix from #246 😊

Could be, but at the same time, it can be done in another PR, feel free to create one :)

azgabur:
^ renamed commit

azgabur:
Created the other MR :)
#265

averevki (approved):

PR #266, opened 2023-11-06
Fix mgc tls secret deletion

averevki:
Fix TLSPolicy secret cleanup if secret does not exist

### DIFF
--- a/testsuite/openshift/objects/gateway_api/gateway.py
+++ b/testsuite/openshift/objects/gateway_api/gateway.py
@@ -124,8 +124,8 @@ class MGCGateway(Gateway):
 
     def delete_tls_secret(self):
         """Deletes secret with TLS certificate used by the gateway"""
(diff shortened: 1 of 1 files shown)


testsuite/openshift/objects/gateway_api/gateway.py
```diff
@@ -124,8 +124,8 @@ def get_tls_cert(self) -> Certificate:
 
     def delete_tls_secret(self):
         """Deletes secret with TLS certificate used by the gateway"""
-        tls_secret = self.openshift.get_secret(self.cert_secret_name)
-        tls_secret.delete(ignore_not_found=True)
+        if tls_secret := self.openshift.get_secret(self.cert_secret_name, ignore_not_found=True):
```
pehala, line 127: how about using `oc.selector().delete()`?

pehala (approved):

jsmolar (approved):

PR #267, opened 2023-11-06
Refactor settings structure

pehala:
* Rename `kuadrant` to `service_protection`
  * Move `authorino` and `envoy` under `service_protection`
* Rename `mgc` to `control_plane`
* Move `kuadrant.enable` to `gateway_api`
* Change how OpenShift are configured:
   * Introduce `cluster` configuration, which represent the cluster info where everything should run
      * By default is completely empty, which means that whatever server user is logged in will be used
      * This would be overridden if ran with MGC as the workload will be automatically moved to a spoke
   * All settings in `service_protection` section are only a projects on the cluster, no other configuration is allowed
   * `control_plane.spokes` and `control_plane.hub` (new settings) require a full cluster configuration, same as `cluster`
* Change `openshift()` and `openshift2()` fixtures to a module scope
  * When running future tests with MGC, the actual client is dependant on the hub gateway and spoke
  * This change means that `backend` is now also `module` scope, meaning slower tests
  
## Current minimal configuration:
```
default:
  skip_cleanup: false
  dynaconf_merge: true
  gateway_api: true
  control_plane:
    hub:
      project: "multi-cluster-gateways"
    spokes:
      local-cluster: {}
```

### DIFF
--- a/testsuite/config/__init__.py
+++ b/testsuite/config/__init__.py
@@ -33,12 +33,16 @@ settings = Dynaconf(
     envvar_prefix="KUADRANT",
     merge_enabled=True,
     validators=[
-        Validator("authorino.deploy", must_exist=True, eq=True)
-        | (Validator("authorino.auth_url", must_exist=True) & Validator("authorino.oidc_url", must_exist=True)),
+        Validator("service_protection.authorino.deploy", must_exist=True, eq=True)
+        | (
+            Validator("service_protection.authorino.auth_url", must_exist=True)
+            & Validator("service_protection.authorino.oidc_url", must_exist=True)
+        ),
         DefaultValueValidator("rhsso.url", default=fetch_route("no-ssl-sso")),
         DefaultValueValidator("rhsso.password", default=fetch_secret("credential-sso", "ADMIN_PASSWORD")),
         DefaultValueValidator("mockserver.url", default=fetch_route("mockserver", force_http=True)),
-        Validator("kuadrant.enable", must_exist=False, eq=False) | Validator("kuadrant.gateway.name", must_exist=True),
+        Validator("gateway_api", must_exist=False, eq=False)
+        | Validator("service_protection.gateway.name", must_exist=True),
     ],
     validate_only=["authorino", "kuadrant"],
     loaders=["dynaconf.loaders.env_loader", "testsuite.config.openshift_loader"],
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -5,7 +5,6 @@ from urllib.parse import urlparse
 import pytest
 from dynaconf import ValidationError
 from keycloak import KeycloakAuthenticationError
-from weakget import weakget
 
 from testsuite.certificates import CFSSLClient
 from testsuite.config import settings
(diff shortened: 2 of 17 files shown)


pehala:
> I was thinking about `control_plane.enabled: true` flag instead of marking every mgc test (I'm open to creating PR after this is merged). What do you think?

I think we will use a different "smarter" way of marking them. I would like to keep the commandline switch

jsmolar (dismissed):

testsuite/tests/conftest.py
```diff
@@ -215,15 +214,14 @@ def module_label(label):
     return randomize(label)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def kuadrant(testconfig, openshift):
     """Returns Kuadrant instance if exists, or None"""
-    settings = weakget(testconfig)
-    if not settings["kuadrant"]["enabled"] % True:
+    if not testconfig.get("gateway_api", True):
```
averevki, line 220: I don't like empty settings check in the fixtures. Maybe move all settings populations (get, weakget) to the upper, dynaconf loader level?

testsuite/config/__init__.py
```diff
@@ -33,12 +33,16 @@ def __init__(self, name, default, **kwargs) -> None:
     envvar_prefix="KUADRANT",
     merge_enabled=True,
     validators=[
-        Validator("authorino.deploy", must_exist=True, eq=True)
-        | (Validator("authorino.auth_url", must_exist=True) & Validator("authorino.oidc_url", must_exist=True)),
+        Validator("service_protection.authorino.deploy", must_exist=True, eq=True)
+        | (
+            Validator("service_protection.authorino.auth_url", must_exist=True)
+            & Validator("authorino.oidc_url", must_exist=True)
```
averevki, line 39:
```suggestion
            & Validator("service_protection.authorino.oidc_url", must_exist=True)
```

testsuite/tests/conftest.py
averevki, line 220: Why defaulting `gateway_api` to `True` here and defaulting to `False` in the validator?

testsuite/tests/mgc/conftest.py
```diff
@@ -103,13 +112,18 @@ def gateway(upstream_gateway, spokes, hub_policies_commit):
     return gw
 
 
+@pytest.fixture(scope="module")
+def openshift(gateway):
+    """OpenShift client for the primary namespace"""
+    return gateway.openshift
```
averevki, line 118: I think this would cause troubles with circular fixtures in the feature tests.

testsuite/tests/mgc/conftest.py
```diff
@@ -25,16 +25,25 @@ def backend(request, gateway, blame, label):
...
+    """Openshift client for a Hub cluster"""
+    client = testconfig["control_plane"]["hub"]
+    if not client.connected:
+        pytest.fail("You are not logged into Openshift or the namespace doesn't exist")
+    return client
+
+
 @pytest.fixture(scope="module")
-def upstream_gateway(request, openshift, blame, hostname, module_label):
+def upstream_gateway(request, hub_openshift, blame, hostname, module_label):
```
averevki, line 43: You should also change it in `dnspolicy/conftest.py`

testsuite/tests/mgc/conftest.py
pehala: Could you name which tests you see as potential issues? I ran all the tests (at least with gateway_api.enabled) and it all passed.

averevki (approved):

PR #282, opened 2023-11-15
Fix tests with credentials

averevki:
Tests compared `Credentials` dataclass with a string before, which always resulted in a false condition.


testsuite/objects/__init__.py
```diff
@@ -82,6 +82,10 @@ def asdict(self):
             return {self.in_location: {"prefix": self.keySelector}}
         return {self.in_location: {"name": self.keySelector}}
 
+    def __eq__(self, other_location):
```
pehala, line 85: I think `__eq__` should check all the vars, I would much rather prefer changing it in the tests

jsmolar (approved):

PR #288, opened 2023-11-20
Add capabilities

### LINKED ISSUE #164: Implement Capabilities
**Summary**

We need to design and implement capabilities in some form. 
e.g. some tests should be skipped when running against Kuadrant. The current solution is to mark test modules with `run_on_kuadrant` or `skip_on_kuadrant`, which has serious limitations and does not cover all the posibilities

pehala:
# Summary

* Add marks for `authorino`, `limitador`, `kuadrant`, `mgc`, `standalone`, which serve both as metadata to select tests and a way how to skip if the requires setup is not present
* Add option `--enforce` which fails the test if the required setup is not present instead of a skip
* Add makefile options to select tests based on those marks
* Remove `--mgc` mark
* Annotate all tests
* Rename `gateway_api` option to `standalone` and use inverted values
    * Makes naming more consistent 

Fixes #164

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: commit-acceptance pylint mypy black reformat test performance authorino poetry poetry-no-dev mgc container-image polish-junit reportportal
+.PHONY: commit-acceptance pylint mypy black reformat test performance authorino poetry poetry-no-dev mgc container-image polish-junit reportportal authorino-standalone limitador kuadrant kuadrant-only
 
 TB ?= short
 LOGLEVEL ?= INFO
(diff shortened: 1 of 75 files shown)


testsuite/tests/kuadrant/authorino/metrics/conftest.py
```diff
@@ -8,11 +8,7 @@
 from testsuite.openshift.config_map import ConfigMap
 from testsuite.openshift.metrics import ServiceMonitor, MetricsEndpoint, Prometheus
 
-
-@pytest.fixture(scope="module")
-def run_on_kuadrant():
-    """Kuadrant doesn't allow customization of Authorino parameters"""
-    return False
+pytestmark = [pytest.mark.authorino, pytest.mark.standalone]
```
averevki, line 11: Do you need to specify the mark in conftest also?

testsuite/utils.py
```diff
@@ -24,6 +24,17 @@
...
 
+class Singleton(type):
+    """Metaclass for creating Singletons"""
+
+    def __init__(cls, name, bases, mmbs):
+        super().__init__(name, bases, mmbs)
+        cls._instance = super().__call__()
+
+    def __call__(cls, *args, **kw):
+        return cls._instance
```
averevki, line 35: You don't use this class anymore

testsuite/tests/kuadrant/authorino/operator/clusterwide/test_all_namespace_api_key.py
```diff
@@ -7,6 +7,8 @@
 
 from testsuite.httpx.auth import HeaderApiKeyAuth
 
+pytestmark = [pytest.mark.authorino, pytest.mark.standalone]
```
averevki, line 10: I'd better explicitly specify which test can run on kuadrant with kuadrant mark instead of specifying which test can't run with it. So every test will contain every mark that this test is able to run with (e.g. limitador will have kuadrant and limitador marks, authorino tests will have kuadrant and authorino marks, but authorino metrics tests will have only authorno mark). Like a tree structure of marks.

pehala: Tests that run on the Kuadrant deployment are much more prevalent than those who do not

testsuite/capabilities.py
```diff
@@ -0,0 +1,56 @@
...
+
+    return True, ""
+
+
+@functools.cache
+def is_standalone():
+    """Return True, if the testsuite is configured to run with envoy in standalone mode, without Gateway API"""
+    if settings.get("gateway_api", True):
+        return False, "Gateway API is turned on and Kuadrant is enabled"
+    return True
```
zkraus, line 36: Missing second `error` value in the return when result is True.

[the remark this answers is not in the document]
pehala: Changed to `standalone` (required a bit more lines and added a bit more complexity to this PR so requires a retest)

Makefile
```diff
@@ -38,23 +38,35 @@ all-is-package:
...
+	$(PYTEST) -n4 -m 'kuadrant_only' --dist loadfile --enforce $(flags) testsuite
 
 performance: ## Run performance tests
 performance: poetry-no-dev
 	$(PYTEST) --performance $(flags) testsuite/tests/kuadrant/authorino/performance
 
 mgc: ## Run mgc tests
 mgc: poetry-no-dev
-	$(PYTEST) --mgc $(flags) testsuite/tests/mgc
+	$(PYTEST) -m "mgc" $(flags) testsuite
```
jsmolar, line 69: --enforce

config/settings.local.yaml.tpl
```diff
@@ -1,7 +1,7 @@
 #default:
 #  skip_cleanup: false
 #  tester: "someuser"                          # Optional: name of the user, who is running the tests, defaults to whoami/uid
-#  gateway_api: true                           # True, if Testsuite should test with Gateway API enabled (e.g. Full Kuadrant) or individual components (e.g. Authorino)
+#  standalone: false                           # True, if Testsuite should test only individual components (e.g. Authorino/limitador operators)
```
jsmolar, line 4: I'd like to see all configurations defining a subset of tests that should run in one place (make). But this is for another PR if needed.

pehala: Agree, will tackle it as part of a Follow-up PR

testsuite/tests/kuadrant/gateway/conftest.py
```diff
@@ -6,6 +6,9 @@
 from testsuite.policy.authorization.auth_policy import AuthPolicy
 
 
+pytestmark = [pytest.mark.kuadrant_only]
```
jsmolar, line 9: conftest with mark

jsmolar (approved):

averevki (approved):

ficap (approved):

PR #292, opened 2023-11-23
Reorganize testsuite modules

pehala:
# Why

Testsuite modules are a mess, with random stuff hidden everywhere. This PR aims to introduce a better structure for all of our code.

# Changes

* Add `testsuite.gateway` package
   * Contains interface for Gateway and related classes and implementation for both Envoy (`testsuite.gateway.envoy`) and GW API ((`testsuite.gateway.gateway_api`) 
* Add `testsuite.policy` package
   * Contains all policies
   * `testsuite.policy.authorization` contains AC and AP and also all the dataclasses that are related to this
* Migrate `testsuite.certificates` from a package to a simple file
   * No longer needed
* Remove `testsuite.objects`
  * Move all classes to their respective packages
* Remove `testsuite.openshift.objects`
  * Move all modules to `testsuite.openshift`


testsuite/oidc/rhsso/__init__.py
```diff
@@ -5,7 +5,7 @@
 from keycloak import KeycloakOpenID, KeycloakAdmin, KeycloakPostError
 
 from testsuite.oidc import OIDCProvider, Token
-from testsuite.objects import LifecycleObject
+from ...lifecycle import LifecycleObject
```
pehala, line 8: this was done automatically, are we ok with this or should I change it to absolute path?

averevki: I would always prefer full paths

pehala: Would you also have an issue with first-level relative imports like `.auth_config`?

averevki: I don't have a problem with "one-dot" imports with such a complex repo structure as ours :+1:

averevki (dismissed):

jsmolar (dismissed):

PR #293, opened 2023-11-24
Fix broken mtls test

averevki:
Fix broken mtls test after https://github.com/Kuadrant/testsuite/pull/259

* Slightly refactor `Result` class to support more errors
* Refactor tls tests to reflect changes in the httpx errors handling
* Fix mtls test


testsuite/httpx/__init__.py
```diff
@@ -30,27 +30,27 @@ def __init__(self, retry_codes, response=None, error=None):
...
-    def has_tls_error(self):
-        """True, if the result failed due to TLS failure"""
-        return (
-            self.has_error()
-            and len(self.error.args) > 0
-            and any("SSL: CERTIFICATE_VERIFY_FAILED" in arg for arg in self.error.args)
-        )
+        return self.has_error("Name or service not known")
+
+    def has_tls_cert_verify_error(self):
```
pehala, line 43: I actually prefer the original solution with pytest.raises, as this hides the actual error we are trying to verify. The reason for has_dns_error and has_tls_error to catch any error occurring when we do not care about the specifics

averevki: I don't think that by calling a method named `has_..._error` we are hiding this error. I see this as an abstraction for error catching that can be used in other tls tests.

testsuite/tests/kuadrant/authorino/operator/tls/mtls/test_mtls_identity.py
```diff
@@ -11,24 +13,22 @@ def test_mtls_success(envoy_authority, valid_cert, hostname):
 
 
 @pytest.mark.parametrize(
-    "cert_authority, certificate, err, err_match",
+    "cert_authority, certificate, has_error",
```
pehala, line 16: What about `check_error`?

testsuite/httpx/__init__.py
pehala, line 43: `has_cert_verify_error`?

testsuite/httpx/__init__.py
```diff
@@ -30,27 +30,27 @@ def __init__(self, retry_codes, response=None, error=None):
...
+        """True, if the result failed due to TLS certificate verification failure"""
+        return self.has_error("SSL: CERTIFICATE_VERIFY_FAILED")
+
+    def has_tls_unknown_ca_error(self):
```
pehala, line 47: `has_unknown_ca_error`?

testsuite/httpx/__init__.py
```diff
@@ -30,27 +30,27 @@ def __init__(self, retry_codes, response=None, error=None):
...
+        """True, if the result failed due to TLS unknown certificate authority failure"""
+        return self.has_error("SSL: TLSV1_ALERT_UNKNOWN_CA")
+
+    def has_tls_cert_required_error(self):
```
pehala, line 51: What about `has_no_cert_provided_error` or something similar. I am not really fond of having tls explicitly mentioned there as it is clear from the description it is TLS related

testsuite/httpx/__init__.py
averevki: Another solution can be to have only the `has_error` method and parse the expected error message to it as a parameter

pehala (approved):

averevki:
^rebase

pehala (approved):

PR #295, opened 2023-11-27
Add test for gateway attached to AuthPolicy

### LINKED ISSUE #269: Add a simple test for AuthPolicy attached to a Gateway
**Summary**

Currently, we bind all of our policies only to HTTPRoute, we should add some tests for AuthPolicy bound to Gateway

**Documentation**

[Gateway API](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute)

**Test cases**

- [ ] Happy path for AP applying to all HTTPRoutes when applying it to Gateway

averevki:
### Test case
AuthPolicy attached straight to gateway

### Changes
- Change authorization `targetRef` parameter name from `route` to `target`
- Add smoke test for AuthPolicy attached straight to gateway

### Note
Also an automatization for the following issue: https://github.com/Kuadrant/kuadrant-operator/pull/287

closes #269

### DIFF
--- a/testsuite/policy/authorization/auth_policy.py
+++ b/testsuite/policy/authorization/auth_policy.py
@@ -18,13 +18,12 @@ class AuthPolicy(AuthConfig):
     def auth_section(self):
         return self.model.spec.setdefault("rules", {})
 
-    # pylint: disable=unused-argument
     @classmethod
-    def create_instance(  # type: ignore
+    def create_instance(
         cls,
         openshift: OpenShiftClient,
         name,
-        route: Referencable,
+        target: Referencable,
         labels: Dict[str, str] = None,
     ):
         """Creates base instance"""
@@ -33,7 +32,8 @@ class AuthPolicy(AuthConfig):
             "kind": "AuthPolicy",
             "metadata": {"name": name, "namespace": openshift.project, "labels": labels},
             "spec": {
-                "targetRef": route.reference,
+                "targetRef": target.reference,
+                "rules": {},
             },
         }
 
--- a/testsuite/tests/kuadrant/gateway/conftest.py
+++ b/testsuite/tests/kuadrant/gateway/conftest.py
@@ -1,0 +1,37 @@
+"""Conftest for gateway tests"""
+import pytest
+
+from testsuite.httpx.auth import HttpxOidcClientAuth
+from testsuite.policy.authorization.auth_policy import AuthPolicy
+
+
+@pytest.fixture(scope="module")
+def kuadrant(kuadrant):
+    """Skip if not running on Kuadrant"""
+    if not kuadrant:
+        pytest.skip("Gateway tests are only for Kuadrant")
+    return kuadrant
+
+
+@pytest.fixture(scope="module")
+def gateway_ready(gateway):
+    """Returns ready gateway"""
+    gateway.wait_for_ready()
+    return gateway
+
+
+@pytest.fixture(scope="module")
+def authorization(gateway_ready, route, oidc_provider, authorization_name, openshift, module_label):
+    # pylint: disable=unused-argument
+    """Create AuthPolicy attached to gateway"""
+    authorization = AuthPolicy.create_instance(
+        openshift, authorization_name, gateway_ready, labels={"testRun": module_label}
+    )
+    authorization.identity.add_oidc("rhsso", oidc_provider.well_known["issuer"])
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def auth(oidc_provider):
+    """Returns RHSSO authentication object for HTTPX"""
+    return HttpxOidcClientAuth(oidc_provider.get_token, "authorization")
--- a/testsuite/tests/kuadrant/gateway/test_basic.py
+++ b/testsuite/tests/kuadrant/gateway/test_basic.py
@@ -1,0 +1,18 @@
+"""Test for AuthPolicy attached directly to gateway"""
+import pytest
+
+
+@pytest.fixture(scope="module")
+def rate_limit():
+    """Basic gateway test doesn't utilize RateLimitPolicy component"""
+    return None
+
+
+@pytest.mark.issue("https://github.com/Kuadrant/kuadrant-operator/pull/287")
+def test_smoke(client, auth):
+    """Test if AuthPolicy attached directly to gateway works"""
+    response = client.get("/get", auth=auth)
+    assert response.status_code == 200
+
+    response = client.get("/get")
+    assert response.status_code == 401
--- a/testsuite/policy/authorization/auth_config.py
+++ b/testsuite/policy/authorization/auth_config.py
@@ -41,7 +41,7 @@ class AuthConfig(OpenShiftObject):
         cls,
         openshift: OpenShiftClient,
         name,
-        route,
+        target,
         labels: Dict[str, str] = None,
     ):
         """Creates base instance"""
@@ -52,7 +52,7 @@ class AuthConfig(OpenShiftObject):
             "spec": {"hosts": []},
         }
         obj = cls(model, context=openshift.context)
-        route.add_auth_config(obj)
+        target.add_auth_config(obj)
         return obj
 
     @modify


testsuite/openshift/objects/gateway_api/route.py
```diff
@@ -73,7 +73,14 @@ def remove_all_hostnames(self):
         self.model.spec.hostnames = []
 
     @modify
-    def set_match(self, backend: "Httpbin", path_prefix: str = None):
+    def set_path_match(self, path_prefix: str):
+        """TODO"""
+        self.model.spec.rules.append(
```
pehala, line 78: `set_backend_match` wouldnt work?

testsuite/tests/kuadrant/conftest.py
```diff
@@ -39,9 +39,7 @@ def authorization_name(blame):
 def authorization(authorino, kuadrant, oidc_provider, route, authorization_name, openshift, module_label):
     """Authorization object (In case of Kuadrant AuthPolicy)"""
     if kuadrant:
-        policy = AuthPolicy.create_instance(openshift, authorization_name, route, labels={"testRun": module_label})
-        policy.identity.add_oidc("rhsso", oidc_provider.well_known["issuer"])
-        return policy
+        return AuthPolicy.create_instance(openshift, authorization_name, route, labels={"testRun": module_label})
```
pehala, line 42: that's a breaking change, what is the reasoning behind this?

testsuite/tests/kuadrant/gateway/test_authpolicy_to_gateway.py
```diff
@@ -0,0 +1,63 @@
...
+def sub_nonexistent_hostname_client(gateway, exposer, blame):
+    """TODO"""
+    hostname = exposer.expose_hostname(blame("sub.nonexistent-hostname"), gateway)
+    client = hostname.client()
+    yield client
+    client.close()
+
+
+def test_authpolicy_attached_to_gateway(client, gateway_httproute, gateway_authorization, nonexistent_hostname_client, sub_nonexistent_hostname_client):
+    """TODO"""
```
pehala, line 50: TODOs :D

testsuite/openshift/objects/gateway_api/route.py
averevki: It will, but I'm not using requests to this endpoint to reach the backend, only requests for 500 and 403, so it is redundant in my opinion. It is also possible to extend the backend match method below, but I wanted to introduce a different interface for this

testsuite/tests/kuadrant/gateway/test_authpolicy_to_gateway.py
```diff
@@ -0,0 +1,60 @@
...
+        - send request to the undefined domain with deny-all AuthPolicy attached to gateway
+    """
+    response = client.get("/get")
+    assert response.status_code == 200
+
+    response = client_to_deny.get("/get")
+    assert response.status_code == 500
+
+    gateway_authorization.commit()
+    sleep(5)
```
pehala, line 57: why sleep?

testsuite/openshift/objects/gateway_api/route.py
```diff
@@ -73,8 +73,13 @@ def remove_all_hostnames(self):
         self.model.spec.hostnames = []
 
     @modify
-    def set_match(self, backend: "Httpbin", path_prefix: str = None):
+    def set_path_match(self, path_prefix: str):
```
pehala, line 76: this is add not set, and is it still used?

testsuite/tests/kuadrant/gateway/test_authpolicy_to_gateway.py
```diff
@@ -0,0 +1,60 @@
...
+    route.add_hostname(wildcard_domain)
+    route.set_path_match("/")
+
+    request.addfinalizer(route.delete)
+    route.commit()
+    return route
+
+
+@pytest.fixture(scope="module")
+def gateway_authorization(request, gateway, authorization_name, openshift, module_label):
```
pehala, line 22: What if you overide authorization for all gateway tests to be at the Gateway level, same with hostname and route?

testsuite/tests/kuadrant/gateway/test_authpolicy_to_gateway.py
pehala: Wouldn't it be better to improve the ready method of AuthPolicy?

testsuite/tests/kuadrant/gateway/test_authpolicy_to_gateway.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_authpolicy_attached_to_gateway(gateway_httproute, gateway_authorization, client, client_to_deny):
+    # pylint: disable=unused-argument
+    """
+    Test if AuthPolicy denying all requests sent to the undefined domains:
+        - send request to the hostname defined
+        - send request to the undefined domain without AuthPolicy attached to gateway
+        - commit deny-all AuthPolicy attached to gateway
+        - send request to the undefined domain with deny-all AuthPolicy attached to gateway
```
pehala, line 48: I am not sold on this being a good test case, what exactly are you trying to test? That AuthPolicy is correctly applied for Gateway or is there something else I am not seeing?

testsuite/tests/kuadrant/gateway/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+@pytest.fixture(scope="module")
+def authorization(gateway, authorization_name, openshift, module_label):
+    """Create AuthPolicy attached to gateway"""
+    return AuthPolicy.create_instance(openshift, authorization_name, gateway, labels={"testRun": module_label})
+
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization):
```
pehala, line 14: why is this override needed?

testsuite/tests/kuadrant/gateway/conftest.py
```diff
@@ -0,0 +1,17 @@
+"""Conftest for gateway tests"""
+import pytest
+
+from testsuite.openshift.objects.auth_config.auth_policy import AuthPolicy
+
+
```
pehala, line 6: There is a missing skip if Kuadrant is not enabled, see operator tests

testsuite/tests/kuadrant/gateway/test_basic.py
```diff
@@ -0,0 +1,9 @@
+"""Test for AuthPolicy attached directly to gateway"""
+import pytest
+
+
+@pytest.mark.issue("https://github.com/Kuadrant/kuadrant-operator/pull/287")
+def test_smoke(client):
+    """Test if AuthPolicy attached directly to gateway works"""
+    response = client.get("/get")
+    assert response.status_code == 200
```
pehala, line 9: This does not test that AuthPolicy is working, it would throw 200 even if it didn't, you should do a more complicated AuthPolicy to truly verify the result (like sending it with and without auth)

testsuite/tests/kuadrant/gateway/conftest.py
averevki: I don't need a second component, `RateLimitPolicy`, here

testsuite/tests/kuadrant/gateway/conftest.py
averevki: I believe we only have skip if Kuadrant IS enabled and this fixture enables Kuadrant tests by default. Authorino operator tests get disabled if Kuadrant is enabled. Can you please send a link to what you referring to?

pehala: Sorry, you are right, The correct example is in limitador tests

testsuite/tests/kuadrant/gateway/conftest.py
pehala: I would override Ratelimit fixture and return None then, we do not need to limit ourselves to AuthPolicy in all gateway tests

averevki: Can't it be done in the feature tests then? Overriding commit for this purpose looks like the conventional way in the testsuite now.

pehala: Well, the issue is not that you are rewriting commit (even though overriding the ratelimit fixture is much much nicer), the issue is you are doing it in the conftest for all future gateway tests

testsuite/tests/kuadrant/gateway/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+def gateway_wait_for_ready(gateway):
+    """Waits for gateway to be ready"""
+    gateway.wait_for_ready()
+
+
+@pytest.fixture(scope="module", autouse=True)
+def authorization_wait_for_ready(gateway_wait_for_ready, commit, route, authorization):
+    # pylint: disable=unused-argument
+    """Waits for authorization to be ready"""
+    authorization.wait_for_ready()
```
pehala, line 26: You can add it to the commit fixture in on of the upper contest, so every test waits for ready

averevki:
Dependent on https://github.com/Kuadrant/testsuite/pull/305

pehala (approved):

testsuite/policy/authorization/auth_policy.py
```diff
@@ -18,13 +18,12 @@ class AuthPolicy(AuthConfig):
     def auth_section(self):
         return self.model.spec.setdefault("rules", {})
```
azgabur, line 19: If the rules object is defined as `{}` already in `create_instance()` would it be possible to remove the `setdefault`?
```suggestion
        return self.model.spec.rules
```

testsuite/tests/kuadrant/gateway/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@pytest.fixture(scope="module", autouse=True)
+def gateway_wait_for_ready(gateway):
+    """Waits for gateway to be ready"""
+    gateway.wait_for_ready()
+
+
+@pytest.fixture(scope="module")
+def authorization(gateway_wait_for_ready, gateway, route, oidc_provider, authorization_name, openshift, module_label):
```
ficap, line 23: I do see the reason to explicitly request `gateway_wait_for_ready` fixture but at the same time I don't see why the fixture is autouse at the same time.
Another way how to refactor it would be to create a `ready_gateway` fixture returning a Gateway on which the `wait_for_ready()` was applied and then request `ready_gateway` instead of `gateway_wait_for_ready` and `gateway` simultaneously.

testsuite/policy/authorization/auth_policy.py
averevki: I left it in case of `rules` key doesn't show in the object at all if it is empty (tell me if I'm wrong and it's not possible). I think it is safer to leave the default here for now

testsuite/tests/kuadrant/gateway/conftest.py
averevki: Good suggestion :+1:

ficap (approved):

pehala (approved):

testsuite/policy/authorization/auth_policy.py
azgabur: I think the `rules` key should always exist if its defined in `model` in `create_instance()` which is run on creation of every `AuthPolicy` object. But the current state is not wrong either, I just think its redundant. 
This PR can be merged without changing this.

azgabur: I run `make test` on your PR with this modification and all tests passed (except the wildcard test error which is caused by something else).

PR #298, opened 2023-11-28
Fix http package name collision

averevki:
### Bug
Can't call pycharm debuger inside `authorino.operator` package because of name collision with `http.client` (`authorino.operator` also have package named `http` containing tests for raw http requests)

### Solution
Rename `http` package to `http_`

### Note
This is conventional renaming for the name clashing. I'm open to other propositions


averevki:
> The proper solution is to also the working directory, but if you really want to rename it, I would liked more if you choose a better name

After consideration, I changed it to `raw_http`, which also, I believe, would be a better name for this package

pehala (approved):

ficap (approved):

PR #302, opened 2023-11-29
Improve string representation of Result

pehala:


testsuite/httpx/__init__.py
```diff
@@ -58,6 +58,11 @@ def __getattr__(self, item):
             return getattr(self.response, item)
         return None
 
+    def __str__(self):
```
averevki, line 61: You probably want to use `__repr__` here, if I understand use-case correctly

testsuite/httpx/__init__.py
```diff
@@ -127,7 +132,7 @@ def request(
                 extensions=extensions,
             )
             return Result(self.retry_codes, response=response)
-        except ConnectError as e:
+        except (ConnectError, RemoteProtocolError) as e:
```
averevki, line 135: what about just catching all `TransportError`'s? https://www.python-httpx.org/exceptions/

averevki (approved):

jsmolar (approved):

PR #303, opened 2023-12-04
Move redirect test under Responses module 

azgabur:
Moving the `test_redirect.py` under `response` module as its more fitting because the test is using "responses" part of AuthConfig.


pehala:
Same as #298, this is not a bug in testsuite. You have the wrong working directory, this is an error on your side

azgabur:
^Rebase on main

pehala (approved):

averevki (approved):

PR #304, opened 2023-12-06
Add limit reset test

jsmolar:
Do you think this test is needed? We test limit reset in multiple iterations.
In addition, I decreased the duration of limits by 5 sec (This decreases the runtime of this module by 60sec total).


averevki:
In my opinion limit reset is covered enough in multiple iterations test. I like limits duration change though :+1:

testsuite/tests/kuadrant/limitador/test_basic_limit.py
```diff
@@ -34,3 +35,11 @@ def test_limit(client, limit):
         r.status_code == 200 for r in responses
     ), f"Rate Limited resource unexpectedly rejected requests {responses}"
     assert client.get("/get").status_code == 429
+
+
+def test_limit_reset(client, limit):
```
pehala, line 40: Wouldn't this collide with the previous test when run in parallel?

jsmolar: It would. Scope change to function is required. But I am inclined to remove this test.

averevki (approved):

PR #305, opened 2023-12-06
Add kuadrant components wait-for-ready

averevki:
### Changes

- Add wait-for-ready methods for authorization and `RateLimitPolicy` components
- Add wait-for-ready method calls after the kuadrants components commit
- Add uma test dependency for client secret so it is created before authorization using it

### DIFF
--- a/testsuite/policy/rate_limit_policy.py
+++ b/testsuite/policy/rate_limit_policy.py
@@ -51,18 +51,15 @@ class RateLimitPolicy(OpenShiftObject):
             limit["counters"] = counters
         self.model.spec.limits[name] = limit
 
-    def commit(self):
-        result = super().commit()
-
-        # wait for RLP to be actually applied, conditions itself is not enough, sleep is needed
-        def _policy_is_ready(obj):
-            return "conditions" in obj.model.status and obj.model.status.conditions[0].status == "True"
-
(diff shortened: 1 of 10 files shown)


testsuite/policy/rate_limit_policy.py
```diff
@@ -51,18 +51,15 @@ def add_limit(self, name, limits: Iterable[Limit], when: Iterable[Rule] = None,
...
-
+    def wait_for_ready(self):
+        """Wait for RLP to be actually applied, conditions itself is not enough, sleep is needed"""
         with oc.timeout(90):
-            success, _, _ = self.self_selector().until_all(success_func=_policy_is_ready, tolerate_failures=5)
-            assert success
+            success, _, _ = self.self_selector().until_all(
+                success_func=lambda obj: "conditions" in obj.model.status
+                and obj.model.status.conditions[0].status == "True",
+                tolerate_failures=5,
```
pehala, line 60: Why is there `tolerate_failures` when it is not present in other waits?

averevki: I just copied it from the previous `RLP` commit function

pehala (approved):

azgabur (approved):

PR #310, opened 2023-12-20
Test DNSPolicy health check additional headers

### LINKED ISSUE #252: Add test for DNSPolicy additionalHeadersRef
**Summary**

The TLSPolicy additionalHeadersRef field specifies a Secret used for storing supplementary HTTP headers. These headers are included when sending probe requests and can contain critical information like authentication tokens.

**Documentation**

- DNSPolicy Health Checks: https://docs.kuadrant.io/multicluster-gateway-controller/docs/dnspolicy/dns-health-checks/
- additionalHeadersRef MGC PR: https://github.com/Kuadrant/multicluster-gateway-controller/pull/409

**Test cases**

- Test for DNSPolicy additionalHeadersRef field

averevki:
### Test case
DNSPolicy health check additional headers

### Changes
- Add MockserverBackend object
- Add request expectation to the mockserver
- Add test for DNSPolicy health check additional headers

closes #252

### DIFF
--- a/testsuite/mockserver.py
+++ b/testsuite/mockserver.py
@@ -2,10 +2,15 @@
 
 from typing import Union
 
-import httpx
 from apyproxy import ApyProxy
 
 from testsuite.utils import ContentType
+from testsuite.httpx import KuadrantClient
+from testsuite.openshift import Selector
+from testsuite.openshift.backend import Backend
+from testsuite.openshift.service import Service, ServicePort
+from testsuite.openshift.deployment import Deployment, ContainerResources
+from testsuite.openshift.client import OpenShiftClient
 
 
 class Mockserver:
@@ -13,20 +18,36 @@ class Mockserver:
     Mockserver deployed in Openshift (located in Tools or self-managed)
     """
 
-    def __init__(self, url):
-        self.client = ApyProxy(url, session=httpx.Client(verify=False, timeout=5))
+    def __init__(self, url, client: KuadrantClient = None):
+        self.client = ApyProxy(url, session=client or KuadrantClient(verify=False))
 
-    def _expectation(self, expectation_id, response_data):
+    def _expectation(self, expectation_id, json_data):
         """
-        Creates an Expectation with given response_data.
+        Creates an Expectation from given expectation json.
         Returns the absolute URL of the expectation
         """
-        json_data = {"id": expectation_id, "httpRequest": {"path": f"/{expectation_id}"}}
-        json_data.update(response_data)
+        json_data["id"] = expectation_id
+        json_data.setdefault("httpRequest", {})["path"] = f"/{expectation_id}"
 
         self.client.mockserver.expectation.put(json=json_data)
         # pylint: disable=protected-access
         return f"{self.client._url}/{expectation_id}"
+
+    def create_request_expectation(
+        self,
+        expectation_id,
+        headers: dict[str, list[str]],
+    ):
+        """Creates an Expectation - request with given headers"""
+        json_data = {
+            "httpRequest": {
+                "headers": headers,
+            },
+            "httpResponse": {
+                "body": "",
+            },
+        }
+        return self._expectation(expectation_id, json_data)
 
     def create_expectation(
         self,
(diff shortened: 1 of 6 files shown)


testsuite/tests/mgc/dnspolicy/health_check/test_additional_headers.py
```diff
@@ -0,0 +1,71 @@
...
+        port=80,
+        protocol="http",
+    )
+
+
+@pytest.fixture(scope="module")
+def reference_grant(request, testconfig, gateway, module_label, blame):
+    """Creates ReferenceGrant object for cross namespace access to Mockserver service"""
+    # get tools project client if tools installed on a spoke cluster
+    tools_openshift = gateway.openshift.change_project(testconfig["tools"].project)
```
pehala, line 49: this is currently unsupported as we expect tools to be deployed only once, it can be changed even in this PR, similarly to how service protection project and project2 are handler

testsuite/mockserver.py
```diff
@@ -4,30 +4,60 @@
...
 
-    def __init__(self, url):
+    def __init__(self, url, service_name: str, tools: OpenShiftClient):
         self.client = ApyProxy(url, session=httpx.Client(verify=False, timeout=5))
+        self.service_name = service_name
+        self.tools = tools
 
-    def _expectation(self, expectation_id, response_data):
+    @property
+    def reference(self):
```
pehala, line 23: this does not feel right, as the mockserver itself is just a client to a any Mockserver even deployed without openshift.

testsuite/mockserver.py
```diff
@@ -4,30 +4,60 @@
...
 
 
-class Mockserver:
+class Mockserver(Referencable):
     """
     Mockserver deployed in Openshift (located in Tools or self-managed)
     """
 
-    def __init__(self, url):
+    def __init__(self, url, service_name: str, tools: OpenShiftClient):
```
pehala, line 17: There should definitely not be a tools requirement for this class

pehala (changes requested):
Overall too many changes for one additional test only due to a refactor. I would prefer to do a separate PR for the rename or keep the name for now

testsuite/tests/mgc/dnspolicy/health_check/test_additional_headers.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_additional_headers(dns_health_probe, mockserver, module_label):
+    """Test if additional headers in health check requests are used"""
+    assert dns_health_probe.is_healthy()
+
+    requests = mockserver.retrieve_requests(module_label)
+    assert len(requests) > 0
+
+    request_headers = requests[0]["headers"]
+    assert request_headers.get(HEADER_NAME) == [HEADER_VALUE]
```
pehala, line 71: Wouldn't this httpbin endpoint work as well? `https://httpbin.org/#/Request_inspection/get_headers`

averevki: It is only possible to check current request headers with httpbin, while the test requires checking requests with headers sent by dns health checks

testsuite/mockserver.py
pehala: I would suggest creating another Backend implementation for mockserver and replacing backend fixture

averevki: > this does not feel right, as the mockserver itself is just a client to a any Mockserver even deployed without openshift.

but our tests will always use mockserver deployed inside the k8s as part of the tools, am I missing something?

> I would suggest creating another Backend implementation for mockserver and replacing backend fixture

I think another mockserver implementation is unnecessary here, it creates additional overhead and everything backend mockserver require is a reference to his service. And backend fixture has already been replaced

pehala: > but our tests will always use mockserver deployed inside the k8s as part of the tools, am I missing something?

Tools can be used from a different cluster or you can just use external mockserver as well

> I think another mockserver implementation is unnecessary here, it creates additional overhead and everything backend mockserver require is a reference to his service. And backend fixture has already been replaced

You would either deploy httpbin or mockserver, I do not see a problem, besides if we agree that mockserver always need to be on the same cluster than you do not have to deploy it.

averevki:
rebase^

testsuite/mockserver.py
```diff
@@ -14,21 +19,37 @@ class Mockserver:
...
+            "httpRequest": {
+                "headers": headers,
+            },
+            "httpResponse": {
+                "body": "",
+            },
+        }
+        return self._expectation(expectation_id, json_data)
+
+    def create_response_expectation(
```
pehala, line 52: What about keeping the original name and just adding `create_request_expectation`

averevki:
rebase^

testsuite/tests/mgc/dnspolicy/health_check/test_additional_headers.py
```diff
@@ -0,0 +1,69 @@
...
+    request.addfinalizer(mockserver.delete)
+    mockserver.commit()
+
+    return mockserver
+
+
+@pytest.fixture(scope="module")
+def mockserver_client(client):
+    """Returns Mockserver client"""
+    return Mockserver(str(client.base_url), client=client)
```
pehala, line 28: If URL is dependent on the client, why not require only client?

pehala (approved):

martinhesko:
I tried running this myself but ran into some issues with `DNSHealthCheckProbe` not having any status. First only on `gcp`, later runs on both DNS providers

averevki:
rebase^

martinhesko (approved):

PR #311, opened 2023-12-21
Added backoff to dns health probe

jsmolar:


averevki:
This may not work. There were a few problems with the initial probe creation - 1) It wasn't created for some time (your PR resolves this issue) 2) Health check status wasn't appearing in the object at the beginning 3) Health check status was false right after the status field addition (need to wait a little bit). Creation was flaky so in different runs, it would fail for different reasons

jsmolar:
I had a problem described in your point 1, so I shared a solution that fixed it. 
We need to investigate your 2 and 3 points to stabilize MGC tests further.

testsuite/policy/dns_policy.py
```diff
@@ -57,6 +59,7 @@ def set_health_check(self, health_check: HealthCheck):
         """Sets health check for DNSPolicy"""
         self.model["spec"]["healthCheck"] = asdict(health_check)
 
+    @backoff.on_exception(backoff.fibo, OpenShiftPythonException, max_tries=13, jitter=None)
```
zkraus, line 62: How time consuming is oc.selector() call?

potential time spent here is 13*call_elapsed + 144 [sec], which might be quite a lot.

PR #312, opened 2023-12-22
Fixed response code check in Result class

jsmolar:


testsuite/httpx/__init__.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, retry_codes, response=None, error=None):
 
     def should_backoff(self):
         """True, if the Result can be considered an instability and should be retried"""
-        return self.has_dns_error() or (self.error is None and self.status_code in self.retry_codes)
+        return self.has_dns_error() or (self.error is None and self.response.status_code in self.retry_codes)
```
pehala, line 33: This should work, due to `getattr`, I do not know what must happen that self.status_code does not work and self.response.status_code does

PR #313, opened 2023-12-22
Dinosaur conftest refactor to support AuthPolicy v2

azgabur:
Resulting AuthConfig is almost identical to currently used [dinosaur_config.yaml](https://github.com/Kuadrant/testsuite/blob/89dfb987cc075fc091320a8aaea63ef57c9365be/testsuite/resources/dinosaur_config.yaml) with few exceptions:

- New AuthConfig contains `credentials.authorizationHeader.prefix: Bearer` in both oidc identity providers objects which is automatically added by testsuite while old AuthConfig has empty `credentials` object.
- In new AuthConfig `metadata.http.<name>.contentType` has value `application/x-www-form-urlencoded` and in headers it specifies the contentType `application/json` while old AuthConfig has `contentType` value set to `application/json`
- I removed debugging `responses` headers denoted as "TMP" in old dinosaur_config.yaml

This PR implements `ttl` functionality in add oidc identity function. Testing case for this new feature will be done #314


testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -86,49 +78,155 @@ def _resource_info(org_id, owner):
...
+
+    authorization.identity.clear_all()
+    authorization.identity.add_oidc(
+        "redhat-sso",
+        rhsso.well_known["issuer"],
+        ttl=3600,
+        defaults_properties={"org_id": ValueFrom("auth.identity.middle_name")},
+    )
+    authorization.identity.add_oidc(
+        "admin-sso", admin_rhsso.well_known["issuer"], ttl=3600, when=[PatternRef("admin-route")]
```
pehala, line 123: I believe the TTL was not tested for this exact use case, do you think it would make sense to add a test case for it also here?

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -86,49 +78,155 @@ def _resource_info(org_id, owner):
...
+            "redhat-sso": [Pattern("auth.identity.iss", "eq", rhsso.well_known["issuer"])],
+            "admin-sso": [Pattern("auth.identity.iss", "eq", admin_rhsso.well_known["issuer"])],
+            "require-org-id": [Pattern("auth.identity.org_id", "neq", "")],
+        }
+    )
+    authorization.add_rule([PatternRef("api-route"), PatternRef("v1-route")])
+
+    authorization.identity.clear_all()
+    authorization.identity.add_oidc(
+        "redhat-sso",
```
pehala, line 117: what about naming it `user-sso`?

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
azgabur: The new testcase can be implemented as part of #314. Added a new task to the issue.

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
azgabur: That is better fitting name. Renamed.

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -86,49 +78,155 @@ def _resource_info(org_id, owner):
...
+allow { method == "GET";    roles[_] == "admin-read" }
+allow { method == "GET";    roles[_] == "admin-write" }
+allow { method == "PATCH";  roles[_] == "admin-full" }
+allow { method == "PATCH";  roles[_] == "admin-write" }
+allow { method == "DELETE"; roles[_] == "admin-full" }
+""",
+        when=[PatternRef("admin-route"), PatternRef("admin-sso")],
+    )
+    authorization.authorization.add_opa_policy(
+        "admin-redhat-rhsso",
```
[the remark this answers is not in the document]
azgabur: Renamed

pehala (approved):

jsmolar (approved):

PR #316, opened 2024-01-04
Add Route Select for Rate Limit

jsmolar:
This PR creates two new sets of objects:

- `RouteSelect`: https://docs.kuadrant.io/kuadrant-operator/doc/reference/route-selectors/#routeselector
- `HTTP Route Matchers`: https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteMatch

The placement of these resources can be discussed as I was not 100% sure.
Tests will be added later to this PR.

Usage example:
```
   selector = RouteSelector(
        RouteMatch(
            path=HTTPPathMatch(type=MatchType.EXACT, value="/get"),
            headers=[HeadersMatch(name="header", value="value"), HeadersMatch(name="another", value="value")],
            query_params=[HTTPQueryParamsMatch(name="param", value="value")]
            method=HTTPMethod.GET,
        )
    )
    rate_limit.add_limit("multiple", [limit], route_selectors=[selector])
```

### DIFF
--- a/testsuite/gateway/__init__.py
+++ b/testsuite/gateway/__init__.py
@@ -1,5 +1,4 @@
 """Classes related to Gateways"""
-import abc
 import enum
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -9,7 +8,7 @@ from httpx import Client
 
 from testsuite.certificates import Certificate
 from testsuite.lifecycle import LifecycleObject
-from testsuite.utils import asdict, _asdict_recurse
+from testsuite.utils import asdict
 
 if TYPE_CHECKING:
     from testsuite.openshift.client import OpenShiftClient
@@ -76,10 +75,6 @@ class PathMatch:
     type: Optional[MatchType] = None
     value: Optional[str] = None
 
-    # def asdict(self):
-    #     """Custom dict due to nested structure of matchers."""
-    #     return {"path": _asdict_recurse(self, False)}
-
 
 @dataclass
 class HeadersMatch:
@@ -89,10 +84,6 @@ class HeadersMatch:
     value: str
     type: Optional[Literal[MatchType.EXACT, MatchType.REGULAR_EXPRESSION]] = None
 
-    # def asdict(self):
-    #     """Custom dict due to nested structure of matchers."""
-    #     return {"headers": [_asdict_recurse(self, False)]}
-
 
 @dataclass
 class QueryParamsMatch:
@@ -101,23 +92,6 @@ class QueryParamsMatch:
     name: str
     value: str
     type: Optional[Literal[MatchType.EXACT, MatchType.REGULAR_EXPRESSION]] = None
-
-    # def asdict(self):
-    #     """Custom dict due to nested structure of matchers."""
-    #     return {"queryParams": [_asdict_recurse(self, False)]}
-
-
-@dataclass
-class MethodMatch:
-    """
-    HTTPMethod describes how to select a HTTP route by matching the HTTP method. The value is expected in upper case.
-    """
-
-    value: HTTPMethod = None
-
-    def asdict(self):
-        """Custom dict due to nested structure of matchers."""
-        return {"method": self.value.value}
 
 
 @dataclass
@@ -132,7 +106,7 @@ class RouteMatch:
     path: Optional[PathMatch] = None
     headers: Optional[List[HeadersMatch]] = None
     query_params: Optional[List[QueryParamsMatch]] = None
-    method: HTTPMethod = None
+    method: Optional[HTTPMethod] = None
 
 
 class Gateway(LifecycleObject, Referencable):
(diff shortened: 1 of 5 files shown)


testsuite/policy/rate_limit_policy.py
```diff
@@ -21,6 +21,17 @@ class Limit:
     unit: Literal["second", "minute", "day"] = "second"
 
 
+@dataclass
+class RouteSelect:
```
pehala, line 25: `RouteSelector`

testsuite/gateway/__init__.py
```diff
@@ -1,13 +1,15 @@
...
 from dataclasses import dataclass
-from typing import Any, Optional, TYPE_CHECKING
+from typing import Any, Optional, TYPE_CHECKING, Literal
 
 from httpx import Client
 
 from testsuite.certificates import Certificate
 from testsuite.lifecycle import LifecycleObject
-from testsuite.utils import asdict
+from testsuite.utils import asdict, _asdict_recurse
```
ficap, line 12: In this situation I would consider making `_asdict_recurse` a public function as it seems that more and more code depends on it.

jsmolar: Removed in last PR update

testsuite/utils.py
```diff
@@ -138,9 +138,6 @@ def asdict(obj) -> dict[str, JSONValues]:
 
 
 def _asdict_recurse(obj):
-    if hasattr(obj, "asdict"):
```
pehala, line 141: I think changes to this  function can be reverted

jsmolar: removed

testsuite/policy/rate_limit_policy.py
```diff
@@ -21,6 +21,23 @@ class Limit:
     unit: Literal["second", "minute", "day"] = "second"
 
 
+@dataclass
+class RouteSelector:
+    """
+    RouteSelector is an object composed of a set of HTTPRouteMatch objects (from Gateway API -
+    HTRTPPathMatch, HTTPHeaderMatch, HTTPQueryParamMatch, HTTPMethodMatch),
```
ficap, line 28: typo: `HTRTPPathMatch` -> `HTTPPathMatch

testsuite/gateway/gateway_api/route.py
```diff
@@ -73,17 +74,22 @@ def remove_all_hostnames(self):
...
-                    return
-        raise NameError("This backend is not assigned to this Route")
+    def add_rule(self, backend: "Httpbin", *route_matches: RouteMatch):
+        """Adds rule to the Route"""
+        rules = {"backendRefs": [backend.reference]}
+        matches = (
+            list(route_matches)
+            if route_matches
+            else [RouteMatch(path=PathMatch(type=MatchType.PATH_PREFIX, value="/"))]
+        )
```
pehala, line 84:
```suggestion
   matches = list(route_matches)
   if len(matches) == 0:
       matches.append(RouteMatch(path=PathMatch(type=MatchType.PATH_PREFIX, value="/"))
```

testsuite/gateway/gateway_api/route.py
```diff
@@ -73,17 +74,22 @@ def remove_all_hostnames(self):
...
+
+        rules.update({"matches": list(map(asdict, matches))})
```
pehala, line 86:
```suggestion
        rules["matches"] = [asdict(match) for match in matches]
```

testsuite/gateway/__init__.py
```diff
@@ -45,6 +46,69 @@ def reference(self) -> dict[str, Any]:
...
+
+    name: str
+    value: str
+    type: Optional[Literal[MatchType.EXACT, MatchType.REGULAR_EXPRESSION]] = None
+
+
+@dataclass
+class RouteMatch:
+    """
+    Abstract Dataclass for HTTPRouteMatch.
```
pehala, line 100: Description no longer valid

testsuite/gateway/gateway_api/route.py
jsmolar: I somehow agree with your suggestion.
But the thing to discuss here is that if no `route_matches` is used for the rule,
```
        - path:
            type: PathPrefix
            value: /
```
is used automatically. No need to specifically defined `RouteMatch(path=PathMatch(type=MatchType.PATH_PREFIX, value="/"))`

pehala: in that case, the condition does not need to be there

pehala (approved):

testsuite/gateway/__init__.py
```diff
@@ -1,7 +1,8 @@
 """Classes related to Gateways"""
+import enum
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Any, Optional, TYPE_CHECKING
+from typing import Any, Optional, TYPE_CHECKING, Literal, List
```
averevki, line 5: We're moving to `list` from a standard python library. Not a PR blocker though

averevki (approved):
LGTM. Is this a setup for tests in the feature PR, or do you want to also add them here?

PR #337, opened 2024-02-07
Replace standalone config option for --standalone flag

pehala:
Makes configuring standalone run to be all done through flags instead of config + flag

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ authorino: poetry-no-dev
 
 authorino-standalone: ## Run only test capable of running with standalone Authorino
 authorino-standalone: poetry-no-dev
-	$(PYTEST) -n4 -m 'authorino and not kuadrant_only' --dist loadfile --enforce $(flags) testsuite/tests/kuadrant/authorino
+	$(PYTEST) -n4 -m 'authorino and not kuadrant_only' --dist loadfile --enforce --standalone $(flags) testsuite/tests/kuadrant/authorino
 
 limitador: ## Run only Limitador related tests
 limitador: poetry-no-dev
--- a/config/settings.local.yaml.tpl
+++ b/config/settings.local.yaml.tpl
@@ -1,7 +1,6 @@
 #default:
 #  skip_cleanup: false
 #  tester: "someuser"                          # Optional: name of the user, who is running the tests, defaults to whoami/uid
-#  standalone: false                           # True, if Testsuite should test only individual components (e.g. Authorino/limitador operators)
 #  cluster:                                    # Workload cluster where tests should run, will get overriden if run on Multicluster
 #    project: "kuadrant"                       # Optional: Default namespace for this cluster
 #    api_url: "https://api.openshift.com"      # Optional: OpenShift API URL, if None it will OpenShift that you are logged in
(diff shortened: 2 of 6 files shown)


azgabur (approved):

Makefile
```diff
@@ -50,7 +50,7 @@ authorino: poetry-no-dev
 
 authorino-standalone: ## Run only test capable of running with standalone Authorino
 authorino-standalone: poetry-no-dev
-	$(PYTEST) -n4 -m 'authorino and not kuadrant_only' --dist loadfile --enforce $(flags) testsuite/tests/kuadrant/authorino
+	$(PYTEST) -n4 -m 'authorino and not kuadrant_only' --dist loadfile --enforce --standalone $(flags) testsuite/tests/kuadrant/authorino
```
averevki, line 53: 3 marking flags for the simple authorino without kuadrant tests. Seems unnecessary

config/settings.local.yaml.tpl
```diff
@@ -1,7 +1,6 @@
 #default:
 #  skip_cleanup: false
 #  tester: "someuser"                          # Optional: name of the user, who is running the tests, defaults to whoami/uid
-#  standalone: false                           # True, if Testsuite should test only individual components (e.g. Authorino/limitador operators)
```
averevki, line 4: I'd like to also have a standalone option in the settings file, at least for now.

averevki (commented):
It looks like `--standalone` flag doing the same functionality as `-m "not kuadrant_only"`. And still, it is only possible to use it with another "standalone" flag.

config/settings.local.yaml.tpl
pehala: Why?

testsuite/tests/conftest.py
```diff
@@ -43,15 +46,19 @@ def pytest_runtest_setup(item):
...
             skip_func(f"Unable to locate Kuadrant installation: {error}")
     if "standalone_only" in marks:
-        status, error = is_standalone()
-        if not status:
-            skip_func(f"Unable to run Standalone tests: {error}")
+        if not standalone:
+            skip_func("Unable to run Standalone only test: Standalone mode is disabled, please use --standalone flag")
     if "mgc" in marks:
+        if standalone:
+            skip_func("Unable to run MGC test: Standalone mode is enabled")
```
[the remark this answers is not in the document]
pehala: MGC can never run on standalone and if I explicitly set that I want standalone I want it to not execute MGC tests

config/settings.local.yaml.tpl
averevki: Well, I guess it doesn't matter now as we started using the pytest marks anyway. But I don't like that I need to change my code editor settings for every test that I'm running. With the settings file this change is only a true/false, without - I should remember every flag every time I change my testing scope. But from a full testsuite run through makefile perspective, it is of course better to have every option as a flag. I hope there are some presets in Pycharm for this, I probably can get used to it. 

Can't settings file just contain an option preset and flag is always overriding?

pehala: With the settings, it is too easy to mistakenly leave it on and think it is running Kuadrant with Gateway API and instead it can be standalone, you do not have an easy way of telling and all would be green anyway. This makes it much more explicit and doesn't allow mistakes.

jsmolar (approved):

averevki (approved):

jsmolar (approved):

PR #340, opened 2024-02-07
Always run Kuadrant check

pehala:
* Run Kuadrant check every time in the beginning, unless there is a `--standalone` flag
* `make test` now has `--enforce` flag
* Run Kuadrant check against default cluster, if no spokes are configured

### DIFF
--- a/testsuite/capabilities.py
+++ b/testsuite/capabilities.py
@@ -10,17 +10,24 @@ from testsuite.config import settings
 @functools.cache
 def has_kuadrant():
     """Returns True, if Kuadrant deployment is present and should be used"""
-    spokes = weakget(settings)["control_plane"]["spokes"] % {}
+    if spokes := weakget(settings)["control_plane"]["spokes"] % {}:
+        for name, openshift in spokes.items():
+            # Try if Kuadrant is deployed
+            if not openshift.connected:
+                return False, f"Spoke {name} is not connected"
+            project = settings["service_protection"]["system_project"]
+            kuadrant_openshift = openshift.change_project(project)
+            kuadrants = kuadrant_openshift.do_action("get", "kuadrant", "-o", "json", parse_output=True)
+            if len(kuadrants.model["items"]) == 0:
+                return False, f"Spoke {name} does not have Kuadrant resource in project {project}"
 
-    for name, openshift in spokes.items():
-        # Try if Kuadrant is deployed
-        if not openshift.connected:
-            return False, f"Spoke {name} is not connected"
+    else:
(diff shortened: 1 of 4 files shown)


testsuite/capabilities.py
```diff
@@ -10,18 +10,34 @@
...
 def has_kuadrant():
     """Returns True, if Kuadrant deployment is present and should be used"""
-    spokes = weakget(settings)["control_plane"]["spokes"] % {}
-
-    for name, openshift in spokes.items():
-        # Try if Kuadrant is deployed
-        if not openshift.connected:
-            return False, f"Spoke {name} is not connected"
+    if "spokes" in weakget(settings)["control_plane"] % {}:
+        spokes = settings["control_plane"]["spokes"]
```
averevki, line 14: ^this will also trigger on `spokes: {}` in the settings
```suggestion
    if spokes := weakget(settings)["control_plane"]["spokes"] % {}:
```

testsuite/tests/conftest.py
```diff
@@ -256,12 +255,6 @@ def kuadrant(request, testconfig, openshift):
...
         return None
 
-    # Try if Kuadrant is deployed
-    kuadrant_openshift = testconfig["service_protection"]["system_project"]
-    kuadrants = kuadrant_openshift.do_action("get", "kuadrant", "-o", "json", parse_output=True)
-    if len(kuadrants.model["items"]) == 0:
-        pytest.fail("Running Kuadrant tests, but Kuadrant resource was not found")
-
     # TODO: Return actual Kuadrant object
     return True
```
averevki, line 259: Looks like you don't need `testconfig` and `openshift` fixtures here anymore (badly placed unused-parameter 😬)

testsuite/capabilities.py
```diff
@@ -10,18 +10,34 @@
...
         kuadrants = kuadrant_openshift.do_action("get", "kuadrant", "-o", "json", parse_output=True)
         if len(kuadrants.model["items"]) == 0:
-            return False, f"Spoke {name} does not have Kuadrant resource in project {project}"
+            return False, f"Kuadrant resource is not installed in project {project}"
+
+    return True, None
+
 
+@functools.cache
+def is_standalone():
```
averevki, line 37: We can add this function later when it will have an actual usage

testsuite/tests/conftest.py
pehala: Removed the fixtures, and kept the unused-argument. There were other issue related to that

averevki (approved):

jsmolar (approved):

PR #343, opened 2024-02-08
Refactor many request assert for limit tests

jsmolar:


testsuite/httpx/__init__.py
```diff
@@ -65,6 +65,15 @@ def __str__(self):
         return f"Result[error={self.error}]"
 
 
+class ResultList(list):
+    """List-like object for Result"""
+
+    def assert_all(self, status_code=None):
```
averevki, line 71:
```suggestion
    def assert_all(self, status_code):
```

averevki (commented):
What about `ResultOne` and `ResultMany`?

pehala:
> What about `ResultOne` and `ResultMany`?

I think `ResultList` is a better name

testsuite/httpx/__init__.py
```diff
@@ -65,6 +65,15 @@ def __str__(self):
...
+        """Assert all responses that contain certain status code"""
+        for request in self:
+            assert request.status_code == status_code, f"Rate Limited resource unexpectedly rejected request: {request}"
```
pehala, line 74: I would improve the error message wording and to also include index of the request that is wrong

testsuite/httpx/__init__.py
jsmolar: agree, we can add named parameters if/when more asserts is needed.

testsuite/httpx/__init__.py
```diff
@@ -65,6 +65,18 @@ def __str__(self):
...
 
+class ResultList(list):
+    """List-like object for Result"""
+
+    def assert_all(self, status_code):
+        """Assert all responses that contain certain status code"""
+        for request in self:
+            assert request.status_code == status_code, (
+                f"Status code assertion failed for request {self.index(request)+1} out of {len(self)} requests: "
+                f"{request} != {status_code}"
```
azgabur, line 76:
```suggestion
                f"{request.status_code} != {status_code}"
```

testsuite/tests/kuadrant/limitador/test_route_rule.py
```diff
@@ -33,10 +33,16 @@ def rate_limit(rate_limit):
 
 def test_rule(client):
     """Tests that RLP correctly applies to the given HTTPRoute rule"""
-    responses = client.get_many("/get", 5)
-    assert all(
-        r.status_code == 200 for r in responses
-    ), f"Rate Limited resource unexpectedly rejected requests {responses}"
+    responses = client.get_many("/get", 7)
```
azgabur, line 36: Why the change to 7? The limit is set to 5.

testsuite/httpx/__init__.py
azgabur: If the response status code can be seen in the error message than its okay

testsuite/tests/kuadrant/limitador/test_route_rule.py
azgabur: Also in the function `test_rule_missmatch`

jsmolar: `test_rule_missmatc` is new test method where 7 is intended

azgabur (approved):

pehala (approved):

PR #355, opened 2024-02-27
Fix authorino clusterwide tests

averevki:
First authconfig now contains only first hostname, second authconfig - only second hostname.
The wildcard collision test is divided into 2 separate tests

### DIFF
--- a/testsuite/tests/kuadrant/authorino/operator/clusterwide/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/operator/clusterwide/conftest.py
@@ -2,6 +2,7 @@
 
 import pytest
 
+from testsuite.gateway.envoy.route import EnvoyVirtualRoute
 from testsuite.policy.authorization.auth_config import AuthConfig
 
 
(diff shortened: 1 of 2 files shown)


azgabur (dismissed):

testsuite/tests/kuadrant/authorino/operator/clusterwide/conftest.py
```diff
@@ -40,4 +50,3 @@ def commit(request, commit, authorization2):
     """Commits all important stuff before tests"""
     request.addfinalizer(authorization2.delete)
     authorization2.commit()
-    authorization2.wait_for_ready()
```
pehala, line 43: This should now pass and should be included

averevki: Won't pass on wildcard test, but probably can be overwritten to wait for hostsNotLinked

testsuite/tests/kuadrant/authorino/operator/clusterwide/test_wildcard_collision.py
```diff
@@ -36,6 +36,30 @@ def authorization2(authorino, blame, route, openshift2, module_label, gateway):
...
+def commit(request, authorization, authorization2):
+    """Commits both AuthConfigs. Second AuthConfig should fail to reconcile"""
+    request.addfinalizer(authorization.delete)
+    authorization.commit()
+    authorization.wait_for_ready()
+
+    request.addfinalizer(authorization2.delete)
+    authorization2.commit()
+
+    def hosts_not_linked(auth_obj):
```
pehala, line 49: this check should be part of the test

azgabur (approved):

pehala (approved):

PR #360, opened 2024-03-06
Fix metrics tests

averevki:
Changes:
- Fix assert on deep-metrics tests
- Increase wait for metrics scrape
- Make a fixture that checks metrics availability on the cluster to run before other fixtures
- Add explicit id's on deep-metrics tests

### DIFF
--- a/testsuite/openshift/metrics/prometheus.py
+++ b/testsuite/openshift/metrics/prometheus.py
@@ -3,9 +3,10 @@
 from typing import Callable
 from datetime import datetime
 
-import httpx
 import backoff
 from apyproxy import ApyProxy
+
+from testsuite.httpx import KuadrantClient
 
 
 def _params(key: str = "", labels: dict[str, str] = None) -> dict[str, str]:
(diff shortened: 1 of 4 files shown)


testsuite/tests/kuadrant/authorino/metrics/test_deep_metrics.py
```diff
@@ -56,5 +56,5 @@ def test_deep_metrics(metric_name, metric_type, deep_metrics):
         lambda x: x["metric"]["evaluator_name"] == metric_name and x["metric"]["evaluator_type"] == metric_type
     )
 
-    assert metrics
+    assert len(metrics.metrics) > 0
```
pehala, line 59: in which scenario can it be more than one?

averevki: If one of the AuthConfig sections was hit two times

averevki:
^rebase

martinhesko:
Tested this locally, ServiceMonitor gets deleted as expected, all the changes seem reasonable, LGTM on my side

martinhesko (approved):

testsuite/openshift/metrics/prometheus.py
```diff
@@ -73,7 +74,7 @@ def wait_for_scrape(self, target_service: str, metrics_path: str = "/metrics"):
         """Wait before next metrics scrape on service is finished"""
         call_time = datetime.utcnow()
 
-        @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=10)
+        @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=20)
```
pehala, line 77: Do we really need this? and if so, do we know why it needs to be that big, what is causing it to be that horribly slow?

pehala (approved):

Issue #111, opened 2022-09-23
Add plain identity test


pehala:
**Summary**
Implement tests for Identity: Plain

**Documentation**
https://github.com/Kuadrant/authorino/blob/main/docs/features.md#plain-identityplain

**Test Cases**

- [x] Use of Authorino as Kubernetes ValidatingWebhook service (Kubernetes 'userInfo' injected in the body of the AdmissionReview request
- [ ] Concider test case without Webhook Validation.

Won't be done for now: Authentication handled in Envoy leveraging the Envoy JWT Authentication filter (decoded JWT injected as 'metadata_context')

mkudlej:
Kubernetes part of this is covered in https://github.com/Kuadrant/testsuite/issues/89

jsmolar:
Use case described in features doc: "Use of Authorino as Kubernetes ValidatingWebhook service (Kubernetes 'userInfo' injected in the body of the AdmissionReview request): is covered by https://github.com/Kuadrant/testsuite/pull/141

jsmolar:
The second use case `Authentication handled in Envoy leveraging the Envoy JWT Authentication filter (decoded JWT injected as 'metadata_context')` requires Envoy CR modification which is not a valid test case for Kuadrant.

PR #363, opened 2024-03-07
Add plain identity test

trepel:
## Test case
Authorino Plain Identity test

## Changes
- Add simple test for Plain Identity

closes https://github.com/Kuadrant/testsuite/issues/111

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ testsuite/%: FORCE poetry-no-dev
 
 test: ## Run all non mgc tests
 test pytest tests: poetry-no-dev
-	$(PYTEST) -n4 -m 'not mgc' --dist loadfile $(flags) testsuite
+	$(PYTEST) -n4 -m 'not mgc' --dist loadfile --enforce $(flags) testsuite
 
 authorino: ## Run only authorino related tests
 authorino: poetry-no-dev
--- a/testsuite/capabilities.py
+++ b/testsuite/capabilities.py
@@ -10,17 +10,24 @@ from testsuite.config import settings
 @functools.cache
 def has_kuadrant():
     """Returns True, if Kuadrant deployment is present and should be used"""
-    spokes = weakget(settings)["control_plane"]["spokes"] % {}
+    if spokes := weakget(settings)["control_plane"]["spokes"] % {}:
+        for name, openshift in spokes.items():
+            # Try if Kuadrant is deployed
+            if not openshift.connected:
+                return False, f"Spoke {name} is not connected"
+            project = settings["service_protection"]["system_project"]
+            kuadrant_openshift = openshift.change_project(project)
+            kuadrants = kuadrant_openshift.do_action("get", "kuadrant", "-o", "json", parse_output=True)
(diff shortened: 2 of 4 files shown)


testsuite/tests/kuadrant/authorino/identity/plain/test_plain_identity.py
```diff
@@ -0,0 +1,23 @@
...
+        - Send request
+        - Assert that response status code is 401 (Unauthorized)
+        - Replace Plain Identity selector value with existent path in auth_json
+        - Assert that response status code is 200 (OK)
+    """
+    authorization.identity.add_plain("nonexistent-path", "auth.identity")
+    response = client.get("/get")
+    assert response.status_code == 401
+
+    authorization.identity.add_plain("existent-path", "context.request.http.method")
```
azgabur, line 21: This will not replace the existing plain identity but will put it alongside the previous "nonexistent-path". This will cause the `auth.identity` to contain the method verb `GET` and both the previous and this will pass with result authenticated.

There is no clean way to replace previously created identities other than using `authorization.identity.clear_all()` which will remove all.

azgabur (commented):
The idea with existing and non-existing path is a good test case for this simple test. Maybe to create this scenario you can use fixture parametrization. You can take inspiration of this functionality for example in this test [testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py](https://github.com/Kuadrant/testsuite/blob/3eb83fee1277eb37b88e9f9281eb28fe98f24d29/testsuite/tests/kuadrant/authorino/identity/rhsso/test_auth_credentials.py)

testsuite/tests/kuadrant/authorino/identity/plain/test_plain_identity.py
```diff
@@ -0,0 +1,23 @@
...
+def test_plain_identity(client, authorization):
+    """
+    Test:
+        - Add Plain Identity using nonexistent path in auth_json as selector
+        - Send request
+        - Assert that response status code is 401 (Unauthorized)
+        - Replace Plain Identity selector value with existent path in auth_json
+        - Assert that response status code is 200 (OK)
+    """
+    authorization.identity.add_plain("nonexistent-path", "auth.identity")
```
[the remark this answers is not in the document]
trepel: I force pushed the changes using authorization fixture as suggested. Thanks!

testsuite/tests/kuadrant/authorino/identity/plain/test_plain_identity.py
trepel: Hey, I should have checked that this only adds stuff and not replace stuff. I refactor the test to use parametrized fixture instead of changing the authconfig in test directly which solves the issue. Thanks for pointing this out.

azgabur (approved):

azgabur (approved):

martinhesko (approved):

trepel:
This has been closed for some reason, I created https://github.com/Kuadrant/testsuite/pull/366 instead

PR #366, opened 2024-03-13
Add plain identity test

trepel:
## Test case
Authorino Plain Identity test

## Changes
- Add simple test for Plain Identity

## Notes
closes https://github.com/Kuadrant/testsuite/issues/111

### DIFF
--- a/testsuite/utils.py
+++ b/testsuite/utils.py
@@ -124,6 +124,11 @@ def extract_response(response, header="Simple", key="data"):
     :param response: http response
     :return: Extracted value
     """
+
+    # Returning None for non-200 responses because the content of such responses might not be a valid JSON
+    if response.status_code != 200:
+        return weakget(None)
+
     return weakget(json.loads(response.json()["headers"][header]))[key]
 
 
--- a/testsuite/tests/kuadrant/authorino/identity/plain/test_plain_identity.py
+++ b/testsuite/tests/kuadrant/authorino/identity/plain/test_plain_identity.py
@@ -1,0 +1,63 @@
+"""
+Test for plain identity.
+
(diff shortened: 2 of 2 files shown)


testsuite/utils.py
```diff
@@ -124,6 +124,11 @@ def extract_response(response, header="Simple", key="data"):
     :param response: http response
     :return: Extracted value
     """
+
+    # Returning None for non-200 responses because the content of such responses might not be a valid JSON
+    if response.status_code != 200:
```
pehala, line 129: I would probably rather have a more generic fix if we ever decide to change how the response is done, WDYT?

```
data = weakget(response.json())["headers"][header]
return weakget(json.loads(data % "{}")[key]
```

azgabur: The issue with this is, that `response.json()` throws an error when the body of the response is an empty string

pehala (approved):

azgabur (approved):

testsuite/utils.py
trepel: @pehala  @azgabur  If you still have a mental power to discuss this, here are two alternatives I considered:
```
    try:
        response.json()
    except JSONDecodeError:
        return weakget(None)
```
```
    if len(response.content) == 0:
        return weakget(None)
```
The former is not optimal imo because there is a potential to "swallow" a valid error - if there is an invalid json there for whatever reason - I can't think of any such reason but still.

The latter is actually quite ok as well. If there is nothing there just return None and try to do the parsing otherwise.  In case of invalid json this would throw an error but I think that it is a good thing.

If you strongly prefer one of these alternatives just let me know and I can create another PR

PR #370, opened 2024-04-03
Add hyperfoil default url

azgabur:
Depends on: https://github.com/3scale-qe/tools/issues/35

The solution of this PR does not load the url and therefore skips the performance tests. I do not know yet how to fix it yet.


pehala:
Closing as Hyperfoil is currently deprecated

Issue #314, opened 2023-12-22
Add ttl test for authentication jwt verification


azgabur:
**Summary**

Authorino by default does not refresh JWKS from OIDC endpoint but can be set to do so in regular intervals by using a `ttl` variable.

**Documentation**

https://github.com/Kuadrant/authorino/blob/main/docs/features.md#jwt-verification-authenticationjwt

**Test cases**

- [x] Implement `ttl` functionality. Done in #313 
- [ ] Test for refreshing JWKS
- [ ] Add a testcase in test_dinosaur

trepel:
Intentionally not added to dinosaur because this would introduce code duplication.

PR #372, opened 2024-04-09
Add TTL test for JWT verification

trepel:
## Test Case

Adding TTL test for JWT validation in Authorino.

Closes https://github.com/Kuadrant/testsuite/issues/314

## Verification Steps

Eye review. Execute the test and verify that it passes.


azgabur (dismissed):

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -282,3 +288,57 @@ def user_with_write_role(admin_rhsso, blame):
     user = admin_rhsso.realm.create_user(blame("someuser"), blame("password"))
     user.assign_realm_role(admin_rhsso.realm.create_realm_role("admin-write"))
     return HttpxOidcClientAuth.from_user(admin_rhsso.get_token, user=user)
+
+
+@pytest.fixture(scope="module")
+def delete_jwks_key(rhsso):
```
pehala, line 294: I think these don't need to be fixtures, they can be standard functions

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -282,3 +288,57 @@ def user_with_write_role(admin_rhsso, blame):
...
+        rhsso.realm.admin.delete_component(provider_id)
+
+    return _delete_jwks_key
+
+
+@pytest.fixture(scope="module")
+def create_jwks_key(rhsso):
+    """Creates a new signing RS256 key in JWKS"""
+
+    def _create_jwks_key():
```
ficap, line 317: how about turning `_create_jwks_key` and `_delete_jwks_key` into methods of RHSSO class instead? they seem pretty general even though I don't think it is used anywhere else in the testsuite. WDYT?

trepel: Thanks! That sounds like a good idea at the first glance. I will look into it and try to refactor as suggested.

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
trepel: I did not get what you meant by this but after Phil's comment below I started to think that you meant exactly what he suggested. I force pushed the turn of these functions to RHSSO methods.

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
trepel: Refactored and force pushed. Please re-review if you can spare some time. Thanks!

trepel:
@pehala @ficap @azgabur There is one thing I don't like about this PR. There was an requirement to have this test both standalone and as part of dinosaur tests. The test was just copy/pasted, I failed to find a nicer way of doing that. If you can think of something I am all ears.

pehala:
In that case, I would either try to find a way how to differentiate those two (i.e make one of them a slightly different test case) or just simply removed it from Dinosaur as the standalone test is better

trepel:
> In that case, I would either try to find a way how to differentiate those two (i.e make one of them a slightly different test case) or just simply removed it from Dinosaur as the standalone test is better

ok guys, updated and force pushed

testsuite/tests/kuadrant/authorino/identity/rhsso/test_jwt_ttl.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # 200 OK expected since Authorino should have the deleted JWKS key still cached.
+    # Potentially unstable if Authorino triggers OIDC discovery in between the JWKS key deletion just above
+    # and the client.get call just below.
+    response = client.get("/anything/dinosaurs_mgmt/v1/dinosaurs", auth=auth)
+    assert response.status_code == 200
+
+    # Sleeping for jwt_ttl seconds (max 60s though) to ensure Authorino triggers the OIDC discovery during the sleep.
+    # OIDC discovery detects that the signing RS256 JWKS key has been removed and updates the Authorino cache.
+    sleep(jwt_ttl if jwt_ttl <= 60 else 30)
```
pehala, line 44: in what scenario is jwt_ttl higher than 60sec?

testsuite/tests/kuadrant/authorino/identity/rhsso/test_jwt_ttl.py
```diff
@@ -0,0 +1,66 @@
...
+        - assert that response status code is still 200 due to deleted key being cached by Authorino
+        - sleep to allow for Authorino to update the cache as part of OIDC discovery
+        - assert that response status is 401 after cache update
+        - create a new signing RS256 key in JWKS to be used for generating new JWTs
+        - generate a new JWT
+        - assert that response status code is 200 if using new JWT
+        - assert that response status is still 401 for old JWT - ie old JWT does not work with new signing RS256 key
+
+    """
+    response = client.get("/anything/dinosaurs_mgmt/v1/dinosaurs", auth=auth)
```
pehala, line 30: you could simplify the URL to just /get

testsuite/tests/kuadrant/authorino/identity/rhsso/conftest.py
```diff
@@ -10,13 +12,34 @@ def oidc_provider(rhsso):
...
+
+
+@pytest.fixture(scope="module")
+def jwt_ttl():
+    """Returns TTL in seconds for Authorino to trigger OIDC discovery and update its cache"""
+    return 30
+
+
+@pytest.fixture(scope="module")
+def create_jwt(rhsso, auth):
```
pehala, line 38: `create_jwt_auth`

testsuite/tests/kuadrant/authorino/identity/rhsso/test_jwt_ttl.py
trepel: Good catch! I am sorry, copy paste error. Will force push once we agree on how to address your other comments

testsuite/tests/kuadrant/authorino/identity/rhsso/conftest.py
trepel: Make sense, will rename as you suggested

testsuite/tests/kuadrant/authorino/identity/rhsso/test_jwt_ttl.py
trepel: Good question. The answer is that there is no such scenario, it is just a safety measure in case somebody changes the value for some long interval for whatever reason (e.g. testing purposes). If you don't like it it is perfectly ok to remove it but in that case I would add a comment to `jwt_ttl` fixture saying that this should not be set for too many seconds. Wdyt?

pehala: Why is there a maximum of 60sec?

trepel: This is apparently confusing so I will remove this safety measure and just add a comment to the fixture that if the value is modified to higher number then some test might take long to finish

pehala (approved):

azgabur (approved):

PR #377, opened 2024-04-22
Add support for Kind

pehala:
* Use LoadBalancer service for tools
  * https://github.com/3scale-qe/tools/pull/36
  * RHSSO is still not done in that PR
* Add `not standalone_only` to test target
* Fix Prometheus check for metrics tests
* Add KindExposer
   * Some tests needs to be explicitly skipped (TLS, raw_http) when used with Kind exposed

### DIFF
--- a/testsuite/config/__init__.py
+++ b/testsuite/config/__init__.py
@@ -2,7 +2,7 @@
 
 from dynaconf import Dynaconf, Validator
 
-from testsuite.config.tools import fetch_route, fetch_service, fetch_secret
+from testsuite.config.tools import fetch_route, fetch_service, fetch_secret, fetch_service_ip
 
 
 # pylint: disable=too-few-public-methods
@@ -41,7 +41,7 @@ settings = Dynaconf(
         DefaultValueValidator(
             "tracing.collector_url", default=fetch_service("jaeger-collector", protocol="rpc", port=4317)
         ),
-        DefaultValueValidator("tracing.query_url", default=fetch_route("jaeger-query", force_http=True)),
+        DefaultValueValidator("tracing.query_url", default=fetch_service_ip("jaeger-query", force_http=True, port=80)),
         Validator(
             "default_exposer",
             # If exposer was successfully converted, it will no longer be a string"""
@@ -52,9 +52,9 @@ settings = Dynaconf(
         Validator("control_plane.managedzone", must_exist=True, ne=None),
         Validator("control_plane.clusterissuer", must_exist=True, ne=None),
         Validator("letsencrypt.clusterissuer", must_exist=True, ne=None),
-        DefaultValueValidator("keycloak.url", default=fetch_route("no-ssl-sso")),
+        DefaultValueValidator("keycloak.url", default=fetch_service_ip("keycloak", force_http=True, port=8080)),
         DefaultValueValidator("keycloak.password", default=fetch_secret("credential-sso", "ADMIN_PASSWORD")),
-        DefaultValueValidator("mockserver.url", default=fetch_route("mockserver", force_http=True)),
+        DefaultValueValidator("mockserver.url", default=fetch_service_ip("mockserver", force_http=True, port=1080)),
     ],
     validate_only=["authorino", "kuadrant", "default_exposer", "control_plane"],
     loaders=["dynaconf.loaders.env_loader", "testsuite.config.openshift_loader", "testsuite.config.exposer"],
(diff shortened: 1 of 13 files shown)


testsuite/config/tools.py
```diff
@@ -48,6 +52,23 @@ def _fetcher(settings, _):
...
+    def _fetcher(settings, _):
+        try:
+            openshift = settings["tools"]
+            with openshift.context:
+                ip = selector(f"service/{name}").object(cls=Service).external_ip
+                return f"http://{ip}:{port}" if force_http else f"https://{ip}:{port}"
+        # pylint: disable=broad-except
+        except Exception:
+            logger.warning("Unable to fetch route %s from tools", name)
+            return None
```
averevki, line 67: I think there is no reason to have broad exception here. It should basically be able to produce only 2 errors - Openshift one and AttributeError, am I wrong?
```suggestion
        try:
            openshift = settings["tools"]
            with openshift.context:
                ip = selector(f"service/{name}").object(cls=Service).external_ip
                return f"http://{ip}:{port}" if force_http else f"https://{ip}:{port}"
        # pylint: disable=broad-except
        except (OpenshiftPythonError, AttributeError):
            logger.warning("Unable to fetch route %s from tools", name)
            return None
```

pehala: Theoretically yes, but it is there to be safe because raising exception here would mean failing whole testsuite

averevki: It's fine. I think testsuite should raise errors on unexpected code paths

averevki (dismissed):

testsuite/openshift/service.py
azgabur, line 1: Adding LoadBalancer service type requires adding longer timeout for deletion. In case of quick creation and deletion - as in quickly running subset of tests - can create Openshift timeout exceptions. I suggest adding this change: b56bfa5631aa384323d34cd4c0b86e0ce91cd1eb

testsuite/config/__init__.py
azgabur, line 1: Forgotten route. Please change line 44 like this:
```
DefaultValueValidator("tracing.query_url", default=fetch_service_ip("jaeger-query", port=80, force_http=True)),
```

azgabur (commented):
WIP review

testsuite/openshift/client.py
```diff
@@ -73,7 +73,7 @@ def project(self):
     def connected(self):
         """Returns True, if user is logged in and the project exists"""
         try:
-            self.do_action("status")
+            self.do_action("get", "ns", self._project)
```
azgabur, line 76: kubectl/oc has command for checking privileges. https://kubernetes.io/docs/reference/kubectl/generated/kubectl_auth/kubectl_auth_can-i/
Not sure which is the best but some candidate examples from the docs are:
`auth can-i '*' '*' `
`auth can-i create pods --all-namespaces`

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -56,6 +56,10 @@ def create_instance(cls, openshift: OpenShiftClient, name, hostname, labels, tls
     def service_name(self) -> str:
         return f"{self.name()}-istio"
 
+    def external_ip(self) -> str:
```
azgabur, line 59: Missing `@property` decorator, or?

azgabur: What about https? Wont the port be 443?

azgabur (commented):
I did not test it on kind, but if @averevki tried it, I will approve.

testsuite/gateway/gateway_api/gateway.py
pehala: Property is not there on purpose, I need to pass this function as function, which is difficult to do with property.

Regarding 443, You are right, that is an issue, the port should be changed based on what is in the listener, but given that TLS is always skipped right now on Kind, I will fix it in follow-up PR as it should not affect the tests if thats ok with you

azgabur: It can be fixed in another PR :+1:

testsuite/openshift/client.py
pehala: I think that would be similar hacky solution as I used, point of the function to determine that you are connected and the namespace exists and both of these solution achieve by "accident", their purpose is to get a different information

averevki (approved):

averevki:
> I did not test it on kind, but if @averevki tried it, I will approve.

@azgabur I tested it, all good :+1:

azgabur (approved):

PR #383, opened 2024-04-25
Use Enforced condition for ready check

pehala:
* For AuthPolicy and RateLimitPolicy


jsmolar (approved):

azgabur (approved):
DNSPolocy and after https://github.com/Kuadrant/kuadrant-operator/issues/572 gets resolved, TLSPolicy should also receive such ready check in future PR.

PR #384, opened 2024-04-25
Use xfail_strict

pehala:
* This means that it will report a Failure when `xfail` test passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ authorino: poetry-no-dev
 
 authorino-standalone: ## Run only test capable of running with standalone Authorino
 authorino-standalone: poetry-no-dev
-	$(PYTEST) -n4 -m 'authorino and not kuadrant_only' --dist loadfile --enforce --standalone $(flags) testsuite/tests/kuadrant/authorino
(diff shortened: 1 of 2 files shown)


jsmolar (dismissed):

azgabur:
I do not recommend this change as the only tests in the testsuite using xfail are the dinosaur tests as added in https://github.com/Kuadrant/testsuite/pull/379 and this tests are failing as expected when running the testsuite with `make test` but when running standalone authorino tests with `make authorino-standalone` the dinosaur tests will pass because the max limit is different in AuthConfig and AuthPolicy objects. This would PR would make `make authorino-standalone` always always fail.

pehala:
You are right. The test should never been a pure xfail in the first place. Sadly, I didn't find a way how to easily specify conditional xfail based on cmdline option, so I included a quickhack to make it work for the time being

averevki (approved):
I'm fine with this solution. Either way it will require some sort of workaround if tests are collected based on marks first thing in the code. Maybe also add `run=False` to the xfail mark so those tests are not executed at all.

azgabur (approved):

PR #386, opened 2024-05-06
Configurable issuer

pehala:
* Add configuration options for ClusterIssuer names used in tests
   * One option for selfsigned and one option for letsencrypt one
   
* Depends on #385

### DIFF
--- a/testsuite/tests/kuadrant/gateway/conftest.py
+++ b/testsuite/tests/kuadrant/gateway/conftest.py
@@ -44,12 +44,12 @@ def exposer(request, hub_openshift) -> Exposer:
 
 
 @pytest.fixture(scope="session")
-def cluster_issuer():
+def cluster_issuer(testconfig):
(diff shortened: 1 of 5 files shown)


testsuite/tests/kuadrant/gateway/conftest.py
```diff
@@ -44,12 +44,12 @@ def exposer(request, hub_openshift) -> Exposer:
 
 
 @pytest.fixture(scope="session")
-def cluster_issuer():
+def cluster_issuer(testconfig):
     """Reference to cluster self-signed certificate issuer"""
     return CustomReference(
```
averevki, line 49: don't you also need to validate dynaconf here?

pehala: `control_plane` is validated by default

averevki (approved):

ficap (approved):

PR #387, opened 2024-05-06
Add affected by status tests

pehala:
* Unifies wait_for_ready into a Policy object
* Add test for trying out Affected by status condition

Depends on #385


averevki (dismissed):

testsuite/tests/kuadrant/gateway/reconciliation/test_affected_by.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_gateway_status(gateway, dns_policy, tls_policy):
+    """Tests affected by status for Gateway"""
+    gateway.refresh()
+    assert gateway.is_affected_by(dns_policy)
+    assert gateway.is_affected_by(tls_policy)
+
+    dns_policy.delete()
+    assert not gateway.refresh().is_affected_by(dns_policy)
```
ficap, line 28: this test may easily become flaky since the status is eventually consistent. I suggest adding backoff or better something like wait_for_condition() to have it hardened by default.

testsuite/utils.py
```diff
@@ -170,17 +170,24 @@ def _asdict_recurse(obj):
     return result
 
 
+def check_condition(condition, condition_type, status, reason, message):
+    """Checks if condition matches expectation"""
```
ficap, line 174: maybe document that when message or reason are None they are ignored during resolution

testsuite/gateway/gateway_api/route.py
```diff
@@ -43,6 +44,21 @@ def create_instance(
...
         return cls(model, context=openshift.context)
 
+    def is_affected_by(self, policy: Policy):
+        """Returns True, if affected by status is found within the object for the specific policy"""
+        for condition_set in self.model.status.parents:
+            if condition_set.controllerName == "kuadrant.io/policy-controller":
+                for condition in condition_set.conditions:
+                    if check_condition(
+                        condition,
+                        f"kuadrant.io/{policy.kind(lowercase=False)}Affected",
```
ficap, line 54: is it intended that there is no space between object reference and word Affected?

pehala: It is, the final name looks like this `kuadrant.io/RateLimitPolicyAffected`

ficap (approved):

ficap (approved):

PR #391, opened 2024-05-09
Add make clean target

azgabur:
This make target will clean objects from Openshift cluster created by current user or additionally any other user when specified like so: 
```
USER=<name> make clean
```
Testsuite by default cleans up these objects but if test ends in error it is common some objects will not be removed.
This make target will make it easy to clean such left-overs while developing tests.

Please comment if I forgot some CR categories!

Note:
Yes the construction `$\` followed by new-line is [official](https://www.gnu.org/software/make/manual/html_node/Splitting-Lines.html) way to split string to multiple lines...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,5 +99,27 @@ reportportal: polish-junit
 help: ## Print this help
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+CR_NAMES = "$\
+Gateway,$\
+HTTPRoute,$\
+Deployment,$\
+Service,$\
+Route,$\
+ServiceAccount,$\
+Secret,$\
+AuthConfig,$\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -99,5 +99,25 @@ reportportal: polish-junit
 help: ## Print this help
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+CR_NAMES = "$\
+Gateway,$\
+HTTPRoute,$\
+Deployment,$\
+Service,$\
```
pehala, line 106: I am missing secrets

azgabur:
^ added `Secret` and a test for the `USER` variable

pehala (approved):

averevki (approved):

azgabur:
^ add parallel cleaning

averevki (approved):

pehala (approved):

Issue #84, opened 2022-09-08
Add test for using Keycloak authorization services


pehala:
**Summary**
Implement tests for authorization with Keycloak Authorization Services. The aim of this issue is to create a more end-to-end oriented test case, that would incorporate a more complex OPA policy (with requests to Keycloak).

**Documentation**
https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/keycloak-authorization-services.md
Keycloak Authorization Services: https://www.keycloak.org/docs/latest/authorization_services/index.html

PR #393, opened 2024-05-13
Add Keycloak Authorization Services test

trepel:
## Test Case

Adding test for Complex OPA policy using Keycloan Authorization Services under the hood.

Closes https://github.com/Kuadrant/testsuite/issues/84

## Verification Steps

Eye review. Execute the test and verify that it passes.


azgabur (approved):

testsuite/tests/kuadrant/authorino/authorization/opa/test_authorization_services.py
```diff
@@ -0,0 +1,184 @@
...
+        protected_resource["_id"],
+        {
+            "name": "Allow GET for requester",
+            "description": "Allow GET for requester",
+            "scopes": ["get"],
+            "users": [f"{requester_auth.username}"],
+        },
+    )
+
+    # Owner is allowed to access the protected resource now too - unclear if this is a bug or feature in Keycloak
```
pehala, line 160: not sure why this should be a bug?

testsuite/tests/kuadrant/authorino/authorization/opa/test_authorization_services.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # Allow HTTP GET access to the protected resource for "requester" user
+    # Only resource owner is allowed to do this
+    owner_uma.policy_resource_create(
+        protected_resource["_id"],
+        {
+            "name": "Allow GET for requester",
+            "description": "Allow GET for requester",
+            "scopes": ["get"],
+            "users": [f"{requester_auth.username}"],
```
pehala, line 156: You can use `[requester_auth.username]`

testsuite/tests/kuadrant/authorino/authorization/opa/test_authorization_services.py
trepel: I did not go too deep into this because from Authorino testing perspective it is not relevant. If you create a protected resource (you being an owner of it) without any permission associated to it, nobody is allowed to access it - not even you. However, you are allowed to create permissions for that resource. It is weird that if you create a permission the way the test does (via `policy_resource_create`) for `requester` user only (`requester` not being you) it also allows you to access the resource. It might be that I am using that `policy_resource_create` incorrectly, when creating permissions via GUI different API endpoints are user (2 calls are required actually) that the API endpoint `policy_resource_create` uses under the hood. Unfortunately I was not able to use the API endpoints GUI uses programmatically (via curl).

testsuite/tests/kuadrant/authorino/authorization/opa/test_authorization_services.py
trepel: nice, will update as suggested

pehala (approved):

testsuite/tests/kuadrant/authorino/authorization/opa/test_authorization_services.py
trepel: updated and force-pushed as suggested, could you please re-review (this was the only change in force-push)

azgabur (approved):

PR #394, opened 2024-05-13
Reintroduce sleep into RLP wait

pehala:
It seems like it still needs a bit of sleep


jsmolar (approved):

azgabur:
I was running `make test` recently and I did not notice limitador tests failing

pehala:
They are just flaky right now, they can pass, but if you look at, for example, nightlies they fail. This should improve stability

PR #395, opened 2024-05-17
Ignore useless-return

pehala:
Not sure if there is better way to handle this specific case, and I am not sure if this is a bug with pylint.

Introduced in pylint 3.2.0


jsmolar:
> Not sure if there is better way to handle this specific case, and I am not sure if this is a bug with pylint.
> 
> Introduced in pylint 3.2.0

`pass`?

pehala:
> `pass`?

Yup, that is much better. Thanks for the tip!

jsmolar (approved):

azgabur (approved):

Issue #396, opened 2024-05-17
Add support for custom attributes for Keycloak users


pehala:
**Summary**

Keycloak 24 introduced user profile management, which means that we cannot easily add new attributes like MiddleName, because support is not there yet in [python-keycloak](https://github.com/marcospereirampj/python-keycloak)

PR #397, opened 2024-05-17
Fix failing dinosaur test on Keycloak 24

pehala:
* You cannot add custom attributes like we did before, you need to register them first. This is temporary fix that removes one test-case until #396 is done
* Fix xfail

### DIFF
--- a/testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
@@ -3,6 +3,7 @@ Conftest for dinosaur test
 """
 
 import pytest
+from openshift_client import OpenShiftPythonException
 
 from testsuite.httpx.auth import HttpxOidcClientAuth
 from testsuite.oidc.keycloak import Keycloak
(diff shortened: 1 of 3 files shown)


testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -28,6 +29,18 @@ def admin_rhsso(blame, keycloak):
...
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization):
+    """Commits all important stuff before tests"""
+    request.addfinalizer(authorization.delete)
+    try:
+        authorization.commit()
+        authorization.wait_for_ready()
+    except OpenShiftPythonException as exc:
+        if "Too many" in exc.result.err():
+            pytest.xfail("AuthPolicy max limit")
```
azgabur, line 41: This xfail usage is basically a skip when used as a function. Therefore when the product get fixed we need to watch for the lack of xfails for the dinosaur tests in testrun log to notice the issue was fixed.

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
```diff
@@ -28,6 +29,18 @@ def admin_rhsso(blame, keycloak):
     return info
 
 
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization):
+    """Commits all important stuff before tests"""
```
azgabur, line 34: Mention in the comment this whole commit fixture is a temporary solution so it can be removed later. Should we track this with an github issue? If yes, maybe put a link in the comment as well.

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
pehala: when authorino gets fixed, this wont trigger and will pass

testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
pehala: It does not need to be temporary, after the fix is done this will still work.

jsmolar (dismissed):

jsmolar (approved):

PR #400, opened 2024-05-20
Add blame to metadata tests

pehala:


jsmolar (dismissed):

azgabur (dismissed):

testsuite/tests/kuadrant/authorino/metadata/test_uma.py
```diff
@@ -30,34 +30,34 @@
...
-    return create_client_secret("uma-client-secret", keycloak.client.auth_id, keycloak.client.secret)
+    return create_client_secret(blame("uma-client-secret"), keycloak.client.auth_id, keycloak.client.secret)
 
 
 @pytest.fixture(scope="module")
-def auth2(keycloak):
-    """Creates new Keycloak User and returns his authentication object for HTTPX"""
-    user = keycloak.realm.create_user("newTestUser", "p")
+def auth2(keycloak, blame):
+    """Creates new RHSSO User and returns his authentication object for HTTPX"""
```
jsmolar, line 40: `RHSSO` -> `Keycloak`

trepel:
> AFAIK `rhsso.client.create_uma_resource("get1", ["/anything"])` and the one below should be also blamed

Should not https://github.com/Kuadrant/testsuite/pull/400/files#diff-afa1a2b71e5e62d18ea25af8e2fa2d50c16bad28fad39df4e03f3d9a573be8e7L127 be blamed as well then?

I wanted to review this PR but I realized that I don't really know what should be blamed and what not. If somebody feels like explaining that to me that would be great.

pehala:
It most likely should, tests share one rhsso realm for the entirety of the test run, so every rhsso resourxe should ideally be blamed to prevent collisions. I changed only a few instances of this bevausw it came up as a actual collisin but there are probably more instances of this scattered among the tests

trepel:
Thanks! So blaming is primarily used for Keycloak resources to avoid potential collisions. With this knowledge I re-reviewed and I think that after blaming the protected resource name as I suggested this is good to go. To be super fair the test creates one more Keycloak resource - Scope called `get` - but the name is derived from HTTP method and is expected to be that way, blaming that would be a bit more involved so I would leave it out of scope of this PR.

jsmolar (approved):

PR #404, opened 2024-05-20
Refactor clusterissuer to issuer

azgabur:
The Letsecrypt issuer requires a Secret which provides secret for DNS01 verification. If ClusterIssuer object is used, the Secret needs to be stored in the namespace where the CertManager operator is installed. I propose using Issuer object inside the testing namespace which already contains secrets for DNSPolicy testing (the `mgc-aws-credentials` and `mgc-gcp-credentials`) to keep all the secrets in the same namespace. Another reason is to prevent accidental deletion of the secret in case of nightly testing.

This PR also adds entries to settings template file.

### DIFF
--- a/testsuite/config/__init__.py
+++ b/testsuite/config/__init__.py
@@ -50,8 +50,14 @@ settings = Dynaconf(
             messages={"condition": "{value} is not valid exposer"},
         ),
         Validator("control_plane.managedzone", must_exist=True, ne=None),
-        Validator("control_plane.clusterissuer", must_exist=True, ne=None),
-        Validator("letsencrypt.clusterissuer", must_exist=True, ne=None),
+        (
+            Validator("control_plane.issuer.name", must_exist=True, ne=None)
+            & Validator("control_plane.issuer.kind", must_exist=True, is_in={"Issuer", "ClusterIssuer"})
+        ),
+        (
+            Validator("letsencrypt.issuer.name", must_exist=True, ne=None)
+            & Validator("letsencrypt.issuer.kind", must_exist=True, is_in={"Issuer", "ClusterIssuer"})
+        ),
         DefaultValueValidator("keycloak.url", default=fetch_service_ip("keycloak", force_http=True, port=8080)),
         DefaultValueValidator("keycloak.password", default=fetch_secret("credential-sso", "ADMIN_PASSWORD")),
         DefaultValueValidator("mockserver.url", default=fetch_service_ip("mockserver", force_http=True, port=1080)),
--- a/testsuite/tests/kuadrant/gateway/test_external_ca.py
+++ b/testsuite/tests/kuadrant/gateway/test_external_ca.py
@@ -1,17 +1,17 @@
 """
 This module contains the most basic happy path test for both DNSPolicy and TLSPolicy
-for a cluster with Let's Encrypt ClusterIssuer
+for a cluster with Let's Encrypt Issuer
 
 Prerequisites:
 * multi-cluster-gateways ns is created and set as openshift["project"]
 * managedclustersetbinding is created in openshift["project"]
 * gateway class "kuadrant-multi-cluster-gateway-instance-per-cluster" is created
 * cert-manager Operator installed
-* Let's Encrypt ClusterIssuer object configured on the cluster matching the template:
+* Let's Encrypt Issuer object configured on the cluster matching the template:
 apiVersion: cert-manager.io/v1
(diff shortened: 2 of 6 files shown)


testsuite/config/__init__.py
```diff
@@ -50,8 +50,16 @@ def __init__(self, name, default, **kwargs) -> None:
...
         Validator("control_plane.managedzone", must_exist=True, ne=None),
-        Validator("control_plane.clusterissuer", must_exist=True, ne=None),
-        Validator("letsencrypt.clusterissuer", must_exist=True, ne=None),
+        Validator("control_plane.issuer", must_exist=True, ne=None)
+        | (
+            Validator("control_plane.issuer.name", must_exist=True, ne=None)
+            & Validator("control_plane.issuer.kind", must_exist=True, ne=None)
+        ),
+        Validator("letsencrypt.issuer", must_exist=True, ne=None)
+        | (
```
pehala, line 59: Why or?

testsuite/config/__init__.py
```diff
@@ -50,8 +50,16 @@ def __init__(self, name, default, **kwargs) -> None:
...
+            Validator("letsencrypt.issuer.name", must_exist=True, ne=None)
+            & Validator("letsencrypt.issuer.kind", must_exist=True, ne=None)
```
pehala, line 61: You should allow only exact strings as well

testsuite/config/__init__.py
azgabur: My mistake, I changed it now.

testsuite/config/__init__.py
azgabur: I added check for this

testsuite/config/__init__.py
```diff
@@ -50,8 +50,14 @@ def __init__(self, name, default, **kwargs) -> None:
...
         Validator("control_plane.managedzone", must_exist=True, ne=None),
-        Validator("control_plane.clusterissuer", must_exist=True, ne=None),
-        Validator("letsencrypt.clusterissuer", must_exist=True, ne=None),
+        (
+            Validator("control_plane.issuer.name", must_exist=True, ne=None)
+            & Validator("control_plane.issuer.kind", must_exist=True, ne=None, is_in={"Issuer", "ClusterIssuer"})
+        ),
+        (
+            Validator("letsencrypt.issuer.name", must_exist=True, ne=None)
+            & Validator("letsencrypt.issuer.kind", must_exist=True, ne=None, is_in={"Issuer", "ClusterIssuer"})
```
pehala, line 59: `ne` is probably not needed, since you use `is_in`

azgabur: Addressed

testsuite/tests/kuadrant/gateway/test_external_ca.py
```diff
@@ -1,15 +1,15 @@
...
 * managedclustersetbinding is created in openshift["project"]
 * gateway class "kuadrant-multi-cluster-gateway-instance-per-cluster" is created
 * cert-manager Operator installed
-* Let's Encrypt ClusterIssuer object configured on the cluster matching the template:
+* Let's Encrypt Issuer object configured on the cluster matching the template:
 apiVersion: cert-manager.io/v1
-kind: ClusterIssuer
+kind: ClusterIssuer | Issuer
 metadata:
   name: letsencrypt-staging
```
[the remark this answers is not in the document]
azgabur: Addressed

pehala (approved):

pehala:
> If we changing that, can we just move to only Issuers instead and create them during the testsuite run with session scope?

We can theoretically do that for self-signed issuer but I think it is not needed as that issuer can be easily shared. For letsencrypt I would try to avoid it because it contains AWS credentials and I would like to avoid having any credentials in the testsuite

azgabur:
^ Fix merge conflict

averevki:
> > If we changing that, can we just move to only Issuers instead and create them during the testsuite run with session scope?
> 
> We can theoretically do that for self-signed issuer but I think it is not needed as that issuer can be easily shared. For letsencrypt I would try to avoid it because it contains AWS credentials and I would like to avoid having any credentials in the testsuite

It'll be only 1 configurable test for letsencrypt that can be skipped without configuration, same as we have `auth0` rn. That would've made testsuite more self-contained. But this PR is also ok I guess

azgabur:
^ Fix merge conflic

pehala (approved):

PR #410, opened 2024-05-27
Add Kuadrant tracing tests

jsmolar:
The first test for Kuadrant level tracing. When testing the presence of authorino and limitador service names in the trace. Do we also want to verify the trace from Istio?

### DIFF
--- a/testsuite/openshift/limitador.py
+++ b/testsuite/openshift/limitador.py
@@ -1,0 +1,42 @@
+"""Limitador CR object"""
+
+from openshift_client import selector
+
+from testsuite.openshift import OpenShiftObject, modify
+from testsuite.openshift.authorino import TracingOptions
+from testsuite.openshift.deployment import Deployment
+from testsuite.utils import asdict
+
+
+class LimitadorCR(OpenShiftObject):
+    """Represents Limitador CR objects"""
+
+    @property
+    def deployment(self) -> Deployment:
+        """Returns Deployment object for this Limitador"""
+        with self.context:
+            return selector("deployment/limitador-limitador").object(cls=Deployment)
+
+    @property
+    def tracing(self) -> dict:
+        """Returns tracing config"""
+        return self.model.spec.setdefault("tracing", {})
+
+    @tracing.setter
+    @modify
+    def tracing(self, config: TracingOptions):
+        """Sets tracing"""
+        self.model.spec.setdefault("tracing", {})
+        self.model.spec["tracing"] = asdict(config)
+
+    @property
+    def verbosity(self) -> int:
+        """Returns verbosity config"""
+        return self.model.spec.setdefault("verbosity", {})
+
+    @verbosity.setter
+    @modify
+    def verbosity(self, level: int):
+        """Sets verbosity"""
+        self.model.spec.setdefault("verbosity", {})
+        self.model.spec["verbosity"] = level
--- a/testsuite/openshift/limitador.py
+++ b/testsuite/openshift/limitador.py
@@ -2,41 +2,15 @@
 
 from openshift_client import selector
 
-from testsuite.openshift import OpenShiftObject, modify
-from testsuite.openshift.authorino import TracingOptions
+from testsuite.openshift import CustomResource
 from testsuite.openshift.deployment import Deployment
-from testsuite.utils import asdict
 
 
-class LimitadorCR(OpenShiftObject):
+class LimitadorCR(CustomResource):
     """Represents Limitador CR objects"""
 
     @property
     def deployment(self) -> Deployment:
         """Returns Deployment object for this Limitador"""
         with self.context:
-            return selector("deployment/limitador-limitador").object(cls=Deployment)
-
-    @property
-    def tracing(self) -> dict:
-        """Returns tracing config"""
-        return self.model.spec.setdefault("tracing", {})
-
-    @tracing.setter
-    @modify
-    def tracing(self, config: TracingOptions):
-        """Sets tracing"""
-        self.model.spec.setdefault("tracing", {})
-        self.model.spec["tracing"] = asdict(config)
-
-    @property
-    def verbosity(self) -> int:
-        """Returns verbosity config"""
-        return self.model.spec.setdefault("verbosity", {})
-
-    @verbosity.setter
-    @modify
-    def verbosity(self, level: int):
-        """Sets verbosity"""
-        self.model.spec.setdefault("verbosity", {})
-        self.model.spec["verbosity"] = level
+            return selector(f"deployment/{self.name()}").object(cls=Deployment)
--- a/testsuite/tests/kuadrant/conftest.py
+++ b/testsuite/tests/kuadrant/conftest.py
@@ -6,6 +6,7 @@ from openshift_client import selector, OpenShiftPythonException
 
 from testsuite.config import settings
 from testsuite.openshift.authorino import AuthorinoCR
+from testsuite.openshift.limitador import LimitadorCR
 from testsuite.policy.authorization.auth_policy import AuthPolicy
 from testsuite.policy.rate_limit_policy import RateLimitPolicy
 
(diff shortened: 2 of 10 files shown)


testsuite/tracing.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, jaeger_backend: bool, collector_url: str, query_url: str, cli
         self.client = client or KuadrantClient(verify=False)
         self.query = ApyProxy(query_url, session=self.client)
 
-    @backoff.on_predicate(backoff.fibo, lambda x: x == [], max_tries=7, jitter=None)
+    @backoff.on_predicate(backoff.fibo, lambda x: x == [], max_tries=5, jitter=None)
```
pehala, line 18: why decrease it?

testsuite/tests/kuadrant/observability/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+@pytest.fixture(scope="session")
+def limitador(kuadrant_system_client) -> LimitadorCR:
+    """Returns Limitador CR"""
+    try:
+        with kuadrant_system_client.context:
+            limitador = selector("limitador").object(cls=LimitadorCR)
+            limitador.committed = True
+    except OpenShiftPythonException:
+        pytest.fail("Running Kuadrant tests, but Authorino resource was not found")
```
pehala, line 34: Limitador resource was not found

testsuite/tests/kuadrant/observability/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return limitador
+
+
+@pytest.fixture(scope="module", autouse=True)
+def enable_tracing(authorino, limitador, tracing):
+    """Enable tracing for Authorino and Limitador"""
+    authorino.tracing = TracingOptions(tracing.collector_url, insecure=True)
+    limitador.tracing = TracingOptions(tracing.collector_url)
```
pehala, line 43: why are we not using insecure for limitador as well?

testsuite/tests/kuadrant/observability/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+    limitador.verbosity = 3
+    limitador.deployment.wait_for_ready()
```
pehala, line 45: We should wait for limitador CR readiness not the deployment

testsuite/tests/kuadrant/observability/test_tracing.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Tests that traces are collected from Authorino and Limitador
+
+    Trace IDs do not propagate to wasm modules in Istio/Envoy, affecting trace continuity in Limitador:
+    https://github.com/envoyproxy/envoy/issues/22028
+    That is the reason why we set parent trace ID in request.
+    """
+    trace_id = os.urandom(16).hex()
+    client.get("/get", auth=auth, headers={"Traceparent": f"00-{trace_id}-{os.urandom(8).hex()}-01"})
+    sleep(5)  # Waits for tracing backend to collect all traces
```
pehala, line 17: Why 5 and is there not a better waiting mechanism? Maybe retry?

testsuite/tests/kuadrant/observability/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+from openshift_client import selector
+
+from testsuite.httpx.auth import HttpxOidcClientAuth
+from testsuite.openshift.authorino import TracingOptions
+from testsuite.openshift.limitador import LimitadorCR
+from testsuite.policy.rate_limit_policy import Limit
+
+
+@pytest.fixture(scope="module")
+def auth(oidc_provider):
```
pehala, line 14: This fixture can be probably mvoed to Kuadrant conftest, but we can discuss it

testsuite/tests/kuadrant/conftest.py
```diff
@@ -24,15 +40,15 @@ def authorization_name(blame):
...
-def authorization(kuadrant, oidc_provider, route, authorization_name, openshift, label):
+def authorization(kuadrant, route, authorization_name, openshift, label):
     """Authorization object (In case of Kuadrant AuthPolicy)"""
     if kuadrant:
         return AuthPolicy.create_instance(openshift, authorization_name, route, labels={"testRun": label})
     return None
 
 
 @pytest.fixture(scope="module")
-def rate_limit(kuadrant, openshift, blame, request, module_label, route, gateway):
```
pehala, line 35: do all other tests pass? Not sure if this is the case for these specifically, but we sometimes do unused imports to make sure everything is ready beforehand

testsuite/openshift/limitador.py
```diff
@@ -0,0 +1,42 @@
...
+        self.model.spec.setdefault("tracing", {})
+        self.model.spec["tracing"] = asdict(config)
+
+    @property
+    def verbosity(self) -> int:
+        """Returns verbosity config"""
+        return self.model.spec.setdefault("verbosity", {})
+
+    @verbosity.setter
+    @modify
```
pehala, line 38: Are we sure we cannot do a better pattern than this? This will get out of hands very quickly

testsuite/tracing.py
jsmolar: `5` should be enough. I've done some testing.

testsuite/tests/kuadrant/observability/conftest.py
jsmolar: This will not work: https://github.com/Kuadrant/limitador-operator/issues/140

testsuite/tests/kuadrant/observability/conftest.py
jsmolar: `insecure` is not defiend for limitador. I created an issue: https://github.com/Kuadrant/limitador-operator/issues/141

testsuite/openshift/limitador.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class LimitadorCR(OpenShiftObject):
+    """Represents Limitador CR objects"""
+
+    @property
+    def deployment(self) -> Deployment:
+        """Returns Deployment object for this Limitador"""
+        with self.context:
+            return selector("deployment/limitador-limitador").object(cls=Deployment)
```
pehala, line 18: Name is not static

testsuite/tests/kuadrant/observability/test_tracing.py
jsmolar: It takes some time for the collector to get all the traces. What happens here, is that the tracing backend correctly returns the trace by `tracreId` but it takes some time for this trace to contain all spans.

testsuite/tests/kuadrant/observability/test_tracing.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Tests that traces are collected from Authorino and Limitador
+
+    Trace IDs do not propagate to wasm modules in Istio/Envoy, affecting trace continuity in Limitador:
+    https://github.com/envoyproxy/envoy/issues/22028
+    That is the reason why we set parent trace ID in request.
+    """
+    trace_id = os.urandom(16).hex()
+    client.get("/get", auth=auth, headers={"Traceparent": f"00-{trace_id}-{os.urandom(8).hex()}-01"})
+    sleep(5)  # Wait for tracing backend to collect all traces
```
pehala, line 21: Does all tracing need this sleep? cannot we use retry instead of it?

testsuite/openshift/limitador.py
```diff
@@ -0,0 +1,16 @@
...
+
+from testsuite.openshift import CustomResource
+from testsuite.openshift.deployment import Deployment
+
+
+class LimitadorCR(CustomResource):
+    """Represents Limitador CR objects"""
+
+    @property
+    def deployment(self) -> Deployment:
```
pehala, line 13: You have two way of achieving the same thing based on deployment type:
Standalone: `limitador_cr.deployment`
Kuadrant: `kuadrant_cr.limitador_deployment`

It would be good to unify path to obtain the same thing regardless of deployment

testsuite/tests/kuadrant/conftest.py
```diff
@@ -30,6 +59,7 @@ def authorization(kuadrant, route, authorization_name, openshift, label):
     return None
 
 
+# pylint: disable=unused-argument
```
pehala, line 62: Remove

testsuite/tests/kuadrant/conftest.py
```diff
@@ -2,18 +2,47 @@
...
+            authorino = selector("authorino").object(cls=AuthorinoCR)
+            authorino.committed = True
+    except OpenShiftPythonException:
+        pytest.fail("Running Kuadrant tests, but Authorino resource was not found")
+
+    return authorino
+
+
+@pytest.fixture(scope="session")
+def limitador(system_openshift) -> LimitadorCR:
```
pehala, line 36: Why we dont do this through Kuadrant if those tests are kuadrant specific?

testsuite/openshift/limitador.py
jsmolar: I'm working on this in another commit/PR

PR #412, opened 2024-05-28
Fix authorization services test

pehala:
There was a missing slash when using service IP. Fixed by using issuer_url from well_known instead of construction URL ourselves


averevki (approved):

trepel (approved):
Nice catch, thanks! I worked like a charm so approving this one.
However, it made me think that the best solution would be to use "well known" endpoints for each and every endpoint.

So created another PR for that:
https://github.com/Kuadrant/testsuite/pull/414

I am ok with getting this merged as it, I will rebase my PR.

pehala:
Closing this PR in favor of #414 as it is much better

PR #415, opened 2024-05-29
Add RateLimitPolicy to affected_by test

pehala:
It was removed in #409 but affected_by test needs it


trepel (approved):
Code looks good to me. Test passed as expected with this change.

averevki (approved):

PR #422, opened 2024-05-31
Add wait_for_ready to test_anonymous_identity

pehala:
Should help with test stability


averevki (approved):

trepel (approved):
LGTM, this should make the test stable. However, I think that it would be better to use parametrized fixture for use cases like this. So the test would effectively be executed twice - 1st without that anonymous identity and 2nd with it thus avoiding updating the AuthPolicy in test and waiting for it to be reconciled altogether.

trepel:
@pehala Probably not worth the effort unless the test turns out to be unstable despite the wait. Just to clarify what I had in mind, it was something along the line of https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/kuadrant/authorino/identity/plain/test_plain_identity.py#L21-L36
The expected behavior is also parametrized.

PR #424, opened 2024-06-03
Rework Prometheus

pehala:
- Separate wait for reconciliation
   - Increase timeout to 350 second 
- Make Prometheus a Lifecycle object
   - Move client stuff to commit  
- Add default Interval for MetricsEndpoint
   - To better aim new wait_for_scrape backoff 
- Use ServiceMonitor as base for methods instead of `service_url`
   - I found it more natural as that is the actual metric definition

### DIFF
--- a/testsuite/openshift/metrics/prometheus.py
+++ b/testsuite/openshift/metrics/prometheus.py
@@ -7,6 +7,7 @@ import backoff
 from apyproxy import ApyProxy
 
 from testsuite.httpx import KuadrantClient
+from .service_monitor import ServiceMonitor
 
 
 def _params(key: str = "", labels: dict[str, str] = None) -> dict[str, str]:
@@ -46,16 +47,16 @@ class Prometheus:
 
     def __init__(self, url: str, token: str, namespace: str = None):
         self.token = token
+        self.url = url
         self.headers = {"Authorization": f"Bearer {self.token}"}
         self.namespace = namespace
 
         self._client = KuadrantClient(headers=self.headers, verify=False)
-        self.client = ApyProxy(url, self._client).api.v1
+        self.client = ApyProxy(self.url, self._client).api.v1
 
-    def get_targets(self) -> dict:
+    def get_active_targets(self) -> dict:
         """Get active metric targets information"""
-        params = {"state": "active"}
-        response = self.client.targets.get(params=params)
+        response = self.client.targets.get(params={"state": "active"})
 
         return response.json()["data"]["activeTargets"]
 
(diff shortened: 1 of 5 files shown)


testsuite/openshift/metrics/prometheus.py
```diff
@@ -70,24 +72,38 @@ def get_metrics(self, key: str = "", labels: dict[str, str] = None) -> Metrics:
 
         return Metrics(response.json()["data"]["result"])
 
-    def wait_for_scrape(self, target_service: str, metrics_path: str = "/metrics"):
+    @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=35)
```
averevki, line 75: 35 seems too much. Works with 20 on my machine. Should we start with a lower number and gradually increase it based on the results? 

Or maybe you already tested it and it is the best choice :))

testsuite/openshift/metrics/prometheus.py
```diff
@@ -70,24 +72,38 @@ def get_metrics(self, key: str = "", labels: dict[str, str] = None) -> Metrics:
...
+    def commit(self):
+        self._client = KuadrantClient(headers=self.headers, verify=False)
+        self.client = ApyProxy(self.url, self._client).api.v1
 
-    def close(self):
-        """Close httpx client connection"""
-        self._client.close()
+    def delete(self):
+        if self._client is not None:
+            self._client.close()
```
averevki, line 109: I don't like much the use of commit/delete functions inside the client-like objects. In my vision, they should commit some data to the server and remove data from it. I'd create a separate interface for client-like objects (KuadrantClient also might be under it) and give it open(?) and close methods. Or maybe httpx already has something like this.  (Interface Segregation or whatever ;) )

testsuite/openshift/metrics/prometheus.py
```diff
@@ -41,21 +43,21 @@ def values(self) -> list[float]:
         return [float(m["value"][1]) for m in self.metrics]
 
 
-class Prometheus:
+class Prometheus(LifecycleObject):
     """Interface to the OpenShift Prometheus client"""
 
     def __init__(self, url: str, token: str, namespace: str = None):
         self.token = token
+        self.url = url
```
averevki, line 51:
```suggestion
class Prometheus(LifecycleObject):
    """Interface to the OpenShift Prometheus client"""
    
    _client: KuadrantClient
    client: ApyProxy

    def __init__(self, url: str, token: str, namespace: str = None):
        self.token = token
        self.url = url
```
maybe something like this would work for variable declaration and to avoid the `type: ignore` thing. But you'd probably need to check `hasattr(self, "_client")` before closing the connection instead. I'm fine with both variants

testsuite/openshift/metrics/prometheus.py
pehala: Well, 20 works on my machine as well, but it keeps steadily failing in nightlies. The number is completely made up and aim is to fix the failing tests as we know it is just instability of ServiceMonitors

testsuite/openshift/metrics/prometheus.py
pehala: `_client` and `client` would be class level variables and would be shared between instances

testsuite/openshift/metrics/prometheus.py
pehala: We already have close() so making it a `LifecycleObject` brings it in line with our other stuff. I can revert this change for now and we can discuss later

testsuite/openshift/metrics/prometheus.py
averevki: fair, 35 is fine. We can tweak it later

averevki (approved):

jsmolar (approved):

PR #425, opened 2024-06-04
Add support methods for CR objects

jsmolar:
This PR adds dunder `__getitem__` and `__setitem__` methods for CR objects. Also Implements authorino-like and limitador-like objects for Kuadrant CR to enable similar manipulation with Limitador and sub-component Limitador from Kuadrant CR.
I had to remove `from testsuite.config import settings` from utils because of cyclical import.

### DIFF
--- a/testsuite/openshift/__init__.py
+++ b/testsuite/openshift/__init__.py
@@ -1,5 +1,6 @@
 """OpenShift common objects"""
 
+import dataclasses
 import functools
 from dataclasses import dataclass, field
 from typing import Optional, Literal
@@ -7,6 +8,7 @@ from typing import Optional, Literal
 from openshift_client import APIObject, timeout, OpenShiftPythonException
 
 from testsuite.lifecycle import LifecycleObject
+from testsuite.utils import asdict
 
 
 class OpenShiftObject(APIObject, LifecycleObject):
@@ -42,6 +44,28 @@ class OpenShiftObject(APIObject, LifecycleObject):
                 return success
         except OpenShiftPythonException:
             return False
+
+
+class CustomResource(OpenShiftObject):
+    """Custom APIObjects that implements methods that improves manipulation with CR objects"""
+
+    def safe_apply(self):
+        """
+        Modifies the model of the apiobj and asserts if a change was applied to a resource.
+        Uses modify_and_apply method from OpenshiftObject.
+        """
+        result, status = self.modify_and_apply(lambda _: True, retries=2)
+        assert status, f"Unable to apply changes for APIObject with result: {result}"
+        return result
+
+    def __getitem__(self, name):
+        return self.model.spec[name]
+
+    def __setitem__(self, name, value):
+        if dataclasses.is_dataclass(value):
+            self.model.spec[name] = asdict(value)
+        else:
+            self.model.spec[name] = value
 
 
 def modify(func):
--- a/testsuite/openshift/kuadrant.py
+++ b/testsuite/openshift/kuadrant.py
@@ -1,29 +1,69 @@
 """Kuadrant CR object"""
 
-from openshift_client import selector
+import dataclasses
 
-from testsuite.openshift import OpenShiftObject, modify
+from openshift_client import selector, timeout
+
+from testsuite.openshift import CustomResource
 from testsuite.openshift.deployment import Deployment
+from testsuite.utils import asdict
 
 
-class KuadrantCR(OpenShiftObject):
+class KuadrantSection:
+    """
+    Base class for Kuadrant sub components:
+        Authorino - spec.authorino
+        Limitador - spec.limitador
+    """
+
+    def __init__(self, kuadrant_cr, spec_name):
+        super().__init__()
+        self.kuadrant_cr = kuadrant_cr
+        self.spec_name = spec_name
+
+    def __getitem__(self, name):
+        return self.kuadrant_cr.model.spec[self.spec_name][name]
+
+    def __setitem__(self, name, value):
+        if dataclasses.is_dataclass(value):
+            self.kuadrant_cr.model.spec[self.spec_name][name] = asdict(value)
+        else:
+            self.kuadrant_cr.model.spec[self.spec_name][name] = value
+
+    def __getattr__(self, item):
+        try:
+            return getattr(self.kuadrant_cr, item)
+        except AttributeError as exc:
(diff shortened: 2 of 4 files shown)


testsuite/openshift/__init__.py
```diff
@@ -44,6 +46,25 @@ def wait_until(self, test_function, timelimit=90):
             return False
 
 
+class CRObject(OpenShiftObject):
```
pehala, line 49: How about naming it `CustomResource`?

testsuite/openshift/__init__.py
```diff
@@ -44,6 +46,25 @@ def wait_until(self, test_function, timelimit=90):
...
+    """Custom APIObjects that implements methods that improves manipulation with CR objects"""
+
+    def safe_apply(self):
+        """Reimplementation of modify_and_apply from OpenshiftObject"""
+        result, status = self.modify_and_apply(lambda _: True, retries=2)
+        assert status
```
pehala, line 55: I would add message to the assert to better indicate what went wrong

testsuite/openshift/__init__.py
```diff
@@ -44,6 +46,25 @@ def wait_until(self, test_function, timelimit=90):
             return False
 
 
+class CRObject(OpenShiftObject):
+    """Custom APIObjects that implements methods that improves manipulation with CR objects"""
+
+    def safe_apply(self):
+        """Reimplementation of modify_and_apply from OpenshiftObject"""
```
pehala, line 53: Better description

testsuite/openshift/kuadrant.py
```diff
@@ -1,29 +1,87 @@
...
+            self.kuadrant_cr.model.spec[self.spec_name][name] = dataclasses.asdict(value)
+        else:
+            self.kuadrant_cr.model.spec[self.spec_name][name] = value
+
+    def wait_for_ready(self):
+        """Waits until Kuadrant CR reports ready status"""
+        self.kuadrant_cr.wait_for_ready()
+
+
+class _Authorino(_CRObject):
```
pehala, line 50: No need to keep this private

pehala, line 50: I would not introduce these subclasses now, we can add them later when they are actually needed

testsuite/openshift/kuadrant.py
```diff
@@ -1,29 +1,87 @@
...
+class KuadrantCR(CRObject):
     """Represents Kuadrant CR objects"""
 
     LIMITADOR = "limitador-limitador"
 
-    @property
-    def limitador(self) -> dict:
-        """Returns spec.limitador from Kuadrant object"""
-        return self.model.spec.setdefault("limitador", {})
+    @cached_property
```
pehala, line 63: Wont hurt but not sure if needed, there is not state saved in the class and it is inexpensive operation

testsuite/openshift/limitador.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class LimitadorCR(CRObject):
+    """Represents Limitador CR objects"""
+
+    @property
+    def deployment(self) -> Deployment:
+        """Returns Deployment object for this Limitador"""
+        with self.context:
+            return selector("deployment/limitador-limitador").object(cls=Deployment)
```
pehala, line 16: Name should be dynamic

testsuite/tests/conftest.py
```diff
@@ -226,6 +226,9 @@ def oidc_provider(keycloak) -> OIDCProvider:
 @pytest.fixture(scope="session")
 def blame(request):
     """Returns function that will add random identifier to the name"""
+    if "tester" in settings:
```
pehala, line 229: unrelated change, probably deserves a separate PR

testsuite/openshift/kuadrant.py
```diff
@@ -1,29 +1,87 @@
...
-    def limitador(self, value: dict):
-        """Sets the value of spec.limitador"""
-        self.model.spec["limitador"] = value
+    @cached_property
+    def limitador(self) -> _Limitador:
+        """Returns spec.limitador from Kuadrant object"""
+        return _Limitador(self, "limitador")
 
     @property
     def limitador_deployment(self):
```
pehala, line 74: This is not consistent with LimitadorCR

testsuite/tests/kuadrant/reconciliation/cr_configuration/test_reconcile_limitador.py
```diff
@@ -18,31 +21,43 @@ def kuadrant_clean_up(request, kuadrant):
...
+    value = asdict(ContainerResources(requests_cpu="300m", requests_memory="40Mi"))
+
+    kuadrant.limitador["resourceRequirements"] = value
+    kuadrant.safe_apply()
+    kuadrant.wait_for_ready()
 
-    assert value == kuadrant.limitador[field]
-    assert str(value) in str(kuadrant.limitador_deployment.model.spec)
+    assert kuadrant.limitador["resourceRequirements"] == value
+    assert kuadrant.limitador_deployment.model.spec.template.spec.containers[0]["resources"] == value
```
pehala, line 60: I would use asdict here instead of above

testsuite/openshift/kuadrant.py
```diff
@@ -1,29 +1,78 @@
 """Kuadrant CR object"""
 
-from openshift_client import selector
+import dataclasses
```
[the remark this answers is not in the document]
jsmolar: for set item: `if dataclasses.is_dataclass(value):`

testsuite/openshift/kuadrant.py
jsmolar: I removed LimitadorCR as it is not used in this PR

testsuite/openshift/kuadrant.py
```diff
@@ -1,29 +1,69 @@
...
+
+class KuadrantSection:
+    """
+    Base class for Kuadrant sub components:
+        Authorino - spec.authorino
+        Limitador - spec.limitador
+    """
+
+    def __init__(self, kuadrant_cr, spec_name):
+        super().__init__()
```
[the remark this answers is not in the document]
jsmolar: We discussed this with Petr, that this is a good practice

averevki (approved):

pehala (approved):

PR #433, opened 2024-06-11
Introduce Multi-cluster tests

### LINKED ISSUE #416: Multicluster support
**Summary**

Add support for running testsuite in multicluster environment.
With focus on distributed DNSPolicy and global RateLimitPolicy

**Documentation**

**Additional context**

**Test cases**

pehala:
* Replace `spokes` with `additional_clusters`
   *  Make the option a list instead of a dict
* `service_protection` config options are no longer clients
* Move Single-cluster specific stuff to Kuadrant/conftest.py 
* Add `multicluster` top level test directory for multicluster tests
* Add `dnspython` to check validity of DNSRecords

Closes #416


config/settings.yaml
```diff
@@ -26,6 +26,7 @@ default:
     reports_dir: "reports"
   default_exposer: "openshift"
   control_plane:
+    additional_clusters: {}
```
ficap, line 29: is it a list as stated in template or a dict?

testsuite/config/openshift_loader.py
```diff
@@ -26,20 +26,18 @@ def load(obj, env=None, silent=True, key=None, filename=None):
         tools = client.change_project(obj["tools"]["project"])
     obj["tools"] = tools
 
-    service_protection = obj.setdefault("service_protection", {})
-    inject_client(service_protection, client, "project")
-    inject_client(service_protection, client, "project2")
+    # service_protection = obj.setdefault("service_protection", {})
```
ficap, line 29: do something with this commented code. either delete or surround by if condition if needed in only some cases

testsuite/tests/multicluster/conftest.py
```diff
@@ -0,0 +1,184 @@
...
+
+@pytest.fixture(scope="session")
+def openshifts(testconfig, hub_openshift, skip_or_fail) -> list[OpenShiftClient]:
+    """Returns list of all OpenShifts on which to run Multicluster"""
+    additional_clusters = testconfig["control_plane"]["additional_clusters"]
+    if len(additional_clusters) == 0:
+        skip_or_fail("Only one cluster was provided for multi-cluster tests")
+    return [
+        hub_openshift,
+        *(cluster.change_project(testconfig["service_protection"]["project"]) for cluster in additional_clusters),
```
ficap, line 55: I know it's done on multiple places but why do you change project to service_protection.project instead of using the one passed in cluster definition? (what if in a weird scenario one wants to use different namespaces on different clusters?)

pehala: Cluster project is no longer used (because you need at least 3 namespaces not one) and using different namespaces for each cluster is not supported for same reason

ficap: if so I'd remove it from settings.local.yaml etc. to avoid confusion (of course in another PR :) ) just came around it and found it a bit weird

testsuite/tests/multicluster/conftest.py
```diff
@@ -0,0 +1,184 @@
...
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, routes, dns_policies, tls_policies):  # pylint: disable=unused-argument
+    """Commits all policies before tests"""
+    components = [*dns_policies.values(), *tls_policies.values()]
+    for component in components:
+        request.addfinalizer(component.delete)
+        component.commit()
+    for component in components:
+        component.wait_for_ready()
```
trepel, line 184: I tried running using kua-stage-single, kua-multi-1, kua-multi-2 and here I needed to increase the timeout. TLSPolicies did not get ready in time, it was due to Certificate was not issued quick enough. If I used 180s timeout the test passed ok.

testsuite/tests/multicluster/conftest.py
pehala: Done

testsuite/tests/multicluster/conftest.py
pehala: It seems fixed

Makefile
```diff
@@ -45,23 +45,28 @@ test pytest tests: kuadrant
...
 kuadrant: poetry-no-dev
-	$(PYTEST) -n4 -m 'not standalone_only' --dist loadfile --enforce $(flags) testsuite
+	$(PYTEST) -n4 -m 'not standalone_only and not multicluster' --dist loadfile --enforce $(flags) testsuite
 
 kuadrant-only: ## Run Kuadrant-only tests
 kuadrant-only: poetry-no-dev
-	$(PYTEST) -n4 -m 'kuadrant_only and not standalone_only' --dist loadfile --enforce $(flags) testsuite
+	$(PYTEST) -n4 -m 'kuadrant_only and not standalone_only and not multicluster' --dist loadfile --enforce $(flags) testsuite
+
+multicluster: ## Run Multicluster only tests
```
jsmolar, line 66: add `multicluster` to `PHONY` targets. And remove `mgc`

testsuite/capabilities.py
```diff
@@ -10,20 +10,20 @@
...
     """Returns True, if Kuadrant deployment is present and should be used"""
-    if spokes := weakget(settings)["control_plane"]["spokes"] % {}:
-        for name, openshift in spokes.items():
+    project = settings["service_protection"]["system_project"]
+    if clusters := weakget(settings)["control_plane"]["additional_clusters"] % []:
+        for openshift in clusters:
+            name = openshift.api_url
             # Try if Kuadrant is deployed
             if not openshift.connected:
                 return False, f"Spoke {name} is not connected"
```
trepel, line 19: "Spoke" is not the right terminology anymore, there are two occurrences in this file that would be nice to replace with something more accurate. Maybe just `Cluster` ?

testsuite/tests/multicluster/conftest.py
trepel: Yes, it works now, thanks!

testsuite/tests/multicluster/test_multicluster_dns.py
```diff
@@ -0,0 +1,33 @@
+"""Tests DNS/TLSPolicy across multiple clusters"""
+
+import dns.resolver
+import pytest
+
+pytestmark = [pytest.mark.multicluster]
```
averevki, line 6: there should also be other marks, no? if it is explicitly set in Makefile to not run multicluster tests with other targets. I see at least additional dns and tls marks here

pehala: For now, the multicluster tests are completely separated from other tests, due to their very different setup and configuration requirements, so I didnt add more marks

averevki (dismissed):

trepel (approved):
The test passed using kua- -stage and -multi-1/2 clusters. Code looks good to me. What would be nice to have is more descriptive error messages in case of issues with clusters. E.g. if invalid token is passed for `additional_clusters` config  the error message was `error: current-context must exist in order to minify`. No need as part of this PR ofc.

pehala:
Rebased to resolve merge conflicts

PR #437, opened 2024-06-12
Add tests for Defaults

### LINKED ISSUE #375: Atomic defaults tests
**Summary**

The kuadrant operator now allows defaults to be set for AuthPolicy and RateLimitPolicy. Test that these defaults work properly and are mutually exclusive with setting policy in spec.

**Documentation**

Implementing PRs: https://github.com/Kuadrant/kuadrant-operator/pull/503, https://github.com/Kuadrant/kuadrant-operator/pull/456, https://github.com/Kuadrant/kuadrant-operator/pull/536
RFC: https://github.com/Kuadrant/architecture/blob/main/rfcs/0009-defaults-and-overrides.md

**Test cases**

Repeat each for AuthPolicy and RateLimitPolicy

* Tests functionality of rules set within the `defaults` block
    - Set HTTPRoute attached policy with an explicit defaults block
    - verify the policy inside the defaults block functions as a normal policy
 
* Test mutual exclusivity of `defaults` block with other rules
    - set HTTPRoute attached policy configured both with regular config and config inside a `defaults` block
    - Check that policy is rejected due to `Implicit and explicit defaults are mutually exclusive`

* Test policy inheritability and PolicyAffected condition
    -  Attach a policy with a configured `defaults` block to your gateway
    - Check the status of the gateway's HTTPRoute
    - status should specify `Object affected by {Auth|RateLimit}Policy x`
    - Curl the route to verify the policy attached to gateway got applied to your HTTPRoute

* Test policy discoverability
    -  Attach a policy with a configured `defaults` block to your gateway
    - Create a new HTTPRoute attached to your gateway
    - Check HTTPRoute status, it should specify `Object affected by {Auth|RateLimit}Policy x`
    - Curl the route to verify the policy attached to gateway got applied to your HTTPRoute

averevki:
### Changes

- Add the Defaults&Overrides interface to the testsuite
- Add tests for the Defaults

Closes #375


testsuite/tests/kuadrant/defaults_overrides/__init__.py
pehala, line 1: If you do not need any conftest you can simply remove this directory

testsuite/policy/authorization/sections.py
```diff
@@ -67,10 +67,18 @@ def section(self):
...
-        self.section.update({name: value})
+
+        if defaults:
+            section = self.obj.model.spec.setdefault("defaults", {}).setdefault("rules", {})
+        elif overrides:
+            section = self.obj.model.spec.setdefault("overrides", {}).setdefault("rules", {})
+        else:
+            section = self.section
+
+        section.setdefault(self.section_name, {})[name] = value
```
pehala, line 81: `section.setdefault(self.section_name, {})[name] = value` if fundamentally different that previous `self.section.update({name: value})` why? Even if I do not specify overrid or default the call is simply very different

testsuite/tests/kuadrant/defaults_overrides/defaults/test_rules_exclusivity.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.fixture(scope="module")
+def commit():
+    """We need to try to commit objects during the actual test"""
+    return None
+
+
+def test_rules_exclusivity(authorization, rate_limit):
+    """Test that server will reject object with implicit and explicit defaults defined simultaneously"""
+    for component in [authorization, rate_limit]:
+        result = component._object_def_action("create", auto_raise=False)  # pylint: disable=protected-access
```
pehala, line 35: This is very weird call I would use component.commit and pytest.raises

averevki: I want to catch mutually exclusive error message here. Commit/create from the oc library would show only something basic like `Error during object creation`

pehala: I believe you can still instropect the error if caught that way. In any way using protected function doesnt seem optimal and there is almost certainly a better method

averevki: You actually were right, pytest.rises match will catch the message. Thanks

testsuite/tests/kuadrant/defaults/test_rules_exclusivity.py
```diff
@@ -0,0 +1,37 @@
...
+    return rate_limit
+
+
+@pytest.fixture(scope="module")
+def commit():
+    """We need to try to commit objects during the actual test"""
+    return None
+
+
+def test_rules_exclusivity(authorization, rate_limit):
```
pehala, line 33: how about parametrizing it instead? In that wa, it will be two separate testcase and it will be much clearer what failed

testsuite/tests/kuadrant/defaults/test_basic.py
```diff
@@ -0,0 +1,45 @@
+"""Test basic enforcement of the rules inside the 'defaults' block of the AuthPolicy and RateLimitPolicy"""
+
+import pytest
+
+from testsuite.httpx.auth import HttpxOidcClientAuth
+from testsuite.policy.rate_limit_policy import Limit
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.limitador]
+
+LIMIT = Limit(3, 5)
```
martinhesko, line 10: Why has this been defined separately as a global variable?

averevki: For consistency across usages. Fixture seems too expansive for this. Tell me if you know of any disadvantages

martinhesko: Makes sense, LGTM

martinhesko (approved):

testsuite/policy/authorization/sections.py
```diff
@@ -67,10 +67,18 @@ def section(self):
         """The actual dict section which will be edited"""
         return self.obj.auth_section.setdefault(self.section_name, {})
 
-    def add_item(self, name: str, value: dict, **common_features):
+    def add_item(self, name: str, value: dict, *, defaults: bool = False, overrides: bool = False, **common_features):
```
pehala, line 70: How will this work with AuthConfig?

averevki: AuthConfig doesn't support D&O. Only AuthPolicy and RateLimitPolicy

pehala: Thats precisely why I am asking, Sections are used by both AuthConfig and AuthPolicy

averevki: The idea was "just don't use it with AuthConfig" 😄 . Pytest marks should restrict this. Do you want this to be completely forbidden?

pehala: Well, I think we are solving defaults in a wrong place. My issues with this solution:
1) It doesnt clearly show how defaults are applied:
  * Consider how we add stuff now: `authorization.identity.add_oidc` -> this clearly indicates that we go into identity section and add some infor about OIDC. It could be better but I still think it conveys information nicely. 
  * Now look how default are done `authorization.identity.add_oidc(default=True)` ->If I read this I will assume it is the same section but we add default parameter somewhere (also could be only me), but that is not true. We need to add in a completely different section (`defaults` instead of `rules`) and I would never guessed it in a first place nor would I be able to replicate it directly in AP.
2) It is used by AuthConfig as well. While this could be easily ignored, it adds another bit of abstraction magic to already magic class and now will not work universally, as was originally intended, but that was long time ago and we can agree to change it.

My preferred way to deal with D&O would be like this as you are adding it to a completely different section:
`authorization.defaults.identity.add_oidc`, `authorization.overrides.identity.add_oidc`, `authorization.rules.identity.add_oidc` (this could be default if you dont specify anything to be backwards compatible).

This would be trickier to implement but I think it would be a cleaner solution. Question is if this solution (or any other cleaner one) warrants the time investment needed.

averevki: These examples are definitely cleaner. I'll look what I can do. Do you want to merge these tests in the meantime? I think it is easily refactorable with what I have now.

Issue #406, opened 2024-05-23
Add tests for multiple HTTPRoutes having/inheriting identical hostnames


trepel:
**Summary**

If there are more HTTPRoutes effectively declare the same hostname the attachment of policies to such HTTPRoutes may lead to unexpected behavior. Add the tests for current behavior as well as tests for desired behavior (these will fail, use xfail for them). This way we have tests ready for desired behavior in advance and also we are able to catch regressions in current behavior.

Add tests for both scenarios described in the docs:
1) Two HTTPRoutes declaring the same hostname and policy attached to each of them
2) Two HTTPRoutes declaring the same hostname, one policy attached to HTTPRoute, another one to parent's Gateway

Ideally to do this for both AuthPolicy and RatelimitPolicy (thus ending with 4 scenarios in total)

Desired behavior is that the policy status is correctly reflected in policy's CR, ie `Enforced: true` is present if and only if the policy is indeed successfully enforced. In case of conflict the first policy (meaning the policy that was created first)  wins and the 2nd policy won't be enforced (and this being properly reflected in 2nd policy's CR).

**Documentation**

[AuthPolicy current behavior](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/auth.md#limitation-multiple-network-resources-with-identical-hostnames)

[RateLimitPolicy current behavior](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/rate-limiting.md#limitation-multiple-network-resources-with-identical-hostnames)

[xfail](https://docs.pytest.org/en/latest/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail)

[Desired behavior discussed in slack](https://kubernetes.slack.com/archives/C05J0D0V525/p1716288141262309?thread_ts=1716274559.142989&cid=C05J0D0V525)

**Test cases**

Split test case into sub-tasks if needed. It should help with tracking progress and navigating more complex tasks.

PR #439, opened 2024-06-14
Added identical hostnames test for AuthPolicy

trepel:
### Overview

Partially addresses https://github.com/Kuadrant/testsuite/issues/406

It covers two scenarios, both testing current behavior
1) Two AuthPolicies being attached to GW and HTTPRoute with identical hostname respectively
2) Two AuthPolicies being attached to two HTTPRoutes with identical hostname each

For details see https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/auth.md#limitation-multiple-network-resources-with-identical-hostnames

### Verification Steps

Eye review, execute both tests and make sure they pass.

### Note

I will proceed with RateLimitPolicy (RLP) scenarios once this is reviewed so that I can incorporate feedback from this PR there. The 1) scenario for RLP is identical, I will think of using parametrized fixture called `policy` instead of having two identical tests one using AuthPolicy and other RLP.
The 2) scenario for RLP is slightly different, not sure if it makes sense to try to use the same test for both

### DIFF
--- a/testsuite/policy/__init__.py
+++ b/testsuite/policy/__init__.py
@@ -23,3 +23,8 @@ class Policy(OpenShiftObject):
         """Wait for a Policy to be Enforced"""
         success = self.wait_until(has_condition("Enforced", "True"))
         assert success, f"{self.kind()} did not get ready in time"
+
+    def wait_for_accepted(self):
+        """Wait for a Policy to be Accepted"""
+        success = self.wait_until(has_condition("Accepted", "True"))
+        assert success, f"{self.kind()} did not get accepted in time"
--- a/testsuite/policy/authorization/auth_policy.py
+++ b/testsuite/policy/authorization/auth_policy.py
@@ -13,7 +13,7 @@ if TYPE_CHECKING:
     from . import Rule
 
 
-class AuthPolicy(AuthConfig, Policy):
+class AuthPolicy(Policy, AuthConfig):
     """AuthPolicy object, it serves as Kuadrants AuthConfig"""
 
     @property
--- a/testsuite/tests/kuadrant/identical_hostnames/conftest.py
+++ b/testsuite/tests/kuadrant/identical_hostnames/conftest.py
@@ -1,0 +1,23 @@
+"""Conftest for "identical hostname" tests"""
+
+import pytest
+
+from testsuite.gateway import GatewayRoute
+from testsuite.gateway.gateway_api.route import HTTPRoute
+
+
+@pytest.fixture(scope="module", autouse=True)
+def route2(request, gateway, blame, hostname, backend, module_label) -> GatewayRoute:
+    """HTTPRoute object serving as a 2nd route declaring identical hostname but different path"""
+    route = HTTPRoute.create_instance(gateway.openshift, blame("route"), gateway, {"app": module_label})
+    route.add_hostname(hostname.hostname)
+    route.add_backend(backend, "/anything/")
+    request.addfinalizer(route.delete)
+    route.commit()
+    return route
+
+
+@pytest.fixture(scope="module")
+def authorization_name2(blame):
+    """Name of the 2nd Authorization resource"""
+    return blame("authz2")
--- a/testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_gw_and_route.py
+++ b/testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_gw_and_route.py
@@ -1,0 +1,61 @@
+"""
+Tests behavior of using one HTTPRoute declaring the same hostname as parent Gateway related to AuthPolicy.
+https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/auth.md#limitation-multiple-network-resources-with-identical-hostnames
+(second topology mentioned there)
+"""
+
+import pytest
+
+from testsuite.policy.authorization.auth_policy import AuthPolicy
+
+pytestmark = [pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="class", autouse=True)
+def authorization2(request, gateway, authorization_name2, openshift, label):
+    """2nd Authorization object"""
+    auth_policy = AuthPolicy.create_instance(openshift, authorization_name2, gateway, labels={"testRun": label})
+    auth_policy.authorization.add_opa_policy("rego", "allow = false")
+    request.addfinalizer(auth_policy.delete)
+    auth_policy.commit()
+    auth_policy.wait_for_ready()
+    return auth_policy
+
+
+def test_identical_hostnames_auth_on_gw_and_route_ignored(client, authorization, hostname):
+    """
+    Tests that Gateway-attached AuthPolicy is ignored on 'route2' if both 'route' and 'route2' declare
+    identical hostname and there is another AuthPolicy already successfully enforced on 'route'.
+    Setup:
+        - Two HTTPRoutes declaring identical hostnames but different paths ('/' and '/anything/')
+        - Empty AuthPolicy enforced on the '/' HTTPRoute
+        - 'deny-all' AuthPolicy (created after Empty AuthPolicy) enforced on the Gateway
+    Test:
+        - Send a request via 'route' and assert that response status code is 200
+        - Send a request via 'route2' and assert that response status code is 200
+        - Delete the Empty AuthPolicy
+        - Send a request via both routes
+        - Assert that both response status codes are 403 (Forbidden)
+    """
+
+    # Verify that the Empty AuthPolicy is still enforced despite 'deny-all' AuthPolicy being enforced too now
+    authorization.wait_for_ready()
+
+    # Access via 'route' is allowed due to Empty AuthPolicy
+    response = client.get("/get")
+    assert response.status_code == 200
+
+    # Despite 'deny-all' Gateway AuthPolicy reporting being successfully enforced
+    # it is still allowed to access the resources via 'route2'
+    response = client.get("/anything/get")
+    assert response.status_code == 200
+
+    # Deletion of Empty AuthPolicy should make the 'deny-all' Gateway AuthPolicy effectively enforced on both routes.
+    # It might take some time hence the use of retry client.
+    authorization.delete()
+    with hostname.client(retry_codes={200}) as retry_client:
+        response = retry_client.get("/get")
+        assert response.status_code == 403
+
+    response = client.get("/anything/get")
+    assert response.status_code == 403
--- a/testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
+++ b/testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
@@ -1,0 +1,79 @@
+"""
+Tests behavior of using two HTTPRoutes declaring the same hostname related to AuthPolicy.
+https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/auth.md#limitation-multiple-network-resources-with-identical-hostnames
+(the first topology mentioned there)
+"""
+
+import pytest
+
+from testsuite.policy import has_condition
+from testsuite.policy.authorization.auth_policy import AuthPolicy
+
+pytestmark = [pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="class")
+def authorization2(request, route2, authorization_name2, openshift, label):
+    """2nd Authorization object"""
+    auth_policy = AuthPolicy.create_instance(openshift, authorization_name2, route2, labels={"testRun": label})
+    auth_policy.authorization.add_opa_policy("rego", "allow = false")
+    request.addfinalizer(auth_policy.delete)
+    auth_policy.commit()
+    auth_policy.wait_for_accepted()
+    return auth_policy
+
+
+def test_identical_hostnames_auth_on_routes_rejected(client, authorization, authorization2):
+    """
+    Tests that 2nd AuthPolicy is rejected on 'route2' declaring identical hostname as 'route' with another
+    AuthPolicy already successfully enforced on 'route'.
+    Setup:
+        - Two HTTPRoutes declaring identical hostnames but different paths ('/' and '/anything/')
+        - Empty AuthPolicy enforced on the '/' HTTPRoute
+        - 'deny-all' AuthPolicy (created after Empty AuthPolicy) accepted on the '/anything/' HTTPRoute
+    Test:
+        - Assert that 'deny-all' AuthPolicy reports an error
+        - Send a request via 'route' and assert that response status code is 200
+        - Send a request via 'route2' and assert that response status code is 200
+        - Delete the Empty AuthPolicy
+        - Change 'deny-all' AuthPolicy to trigger its reconciliation
+        - Send a request via both routes
+        - Assert that access via 'route' is 200 (OK)
+        - Assert that access via 'route2 is 403 (Forbidden)
+    """
+    assert authorization2.wait_until(
+        has_condition(
+            "Enforced",
+            "False",
+            "Unknown",
+            "AuthPolicy has encountered some issues: AuthScheme is not ready yet",
+        ),
+        timelimit=20,
+    ), (
+        f"AuthPolicy did not reach expected status (Enforced False), "
+        f"instead it was: {authorization2.refresh().model.status.conditions}"
+    )
+
+    response = client.get("/get")
+    assert response.status_code == 200
+
+    response = client.get("/anything/get")
+    assert response.status_code == 200
+
+    # Deletion of Empty AuthPolicy should allow for 'deny-all' AuthPolicy to be enforced successfully.
+    authorization.delete()
+
+    # 2nd AuthPolicy only recovers from the "AuthScheme is not ready yet" error if reconciliation is explicitly
+    # triggered, e.g. by changing the AuthPolicy CR content (changing AllValues to True in this particular case)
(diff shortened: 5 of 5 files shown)


testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Despite 'deny-all' Gateway AuthPolicy reporting being successfully enforced
+    # it is still allowed to access the resources via 'route2'
+    response = client.get("/anything/get")
+    assert response.status_code == 200
+
+    # Deletion of Empty AuthPolicy should make the 'deny-all' Gateway AuthPolicy effectively enforced.
+    # It might take some time hence the 15s sleep.
+    authorization.delete()
+    time.sleep(15)
```
pehala, line 57: This will be flaky, we should find better solution that this. Cannot you check the GW AuthPolicy if it is still partially or fully enforced?

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
```diff
@@ -0,0 +1,87 @@
...
+from testsuite.policy.authorization.auth_policy import AuthPolicy
+from testsuite.utils import has_condition
+
+pytestmark = [pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="class")
+def authorization2(request, kuadrant, route2, authorization_name2, openshift, label):
+    """2nd Authorization object (In case of Kuadrant AuthPolicy)"""
+    if kuadrant:
```
pehala, line 18: Not needed, as is is marked as `kuadrant_only`. Also if it ever happened it would fail with an error

testsuite/policy/authorization/auth_config.py
```diff
@@ -84,6 +84,16 @@ def wait_for_ready(self):
             assert success, f"{self.kind()} did not get ready in time"
             self.refresh()
 
+    def wait_for_accepted(self):
+        """Waits until authorization object reports accepted status"""
+        with oc.timeout(90):
+            success, _, _ = self.self_selector().until_all(
```
pehala, line 90: How about utilizing `wait_until`?

testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,63 @@
...
+
+import pytest
+
+from testsuite.policy.authorization.auth_policy import AuthPolicy
+
+pytestmark = [pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="class", autouse=True)
+def authorization2(request, kuadrant, gateway, authorization_name2, openshift, label):
```
pehala, line 17: Will this ever run? I do not see any changes to the commit or client fixtures

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
```diff
@@ -0,0 +1,87 @@
...
+    assert response.status_code == 200
+
+    response = client.get("/anything/get")
+    assert response.status_code == 200
+
+    # Deletion of Empty AuthPolicy should allow for 'deny-all' AuthPolicy to be enforced successfully.
+    authorization.delete()
+
+    # 2nd AuthPolicy only recovers from the "AuthScheme is not ready yet" error if reconciliation is explicitly
+    # triggered, e.g. by changing the AuthPolicy CR content (changing AllValues to True in this particular case)
```
pehala, line 69: Is that intended?

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
```diff
@@ -0,0 +1,87 @@
...
+        auth_policy = AuthPolicy.create_instance(openshift, authorization_name2, route2, labels={"testRun": label})
+        auth_policy.authorization.add_opa_policy("rego", "allow = false")
+        request.addfinalizer(auth_policy.delete)
+        auth_policy.commit()
+        auth_policy.wait_for_accepted()
+        return auth_policy
+    return None
+
+
+def test_2nd_authpolicy_rejected(client, authorization, authorization2):
```
pehala, line 28: Since only test names are propagated into ReportPortal, could we make this more descriptive? So the entire test case is deducible from this name alone?

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Deletion of Empty AuthPolicy should allow for 'deny-all' AuthPolicy to be enforced successfully.
+    authorization.delete()
+
+    # 2nd AuthPolicy only recovers from the "AuthScheme is not ready yet" error if reconciliation is explicitly
+    # triggered, e.g. by changing the AuthPolicy CR content (changing AllValues to True in this particular case)
+    authorization2.authorization.add_opa_policy("rego", "allow = false", True)
+    authorization2.refresh()
+
+    assert authorization2.wait_until(
```
pehala, line 73: Isnt this equal to `wait_for_ready`?

testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
trepel: No, not really. The GW AuthPolicy status is fully enforced even before the deletion so it can't be used. I share your concerns about stability though, I was just not able to come up with better solution.

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: You're right, that was copy/paste from `authorization` fixture but it is not needed for `authorization2` fixture. Will fix it. Thanks!

testsuite/policy/authorization/auth_config.py
trepel: I just copy/pasted `wait_for_ready` function that is defined just above this one. I will take a look into refactoring using `wait_until` and do it if there are no catches there.

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Probably not, I created an issue for this https://github.com/Kuadrant/kuadrant-operator/issues/702
I did that only after this PR but I add the link to the comment here.

testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
pehala: Missed the autouse 🤦

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
pehala: `test_identical_hostnames` is probably good enough

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
pehala: Definitely add a link here, once it is fixed we can use this test to verify that it works

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Nice observation, it seems so from a quick look at it. I will take a closer look and refactor using `wait_for_ready` whenever possible

testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
pehala: How about using retry for the next request?

trepel: +1 I like that, sounds like a better approach to me. Will refactor as suggested

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: There will be similar tests for RateLimitPolicy as well so at least `auth` should be part of the name as well. Also, there will be another tests that verify desired behavior (these are for current behavior) so we will end up having quite a few test functions that will need to be nicely named. I think I will go with `test_identical_hostnames_auth_on_routes` and `test_identical_hostnames_auth_on_gw_and_route`

testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
trepel: What is the best way to do this? I added `client.add_retry_code(200)` but not 100% sure if it does the right thing. Alternatively I can use some retry_strategy as described e.g. here:
https://scrapeops.io/python-web-scraping-playbook/python-httpx-retry-failed-requests/

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Done

testsuite/policy/authorization/auth_config.py
trepel: I refactored both `wait_for_accepted` and `wait_for_ready` using `wait_until` under the hood. It seems to work however, the original implementation also called `self.refresh()` at the very end which is not the case anymore. Is that `self.refresh()` call really needed?
Furthermore, I added `any(x.type == "Enforced" for x in obj.model.status.conditions)` into `wait_for_ready` because my understanding is that it should only return `True` if the policy is Enforced successfully (that's when policy is "ready" if I got it right).

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Link added.

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Renamed.

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Replaced with `wait_for_ready` as suggested.

testsuite/tests/kuadrant/identical_hostname/test_identical_hostnames_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Despite 'deny-all' Gateway AuthPolicy reporting being successfully enforced
+    # it is still allowed to access the resources via 'route2'
+    response = client.get("/anything/get")
+    assert response.status_code == 200
+
+    # Deletion of Empty AuthPolicy should make the 'deny-all' Gateway AuthPolicy effectively enforced.
+    # It might take some time hence the addition of retry code.
+    authorization.delete()
+    client.add_retry_code(200)
```
pehala, line 53: This is fine, but you can also create new client as is done in other tests (https://github.com/Kuadrant/testsuite/blob/3b7ff3a1efe8296f3ae9a25b9bc45a8dfd438976/testsuite/tests/kuadrant/reconciliation/test_httproute_delete.py#L24)

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
pehala: I think only test method names are propagated, not sure about test file names

testsuite/policy/authorization/auth_config.py
```diff
@@ -75,14 +73,20 @@ def remove_all_hosts(self):
...
-                success_func=lambda obj: len(obj.model.status.conditions) > 0
-                and all(x.status == "True" for x in obj.model.status.conditions)
-            )
-            assert success, f"{self.kind()} did not get ready in time"
-            self.refresh()
+        """Waits until authorization object reports Enforced=True status"""
+        success = self.wait_until(
+            lambda obj: all(x.status == "True" for x in obj.model.status.conditions)
+            and any(x.type == "Enforced" for x in obj.model.status.conditions),
+            90,
```
pehala, line 80: 90 is default, I would not explicitly repeat it here until we feel the 90 is really required here

testsuite/tests/kuadrant/identical_hostname/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+from testsuite.gateway.gateway_api.route import HTTPRoute
+
+
+@pytest.fixture(scope="module", autouse=True)
+def route2(request, kuadrant, gateway, blame, hostname, backend, module_label) -> GatewayRoute:
+    """HTTPRoute object serving as a 2nd route declaring identical hostname but different path"""
+    if kuadrant:
+        route = HTTPRoute.create_instance(gateway.openshift, blame("route"), gateway, {"app": module_label})
+    else:
+        route = EnvoyVirtualRoute.create_instance(gateway.openshift, blame("route"), gateway)
```
pehala, line 16: Not needed. It will never work with pure envoy

testsuite/tests/kuadrant/identical_hostname/test_identical_hostnames_auth_on_gw_and_route.py
trepel: yep, I can do it this way for the sake of consistency if for nothing else

testsuite/policy/authorization/auth_config.py
trepel: Ok, will refactor.

testsuite/tests/kuadrant/identical_hostname/conftest.py
trepel: Sure, will refactor. Thanks!

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: sorry, my bad. I renamed files instead of test functions :facepalm:

testsuite/tests/kuadrant/identical_hostname/conftest.py
trepel: Done

testsuite/policy/authorization/auth_config.py
trepel: Done

testsuite/tests/kuadrant/identical_hostname/test_identical_hostnames_auth_on_gw_and_route.py
trepel: Done

testsuite/tests/kuadrant/identical_hostname/test_auth_on_two_routes.py
trepel: Done

testsuite/tests/kuadrant/identical_hostname/test_auth_on_gw_and_route.py
trepel: Updated using a new retry client

testsuite/policy/authorization/auth_config.py
trepel: This will be addressed by https://github.com/Kuadrant/testsuite/pull/443 I will rebase once it gets merged.

testsuite/tests/kuadrant/identical_hostname/test_identical_hostnames_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,57 @@
...
+    response = client.get("/anything/get")
+    assert response.status_code == 200
+
+    # Deletion of Empty AuthPolicy should make the 'deny-all' Gateway AuthPolicy effectively enforced.
+    # It might take some time hence the use of retry client.
+    authorization.delete()
+    with hostname.client(retry_codes={200}) as retry_client:
+        response = retry_client.get("/get")
+        assert response.status_code == 403
+        response = retry_client.get("/anything/get")
```
pehala, line 56: I think this request does not need to be retried as by the time we get to it, it should be already stable

trepel: That's obviously right, I will refactor. Thanks!

trepel: Done

pehala (dismissed):

testsuite/policy/__init__.py
```diff
@@ -23,3 +23,10 @@ def wait_for_ready(self):
         """Wait for a Policy to be Enforced"""
         success = self.wait_until(has_condition("Enforced", "True"))
         assert success, f"{self.kind()} did not get ready in time"
+
+    def wait_for_accepted(self):
+        """Waits until authorization object reports Accepted=True status"""
+        success = self.wait_until(
+            lambda obj: any(x.status == "True" and x.type == "Accepted" for x in obj.model.status.conditions)
```
pehala, line 30: `has_condition`

trepel: Eh, you're right, will force push

trepel: Done

pehala (approved):

PR #440, opened 2024-06-14
Cleanup of Keycloak Realms created for tests

trepel:
### Overview

This seems to delete Keycloak Realms once the tests are over.

### Verification Steps

Eye review if it makes sense.
Run single test and verify there is no Keycloak Realm created by the test run there once the test is over. Try  `./kuadrant/authorino/dinosaur` test for instance as this one creates one additional (`admin`) Realm.

Run the whole test suite (`make kuadrant`) and verify that there is no Keycloak Realm created by the testsuite run present once the testsuite execution is over. For a quicker validation you can mark a few tests as "keycloak":
`pytestmark = [pytest.mark.authorino, pytest.mark.keycloak]`

Make sure that the marked tests use Keycloak and be sure to mark the `./kuadrant/authorino/dinosaur` test.

Then create a "keycloak" make target in Makefile:
```
keycloak: ## Run only keycloak related tests
keycloak: poetry-no-dev
	$(PYTEST) -n4 -m 'keycloak' --dist loadfile --enforce $(flags) testsuite
```

Then execute `make keycloak` and verify that there are no leftover Keycloak Realms present once the tests execution is over.

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -164,7 +164,7 @@ def keycloak(request, testconfig, blame, skip_or_fail):
             cnf["test_user"]["username"],
             cnf["test_user"]["password"],
         )
-
+        request.addfinalizer(info.delete)
         info.commit()
         return info
     except KeycloakAuthenticationError:
--- a/testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
+++ b/testsuite/tests/kuadrant/authorino/dinosaur/conftest.py
@@ -26,7 +26,10 @@ def admin_rhsso(blame, keycloak):
     )
 
     info.commit()
-    return info
+    yield info
+
+    # Realm cleanup
+    info.delete()
 
 
 @pytest.fixture(scope="module", autouse=True)


testsuite/tests/conftest.py
```diff
@@ -152,26 +152,30 @@ def openshift2(testconfig, skip_or_fail):
...
         info.commit()
-        return info
+        yield info
     except KeycloakAuthenticationError:
         return skip_or_fail("Unable to login into SSO, please check the credentials provided")
     except KeyError as exc:
         return skip_or_fail(f"SSO configuration item is missing: {exc}")
 
+    # Realm cleanup
+    info.master_realm.delete_realm(realm_name)
```
pehala, line 177: `info.delete()` is sufficient

pehala (commented):
Well, the missing delete explains why it was not cleaned up...

testsuite/tests/conftest.py
trepel: Thx, I missed this method. Will refactor as suggested.

trepel: Refactored!

testsuite/tests/conftest.py
```diff
@@ -166,11 +167,15 @@ def keycloak(request, testconfig, blame, skip_or_fail):
...
+        yield info
     except KeycloakAuthenticationError:
-        return skip_or_fail("Unable to login into SSO, please check the credentials provided")
+        yield skip_or_fail("Unable to login into SSO, please check the credentials provided")
     except KeyError as exc:
-        return skip_or_fail(f"SSO configuration item is missing: {exc}")
+        yield skip_or_fail(f"SSO configuration item is missing: {exc}")
+
+    # Realm cleanup
+    if info is not None:
```
pehala, line 177: if you need this you can use addfinalizer instead
```
request.addfinalizer(info.delete)
info.commit()
return info
```

pehala: and as a bonus you wont need to change the returns

trepel: I was thinking about using finalizers but internet told me that `yield` is preferred if one deals with single resource. Agree that that `None` check is not nice but it has to be there since the cleanup also gets executed in case of exception (because I had to replace `return` with `yield` in that exception clauses as well). If you consider the finalizer approach nicer in this particular case I will refactor. I think you are actually right, it is nicer.

trepel: Thinking about it, the downside would be that `info.delete()` is to be called even in case of exception on `info` object creation so there might be an (additional) exception raised on cleanup. Which is acceptable I think.

trepel: Refactored using finalizer approach

pehala (approved):

PR #443, opened 2024-06-17
Improve wait_for_ready timeout errors

pehala:
- Timeouts now return human readable message instead of an error for timeout
  - Any other errors will be propagated as usual 
- Use wait_until instead of until_all()
- Lower default timeout from 90sec to 60sec
   - Everything passed on my machine, we can see if it works (or I can delete it based on review)

### DIFF
--- a/testsuite/gateway/gateway_api/gateway.py
+++ b/testsuite/gateway/gateway_api/gateway.py
@@ -76,12 +76,9 @@ class KuadrantGateway(OpenShiftObject, Gateway):
 
     def wait_for_ready(self, timeout: int = 10 * 60):
         """Waits for the gateway to be ready in the sense of is_ready(self)"""
-        with oc.timeout(timeout):
-            success, _, _ = self.self_selector().until_all(
-                success_func=lambda obj: self.__class__(obj.model).is_ready()
-            )
-            assert success, "Gateway didn't get ready in time"
-            self.refresh()
+        success = self.wait_until(lambda obj: self.__class__(obj.model).is_ready(), timelimit=timeout)
+        assert success, "Gateway didn't get ready in time"
+        self.refresh()
 
     def is_affected_by(self, policy: Policy) -> bool:
         """Returns True, if affected by status is found within the object for the specific policy"""
--- a/testsuite/policy/authorization/auth_config.py
+++ b/testsuite/policy/authorization/auth_config.py
@@ -3,8 +3,6 @@
 from functools import cached_property
 from typing import Dict
 
-import openshift_client as oc
-
 from testsuite.utils import asdict
 from testsuite.openshift import OpenShiftObject, modify
 from testsuite.openshift.client import OpenShiftClient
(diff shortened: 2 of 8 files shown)


testsuite/policy/authorization/auth_config.py
```diff
@@ -76,13 +74,12 @@ def remove_all_hosts(self):
...
-        with oc.timeout(90):
-            success, _, _ = self.self_selector().until_all(
-                success_func=lambda obj: len(obj.model.status.conditions) > 0
-                and all(x.status == "True" for x in obj.model.status.conditions)
-            )
-            assert success, f"{self.kind()} did not get ready in time"
-            self.refresh()
+        success = self.wait_until(
+            lambda obj: len(obj.model.status.conditions) > 0
+            and all(x.status == "True" for x in obj.model.status.conditions)
```
trepel, line 79: This apparently works well but would not it be better to specifically check the `Enforced=True` ? I think that in context of Policies the "ready" actually means enforced, doesn't it?

```
lambda obj: all(x.status == "True" for x in obj.model.status.conditions)
            and any(x.type == "Enforced" for x in obj.model.status.conditions)
```

testsuite/policy/tls_policy.py
```diff
@@ -55,9 +53,5 @@ def __getitem__(self, key):
     def wait_for_ready(self):
         """TLSPolicy does not have Enforced
         https://github.com/Kuadrant/kuadrant-operator/issues/572"""
-        with oc.timeout(90):
-            success, _, _ = self.self_selector().until_all(
-                success_func=has_condition("Accepted", "True"),
-                tolerate_failures=5,
-            )
-            assert success, f"{self.kind()} did not get ready in time"
+        success = self.wait_until(has_condition("Accepted", "True"))
```
trepel, line 56: TLSPolicy has been updated with `Enforced` recently: https://github.com/Kuadrant/kuadrant-operator/pull/635
so this whole method can be removed since the version from `super` should be sufficient.

Actually, I think that `wait_for_ready` can be implemented once in base `Policy` class now, there's no need to have specific version of this method for particular policies at all.

testsuite/policy/authorization/auth_config.py
pehala: Authpolicy does have Enforced but AuthConfig is not a policy so I am not sure about that

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -76,12 +76,9 @@ def is_ready(self):
...
         """Waits for the gateway to be ready in the sense of is_ready(self)"""
-        with oc.timeout(timeout):
-            success, _, _ = self.self_selector().until_all(
-                success_func=lambda obj: self.__class__(obj.model).is_ready()
-            )
-            assert success, "Gateway didn't get ready in time"
-            self.refresh()
+        success = self.wait_until(lambda obj: self.__class__(obj.model).is_ready(), timelimit=timeout)
+        assert success, "Gateway didn't get ready in time"
+        self.refresh()
```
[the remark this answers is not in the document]
pehala: Not sure, it was there before so I kept it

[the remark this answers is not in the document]
trepel: Ack, let's keep it, no harm anyway

testsuite/policy/authorization/auth_config.py
trepel: Crap, sorry, I completely missed the fact that this is about AuthConfig, not AuthPolicy. You are right, only policies are enforced.

testsuite/policy/tls_policy.py
trepel: Eh, there is that additional `time.sleep(5)` for RLP so RLP needs its own `wait_for_ready` method.

trepel (approved):
I think that this is good to go. I ran some tests and they passed.

averevki (dismissed):

pehala:
Rebased

averevki (approved):

PR #446, opened 2024-06-18
Fix keycloak fixture unknown variable

averevki:
request variable is required to add a realm clean-up finalizer to the fixture

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -149,7 +149,7 @@ def openshift2(testconfig, skip_or_fail):
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


pehala (approved):

azgabur (approved):

trepel:
Hmm, I spent some time figuring out how this happened. It seems that GH allowed to merge https://github.com/Kuadrant/testsuite/pull/440 despite it not being rebased on `main`. And since https://github.com/Kuadrant/testsuite/pull/431/files got into `main` in the meantime and GH did not detect the conflict it led to this bug (if my analysis is correct). Any idea what to do to avoid this in the future?

trepel:
Ok, thanks. We can always revisit if this started happening more often.

PR #447, opened 2024-06-18
Use kubectl instead of oc

pehala:
* Replaces `oc` binary with `kubectl`
   * This is not configurable because it needs to be resolved before configuration is parsed
* Replace whoami command with kubeconfig introspection
* Replace binary in testsuite image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,9 @@ Bind a dir to /test-run-results to get reports "
 RUN useradd --no-log-init -u 1001 -g root -m testsuite
 RUN dnf install -y python3.11 python3.11-pip make git && dnf clean all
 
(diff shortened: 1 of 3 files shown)


averevki (approved):
I'm getting the `Error from server (BadRequest): error when creating \"STDIN\": Deployment in version \"v1\" cannot be handled as a Deployment: strict decoding error: unknown field \"spec.template.spec.volumes[1].secret.readOnly\", unknown field \"spec.template.spec.volumes[2].secret.readOnly\", unknown field \"spec.template.spec.volumes[3].secret.readOnly\"\n` error with kubectl.
The problem is with readOnly field in SecretVolume dataclass which does not exist (anymore?). This option resides under the volumeMount. I guess oc wasn't that strict about objects we created. Simply removing this parameter should suffice.

I'm creating the PR: #451. LGTM otherwise :+1:

PR #449, opened 2024-06-19
Automatic exposer detection

pehala:
* Automatic exposer detection
  * No longer need to explicitly state `openshift `or `kind `
  * WIll use OpenShift if routes exist and Kubernetes otherwise

Depends on #448


averevki (dismissed):

averevki (approved):

trepel (approved):
I ran a few tests both on Kind and Openshift. Worked as expected.

PR #451, opened 2024-06-19
Remove not existing readOnly field from the SecretVolume

averevki:
The `ReadOnly` field doesn't exist there. Probably was copied from the VolumeMount by mistake


pehala (approved):

trepel (approved):
Validated by running `flags="-vvv --standalone" make testsuite/tests/kuadrant/authorino/operator/tls/mtls/test_mtls_attributes.py::test_mtls_multiple_attributes_success`
With these changes it passed.

PR #452, opened 2024-06-19
Remove need to specify committed property

pehala:
Advantage: No need to specify committed when fetching object from Kubernetes
Disadvantage: One call to Kubernetes to check if it exists per object (it is cached)

Let me know what you think :)

### DIFF
--- a/testsuite/openshift/__init__.py
+++ b/testsuite/openshift/__init__.py
@@ -16,7 +16,14 @@ class OpenShiftObject(APIObject, LifecycleObject):
 
     def __init__(self, dict_to_model=None, string_to_model=None, context=None):
         super().__init__(dict_to_model, string_to_model, context)
-        self.committed = False
+        self._committed = None
+
+    @property
+    def committed(self):
+        """Returns True, if the objects is already committed to the server"""
+        if self._committed is None:
+            self._committed, _ = self.exists()
+        return self._committed
 
     def commit(self):
         """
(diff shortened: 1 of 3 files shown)


averevki:
It looks good, actual up-to-date state on objects overweight disadvantages for me :+1: 

> Disadvantage: One call to Kubernetes to check if it exists per object (it is cached)

What's cached? Committed property or API request?

pehala:
Property, so rhe call is not repeated

jsmolar:
I like this change `committed = True` after the selector always bothered me more than it should.  
Let's "undraft" this.

testsuite/openshift/__init__.py
```diff
@@ -16,22 +16,30 @@ class OpenShiftObject(APIObject, LifecycleObject):
...
     def __init__(self, dict_to_model=None, string_to_model=None, context=None):
         super().__init__(dict_to_model, string_to_model, context)
-        self.committed = False
+        self._committed = None
+
+    @property
+    def committed(self):
+        """Returns True, if the objects is already committed to the server"""
+        if self._committed is None:
+            exists, _ = self.exists()
```
jsmolar, line 25: This can be a one-liner. 
I am thinking of renaming `_committed` to `_exists`? But it may be BS.

pehala: how can this be a one liner?

jsmolar: `self._committed, _ = self.exists()`

pehala: Thanks, I did not think of that

jsmolar (approved):

averevki (approved):

PR #454, opened 2024-06-19
Move has_condition to policy module

pehala:
There is still dependence on `check_condition` from utils. That method is used in route and gateway and moving it here might cause dependency cycle

### DIFF
--- a/testsuite/policy/__init__.py
+++ b/testsuite/policy/__init__.py
@@ -1,7 +1,19 @@
 """Contains Base class for policies"""
 
 from testsuite.openshift import OpenShiftObject
-from testsuite.utils import has_condition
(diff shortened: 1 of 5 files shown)


averevki:
Couldn't has_condition be used for something other than policies in the future?

pehala:
> Couldn't has_condition be used for something other than policies in the future?

Theoretically yes

trepel (approved):
Validated via `make testsuite/tests/./kuadrant/gateway/reconciliation/test_gw_doesnt_exist.py` LGTM

jsmolar (approved):
Such methods (potentially used by any Kubernetes object) should have be in one common file.

PR #455, opened 2024-06-20
Update KuadrantCR sub-component deployment reference

jsmolar:


trepel (commented):
Works like a charm, I just think that `LIMITADOR = "limitador-limitador"` can be removed from `KuadrantCR` class since it is no longer used.

pehala (dismissed):

pehala (approved):

trepel (approved):

PR #456, opened 2024-06-20
Fix wrong key name in webhook test

averevki:
ref: https://docs.kuadrant.io/authorino/docs/features/#kubernetes-subjectaccessreview-authorizationkubernetessubjectaccessreview

P.S. Probably also failing only now because of the change to kubectl


pehala (approved):

trepel (approved):
Thanks for the link, the doc say this is OK. I ran the against Kind (kubectl) and kua-stage-single (oc) , all good.

PR #457, opened 2024-06-24
Add refresh to authorino object

azgabur:
This will fix failing authorino standalone test `test_deploy_authorino` due to testing old cached CR. Not sure when this was introduced.


pehala:
I removed it in https://github.com/Kuadrant/testsuite/commit/b33ef5a9729b65d2270bb6ff5998f337b457f162, because I was not sure if it is needed, turns out it is :D

pehala (dismissed):

testsuite/tests/kuadrant/authorino/operator/conftest.py
```diff
@@ -49,4 +49,5 @@ def authorino(openshift, blame, request, testconfig, label, authorino_parameters
     request.addfinalizer(authorino.delete)
     authorino.commit()
     authorino.wait_for_ready()
+    authorino.refresh()
```
pehala, line 52: Optionally, you can add it to wait_for_ready in authorino directly, or add it to wait_until so every wait is refreshed

azgabur: Yes this seems like better idea. So its called just in one place.

pehala (approved):

averevki (approved):

PR #460, opened 2024-06-25
Fix TLSPolicy timeouts

azgabur:
Some previous change lowered the TLSPolicy wait timeout resulting in TLSPolicy not getting ready erros, this PR will raise it to 180s which is sufficient even for the Letsecrypt test. 
From my testing when the TLSPolicy is `Enforced: true` additional waits are not needed, thats why I removed the sleep in the client fixture.
I also added parameter for the `wait_for_ready` function to set the timeout/timelimit duration more easily.

### DIFF
--- a/testsuite/policy/__init__.py
+++ b/testsuite/policy/__init__.py
@@ -19,9 +19,9 @@ def has_condition(condition_type, status="True", reason=None, message=None):
 class Policy(OpenShiftObject):
     """Base class with common functionality for all policies"""
 
-    def wait_for_ready(self):
+    def wait_for_ready(self, timelimit=60):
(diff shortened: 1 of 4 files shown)


pehala (commented):
I have a few problems with this PR:

1) Adding timelimit to wait_for_ready for only a few classes results in fragmentation, where 90% of our objects do not support timelimit and only those descended from Policy do. IF we want to add timelimit to wait_for_ready I would rather see it done in all wait_for_ready methods

2) This is needlesly big change in unrelated objects to fix this issue

azgabur:
I agree with you @pehala, the timeout handling needs bigger refactor then this. I will leave it for some future PR.

Closing in favor of #461

PR #462, opened 2024-06-25
Add identical hostnames test for RateLimitPolicy

trepel:
## Overview

Partially addresses https://github.com/Kuadrant/testsuite/issues/406
Follows up for https://github.com/Kuadrant/testsuite/pull/439 which added similar tests for AuthPolicy, this PR does the same for RateLimitPolicy (RLP).

It covers two scenarios, both testing current behavior

1. Two RLPs being attached to GW and HTTPRoute with identical hostname respectively
2. Two RLPs being attached to two HTTPRoutes with identical hostname each

For details see https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/rate-limiting.md#limitation-multiple-network-resources-with-identical-hostnames

Note that the 1. scenario has been fixed in the meantime, doc is not up-to-date. The GW RLP is now partially enforced meaning that it is actually applied on the route-b. Partially enforced policy has `Enforced=True` status as well, see https://github.com/Kuadrant/kuadrant-operator/pull/679

## What was done
- added two tests - one for each scenario
- updated `wait_for_ready` so that it only returns `True` only if policy is fully enforced (not only partially)
- added `wait_for_partial_enforced` method
- using `/anything/route1`  and `/anything/route2` paths instead of just `/` and `/anything` which better aligns with doc (`/foo` and `/bar`). Using just `/` is problematic because if is "superset" to `/anything/` so it was not ideal for 2. scenario
- renamed 2nd route to `route2`

## Verification Steps

Eye review, execute both tests and make sure they pass. Also execute the two tests for AuthPolicy

### DIFF
--- a/testsuite/policy/__init__.py
+++ b/testsuite/policy/__init__.py
@@ -20,11 +20,24 @@ class Policy(OpenShiftObject):
     """Base class with common functionality for all policies"""
 
     def wait_for_ready(self):
-        """Wait for a Policy to be Enforced"""
-        success = self.wait_until(has_condition("Enforced", "True"))
-        assert success, f"{self.kind()} did not get ready in time"
+        """Wait for a Policy to be ready"""
+        self.wait_for_full_enforced()
 
     def wait_for_accepted(self):
         """Wait for a Policy to be Accepted"""
         success = self.wait_until(has_condition("Accepted", "True"))
         assert success, f"{self.kind()} did not get accepted in time"
+
+    def wait_for_partial_enforced(self):
+        """Wait for a Policy to be partially Enforced"""
+        success = self.wait_until(
+            has_condition("Enforced", "True", "Enforced", f"{self.kind(False)} has been partially enforced")
+        )
+        assert success, f"{self.kind(False)} did not get partially enforced in time"
+
+    def wait_for_full_enforced(self):
+        """Wait for a Policy to be fully Enforced"""
+        success = self.wait_until(
+            has_condition("Enforced", "True", "Enforced", f"{self.kind(False)} has been successfully enforced")
+        )
+        assert success, f"{self.kind()} did not get fully enforced in time"
--- a/testsuite/tests/kuadrant/identical_hostnames/conftest.py
+++ b/testsuite/tests/kuadrant/identical_hostnames/conftest.py
@@ -6,18 +6,20 @@ from testsuite.gateway import GatewayRoute
 from testsuite.gateway.gateway_api.route import HTTPRoute
 
 
-@pytest.fixture(scope="module", autouse=True)
-def route2(request, gateway, blame, hostname, backend, module_label) -> GatewayRoute:
-    """HTTPRoute object serving as a 2nd route declaring identical hostname but different path"""
-    route = HTTPRoute.create_instance(gateway.openshift, blame("route"), gateway, {"app": module_label})
-    route.add_hostname(hostname.hostname)
-    route.add_backend(backend, "/anything/")
-    request.addfinalizer(route.delete)
-    route.commit()
+@pytest.fixture(scope="module")
+def route(route, backend):
+    """Adding /anything/route1 prefix to the backend"""
+    route.remove_all_backend()
+    route.add_backend(backend, "/anything/route1")
     return route
 
 
-@pytest.fixture(scope="module")
-def authorization_name2(blame):
-    """Name of the 2nd Authorization resource"""
-    return blame("authz2")
+@pytest.fixture(scope="module", autouse=True)
+def route2(request, gateway, blame, hostname, backend, module_label) -> GatewayRoute:
+    """HTTPRoute object serving as a 2nd route declaring identical hostname but different path"""
+    route = HTTPRoute.create_instance(gateway.openshift, blame("route2"), gateway, {"app": module_label})
+    route.add_hostname(hostname.hostname)
+    route.add_backend(backend, "/anything/route2")
+    request.addfinalizer(route.delete)
+    route.commit()
+    return route
--- a/testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
+++ b/testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
@@ -13,9 +13,9 @@ pytestmark = [pytest.mark.kuadrant_only]
 
 
 @pytest.fixture(scope="class")
-def authorization2(request, route2, authorization_name2, openshift, label):
+def authorization2(request, route2, blame, openshift, label):
     """2nd Authorization object"""
-    auth_policy = AuthPolicy.create_instance(openshift, authorization_name2, route2, labels={"testRun": label})
+    auth_policy = AuthPolicy.create_instance(openshift, blame("authz2"), route2, labels={"testRun": label})
     auth_policy.authorization.add_opa_policy("rego", "allow = false")
     request.addfinalizer(auth_policy.delete)
     auth_policy.commit()
@@ -28,9 +28,9 @@ def test_identical_hostnames_auth_on_routes_rejected(client, authorization, auth
     Tests that 2nd AuthPolicy is rejected on 'route2' declaring identical hostname as 'route' with another
     AuthPolicy already successfully enforced on 'route'.
     Setup:
-        - Two HTTPRoutes declaring identical hostnames but different paths ('/' and '/anything/')
-        - Empty AuthPolicy enforced on the '/' HTTPRoute
-        - 'deny-all' AuthPolicy (created after Empty AuthPolicy) accepted on the '/anything/' HTTPRoute
+        - Two HTTPRoutes declaring identical hostnames but different paths: '/anything/route1' and '/anything/route2'
+        - Empty AuthPolicy enforced on the '/anything/route1' HTTPRoute
+        - 'deny-all' AuthPolicy (created after Empty AuthPolicy) accepted on the '/anything/route2' HTTPRoute
     Test:
         - Assert that 'deny-all' AuthPolicy reports an error
         - Send a request via 'route' and assert that response status code is 200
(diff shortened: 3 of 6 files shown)


testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_gw_and_route.py
```diff
@@ -0,0 +1,67 @@
...
+
+import pytest
+
+from testsuite.policy.rate_limit_policy import RateLimitPolicy, Limit
+
+pytestmark = [pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="class")
+def rate_limit_policy2(request, gateway, rate_limit_policy_name2, openshift, label):
```
pehala, line 15: `rate_limit2` to align better with the `rate_limit` original fixture

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
```diff
@@ -28,8 +28,8 @@ def test_identical_hostnames_auth_on_routes_rejected(client, authorization, auth
     Tests that 2nd AuthPolicy is rejected on 'route2' declaring identical hostname as 'route' with another
     AuthPolicy already successfully enforced on 'route'.
     Setup:
-        - Two HTTPRoutes declaring identical hostnames but different paths ('/' and '/anything/')
-        - Empty AuthPolicy enforced on the '/' HTTPRoute
+        - Two HTTPRoutes declaring identical hostnames but different paths ('/nothing/' and '/anything/')
```
pehala, line 31: how using `/get` and `/anything` instead? To limit number of endpoints we require from httpbin

testsuite/policy/__init__.py
```diff
@@ -20,11 +20,20 @@ class Policy(OpenShiftObject):
     """Base class with common functionality for all policies"""
 
     def wait_for_ready(self):
-        """Wait for a Policy to be Enforced"""
-        success = self.wait_until(has_condition("Enforced", "True"))
+        """Wait for a Policy to be fully Enforced"""
+        success = self.wait_until(
```
pehala, line 24: How about adding `wait_for_full_enforced` method and having `wait_for_ready` just call that method, it might make it clearer when using both `wait_for_full_enforced` and `wait_for_partial_enforced` in one tests

testsuite/tests/kuadrant/identical_hostnames/conftest.py
```diff
@@ -21,3 +29,9 @@ def route2(request, gateway, blame, hostname, backend, module_label) -> GatewayR
 def authorization_name2(blame):
     """Name of the 2nd Authorization resource"""
     return blame("authz2")
+
+
+@pytest.fixture(scope="module")
+def rate_limit_policy_name2(blame):
```
pehala, line 35: Do you need to reuse the name anywhere? If no I would remove this fixture

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_routes.py
```diff
@@ -0,0 +1,84 @@
...
+    response = client.get("/nothing/get")
+    assert response.status_code == 200
+    # There might be more than two 200 OKs responses, might be due to '1rp10m' enforcement still being in progres
+    with hostname.client(retry_codes={200}) as retry_client:
+        response = retry_client.get("/nothing/get")
+        assert response.status_code == 429
+
+    # Access via 'route2' is now not limited at all
+    responses = client.get_many("/anything/get", 3)
+    assert all(
```
pehala, line 82: responses.assert_all(status_code=200)

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
trepel: I had it like that originally but it turns out that completely different paths are needed. The thing is that `/anything` is "subpath" of `/` so they are not completely different and the scenario is not reproducible. In the example in doc `/foo` and `/bar` paths are used so I used `/anything` and `/nothing`.

Btw, why do we need to limit the number of httpbin endpoints?

testsuite/tests/kuadrant/identical_hostnames/conftest.py
trepel: Not sure if I understand here. This "name2" fixture is used in both test classes and serves as a name of the 2nd RLP that is created via another fixture defined separately in each test class. So I thing this is the proper place for this fixture. Similarly the `authorization_name2` fixture is used for sister "AuthPolicy" tests.

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_routes.py
trepel: Nice, I missed this. Much nicer approach indeed. The only downside is that the you can't specify an error message but that's not a big deal at all. Would you like me to refactor just this specific occurrence or should I refactor all the occurrences in the tests? I  personally vote for the latter. And I would refactor the sister "AuthPolicy" tests as well.

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
pehala: `/nothing` is also "subpath" of `/`, so I still do not see how `/get` is different

The less endpoints we use on httpbin, the easier is to replace it in the future and our dependency on it is lower. It is not a big deal if we require the functionality but I do not think we do in thius case

testsuite/tests/kuadrant/identical_hostnames/conftest.py
pehala: Well, what other fixtures (other than `rate_limit_policy2`) use this fixture. If None, then it is not needed and you can generate name in the `rate_limit_policy2` directly.

testsuite/policy/__init__.py
trepel: yes, that's a good idea, will refactor. Key thing to discuss here is whether we want to use `wait_for_ready` for full enforcement only. It seems that we are in agreement that that's exactly what we want.

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_gw_and_route.py
trepel: Sure, will refactor

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
trepel: sorry, bad wording. I try again. I need two httproutes declaring identical hostnames but having different paths configured for backend. By different paths I also mean that one path must not be "subpath" of the other one. So in this PR I moved away from `/` and `/anything/` to `/nothing/` and `/anything/`

testsuite/tests/kuadrant/identical_hostnames/conftest.py
trepel: Ah, this is what you meant. I got it know. Make sense, will refactor.

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
pehala: Ok so it is just a prefix you are not changing what httpbin endpoints you are accessing?

trepel: Yes, just a prefix defined for backend in each httproute. However, I think I know what you are getting at now. I am mixing together backend prefixes in httproute CR and httpbin endpoints. `/anything/` is not ideal candidate for backend prefix because it is also actual httpbin endpoint (see https://eu.httpbin.org/#/Anything) So I should probably use completely different backend prefixes. Maybe something descriptive like `/route1/` and `/route2/`, wdyt?

pehala: > Maybe something descriptive like `/route1/` and `/route2/`, wdyt?

This is perfect! Also yes, that was exactly what confused me :D

testsuite/policy/__init__.py
pehala: I think it is a discussion for other PR, but my understanding is that `wait_for_ready` means that is is ready for general testsuite use, which for policies almost always is some kind of enforce, but might differ object to object.

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_routes.py
pehala: Definitely the latter, but not sure how the AP tests can be refactored

trepel: My bad again, AP tests do not use this at all, sorry for confusion! Will refactor all the occurrences here!

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_gw_and_route.py
trepel: Done

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
trepel: This works slightly different than I thought, the backend (httpbin) must have the endpoint that matches the `pathPrefix`. So I updated the PR to use `/anything/route1` and `/anything/route2`. It makes our test suite dependent on Anything HTTPBin endpoint but I can't see a way not to introduce this dependency

testsuite/tests/kuadrant/identical_hostnames/conftest.py
trepel: Done

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_rlp_on_routes.py
trepel: Done

testsuite/tests/kuadrant/identical_hostnames/test_identical_hostnames_auth_on_routes.py
pehala: `/anything` is fine, we already have that

pehala (approved):

jsmolar (approved):

Issue #365, opened 2024-03-12
Add tests for kuadrantctl CLI tool


azgabur:
**Summary**

kuadrantctl is a CLI tool for installing/uninstalling Kuadrant and generating Kuadrant resources HTTPRoute, AuthPolicy, and RateLimitPolicy from OpenAPI specification.

**Documentation**

https://github.com/Kuadrant/kuadrantctl

**Test cases**

- [ ] [Generating Gateway API HTTPRoute Objects](https://github.com/Kuadrant/kuadrantctl/blob/main/doc/generate-gateway-api-httproute.md)
- [ ] [Generating Kuadrant AuthPolicy Objects](https://github.com/Kuadrant/kuadrantctl/blob/main/doc/generate-kuadrant-auth-policy.md)
- [ ] [Generating Kuadrant RateLimitPolicy Objects](https://github.com/Kuadrant/kuadrantctl/blob/main/doc/generate-kuadrant-rate-limit-policy.md)

PR #463, opened 2024-06-25
Add Kuadrantctl tests

### LINKED ISSUE #434: Automation for KuadrantCTL
**Summary**

We need to include KuadrantCTL tests in our test suite to test policy generation based on OAS

- [ ] Figure out how to test binary for KuadrantCTL
- [ ] Implement tests

pehala:
* Kickoff of Kuadrantctl tests
   * Due to significant rework, works with only 0.2.4-alpha and newer releases 
   * New top level test folder `tests/kuadrantctl` and target `make kuadrantctl`
   * Minimal abstraction, 90% things are done directly and can be reproduced easily
* OAS manipulation
   * OAS is loaded as Pure OAS, that is further customized runtime by adding security or x-kuadrant fields
   * ~~Customization done in two parts `base_oas` and `oas_kuadrant` fixtures~~
      * ~~`base_oas` is for editing pure OAS specs, i.e adding security constrains~~
      * ~~`oas_kuadrant` is for adding x-kuadrant fields~~

~~Reviewable but not mergable due to a few outstanding bugs and enhancements:~~
Gonna be merged even with temporary fixes due to outstandiong issues with `kuadrantctl` itself
https://github.com/Kuadrant/kuadrantctl/issues/94, https://github.com/Kuadrant/kuadrantctl/issues/91, https://github.com/Kuadrant/kuadrantctl/commit/d198cc865a2bfa0fed621d0238280e17502deaaf

Fixes #434 #365

### DIFF
--- a/testsuite/gateway/gateway_api/route.py
+++ b/testsuite/gateway/gateway_api/route.py
@@ -114,3 +114,15 @@ class HTTPRoute(OpenShiftObject, GatewayRoute):
     @modify
     def remove_all_backend(self):
         self.model.spec.rules.clear()
+
+    def wait_for_ready(self):
+        """Waits until HTTPRoute is reconcilled by GatewayProvider"""
+
+        def _ready(obj):
+            for condition_set in obj.model.status.parents:
+                if condition_set.controllerName == "istio.io/gateway-controller":
+                    return (all(x.status == "True" for x in condition_set.conditions),)
+            return False
+
+        success = self.wait_until(_ready, timelimit=10)
+        assert success, f"{self.kind()} did got get ready in time"
--- a/testsuite/kuadrantctl.py
+++ b/testsuite/kuadrantctl.py
@@ -1,0 +1,52 @@
+# pylint: disable=line-too-long
+"""
+Help as of 0.2.3
+Kuadrant configuration command line utility
+
+Usage:
+  kuadrantctl [command]
+
+Available Commands:
+  completion  Generate the autocompletion script for the specified shell
+  generate    Commands related to kubernetes object generation
+    gatewayapi  Generate Gataway API resources
+        httproute   Generate Gateway API HTTPRoute from OpenAPI 3.0.X
+    kuadrant    Generate Kuadrant resources
+        authpolicy      Generate Kuadrant AuthPolicy from OpenAPI 3.0.X
+        ratelimitpolicy Generate Kuadrant Rate Limit Policy from OpenAPI 3.0.X
+
+
+  help        Help about any command
+  version     Print the version number of kuadrantctl
+
+Flags:
+  -h, --help                   help for httproute
+      --oas string             Path to OpenAPI spec file (in JSON or YAML format), URL, or '-' to read from standard input (required)
+  -o, --output-format string   Output format: 'yaml' or 'json'. (default "yaml")
+
+Global Flags:
+  -v, --verbose   verbose output
+
+
+Use "kuadrantctl [command] --help" for more information about a command.
+
+"""
+
+import subprocess
+
+
+class KuadrantCTL:
+    """Wrapper on top of kuadrantctl binary"""
+
+    def __init__(self, binary) -> None:
+        super().__init__()
+        self.binary = binary
+
+    def run(self, *args, **kwargs):
+        """Passes arguments to Subprocess.run, see that for more details"""
+        args = (self.binary, *args)
+        kwargs.setdefault("capture_output", True)
+        kwargs.setdefault("check", True)
+        kwargs.setdefault("text", True)
+        # We do supply value for check :)
+        return subprocess.run(args, **kwargs)  # pylint: disable= subprocess-run-check
--- a/testsuite/oas.py
+++ b/testsuite/oas.py
@@ -1,0 +1,49 @@
+"""OAS processing"""
+
+import contextlib
+import json
+import tempfile
+from collections import UserDict
+
+import yaml
+
+from testsuite.backend import Backend
+from testsuite.gateway import Referencable, Hostname
+
+
+@contextlib.contextmanager
+def as_tmp_file(text):
+    """Saves text in a temporary file and returns absolute path"""
+    with tempfile.NamedTemporaryFile("w") as file:
+        file.write(text)
+        file.flush()
+        yield file.name
+
+
+class OASWrapper(UserDict):
+    """Wrapper for OpenAPISpecification"""
+
+    def as_json(self):
+        """Returns OAS as JSON"""
+        return json.dumps(self.data)
+
+    def as_yaml(self):
+        """Returns OAS as YAML"""
+        return yaml.dump(self.data)
+
+    def add_backend_to_paths(self, backend: Backend):
+        """Adds backend to all paths, should be only used in tests that do not test this section"""
+        for path in self["paths"].values():
+            path["x-kuadrant"] = {
+                "backendRefs": [backend.reference],
+            }
+
+    def add_top_level_route(self, parent: Referencable, hostname: Hostname, name: str):
+        """Adds top-level x-kuadrant definition for Route, should be only used in tests that do not test this section"""
+        self["x-kuadrant"] = {
+            "route": {
+                "name": name,
+                "hostnames": [hostname.hostname],
+                "parentRefs": [parent.reference],
+            }
+        }
--- a/testsuite/tests/kuadrantctl/cli/test_simple_auth.py
+++ b/testsuite/tests/kuadrantctl/cli/test_simple_auth.py
@@ -1,0 +1,66 @@
+"""Tests that you can generate simple AuthPolicy, focused on the cmdline options more than on extension functionality"""
+
+import pytest
+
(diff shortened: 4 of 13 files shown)


testsuite/gateway/gateway_api/route.py
```diff
@@ -114,3 +114,15 @@ def add_backend(self, backend: "Backend", prefix="/"):
...
     def remove_all_backend(self):
         self.model.spec.rules.clear()
+
+    def wait_for_ready(self):
+        """Waits until HTTPRoute is reconcilled by GatewayProvider"""
+
+        def _ready(obj):
+            for condition_set in obj.model.status.parents:
+                if condition_set.controllerName == "istio.io/gateway-controller":
+                    return (all(x.status == "True" for x in condition_set.conditions),)
```
trepel, line 124: `all` returns `True` in case of empty collection (`condition_set.conditions` in this case). Is it ok in this context or better to add check that `len(condition_set.conditions) > 0`

testsuite/kuadrantctl.py
```diff
@@ -0,0 +1,52 @@
...
+
+Use "kuadrantctl [command] --help" for more information about a command.
+
+"""
+
+import subprocess
+
+
+class KuadrantCTL:
+    """The doc string"""
```
trepel, line 39: This doc string does not make me smarter

testsuite/oas.py
```diff
@@ -0,0 +1,29 @@
...
+
+class OASWrapper(UserDict):
+    """Simple wrapper for OAS"""
+
+    def as_json(self):
+        """Returns OAS as JSON"""
+        return json.dumps(self.data)
+
+    def as_yaml(self):
+        """Returns OAS as JSON"""
```
trepel, line 28: ...as YAML

testsuite/gateway/gateway_api/route.py
pehala: I think it should be fine, only potential issue is if the condition set was added but has no conditions when we check. I find it unlikely (I sincerely hope they are adding the condition set with condition at once, otherwise it would not make much sense tome) but I can add it if you want

trepel: I think you're right that it is quite unlikely, no need to add.

trepel:
Sorry for long comment,  I was lazy/unable to identify pieces of code that these bulletpoints relate to

1)  running `make kuadrantctl` causes three gateway CRs being created instantly at once. Not sure if this can cause any potential issues or not, I just find it strange.

2) RLPs and AuthPolicies name is `route-*` (it is actually the same as the target HTTPRoute). Again, not sure if this can cause any issues but less confusing names would be better

3) Intermittent error on RLP test if executed via `make kuadrantctl`:
```
testsuite/tests/kuadrantctl/cli/test_simple_limit.py:49: in test_generate_limit
    responses.assert_all(status_code=200)
testsuite/httpx/__init__.py:93: in assert_all
    assert request.status_code == status_code, (
E   AssertionError: Status code assertion failed for request 3 out of 3 requests: Result[status_code=429] != 200
---------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------
10:08:13 +0200 INFO:httpx:HTTP Request: GET http://hostname-trepel--qmv8-kuadrant.apps.kua-stage-single.osp.api-qe.eng.rdu2.redhat.com/anything "HTTP/1.1 200 OK"
10:08:13 +0200 INFO:httpx:HTTP Request: GET http://hostname-trepel--qmv8-kuadrant.apps.kua-stage-single.osp.api-qe.eng.rdu2.redhat.com/anything "HTTP/1.1 200 OK"
10:08:13 +0200 INFO:httpx:HTTP Request: GET http://hostname-trepel--qmv8-kuadrant.apps.kua-stage-single.osp.api-qe.eng.rdu2.redhat.com/anything "HTTP/1.1 429 Too Many Requests"
```
I suspect it is due to the fact that the RLPs are identical and HTTPRoute is re-used across the tests so there might be a clash time to time. Potential solution might be to make the duration shorter (10s or even less) or maybe re-create the HTTPRoute as well or maybe do not have the RLPs identical (might be enough to have some random string as part of the name or smt like that)

4) It does not work with kuadrantctl binary built from `main` branch.  It fails generating HTTPRoute with an error: `resource name may not be empty`. Probably bug in kuadrantctl itself rather that this PR so just fyi

pehala:
> 1. running `make kuadrantctl` causes three gateway CRs being created instantly at once. Not sure if this can cause any potential issues or not, I just find it strange.

This is plausible because tests are running in parallel

> 2\. RLPs and AuthPolicies name is `route-*` (it is actually the same as the target HTTPRoute). Again, not sure if this can cause any issues but less confusing names would be better

That is out of my control, I apply exactly what I get from the kuadrantctl and as far as I know there is not option to specify name


> 3\. Intermittent error on RLP test if executed via `make kuadrantctl`:

I will investigate that one, it could be because that route



> 4\. It does not work with kuadrantctl binary built from `main` branch.  It fails generating HTTPRoute with an error: `resource name may not be empty`. Probably bug in kuadrantctl itself rather that this PR so just fyi

Will try it out, so far I only tested released version

jsmolar:
`tests/conftest.py` should not contain `kuadrant` fixture

pehala:
> `tests/conftest.py` should not contain `kuadrant` fixture

I do not touch that file at all, not sure if the change needs to be done here

testsuite/oas.py
```diff
@@ -0,0 +1,49 @@
...
+        return yaml.dump(self.data)
+
+    def add_backend_to_paths(self, backend: Backend):
+        """Adds backend to all paths, should be only used in tests that do not test this section"""
+        for path in self["paths"].values():
+            path["x-kuadrant"] = {
+                "backendRefs": [backend.reference],
+            }
+
+    def top_level_route(self, parent: Referencable, hostname: Hostname, name: str):
```
trepel, line 41: Minor, but I can't help thinking that `add_top_level_route` would be a better name for this method, wdyt?

testsuite/tests/kuadrantctl/cli/test_simple_auth.py
```diff
@@ -0,0 +1,79 @@
...
+    anything = oas["paths"]["/anything"]
+    anything["x-kuadrant"] = {
+        "backendRefs": [backend.reference],
+    }
+    anything["get"]["security"] = [{"oidc": []}]
+    return oas
+
+
+@pytest.fixture(scope="module")
+def oas_kuadrant(base_oas, blame, gateway, hostname, backend):
```
trepel, line 40: This fixture is nowhere used and should be removed if I am not mistaken

trepel (commented):
This time all the tests passed using kuadrantctl built from current `main` branch but all the tests failed using kuadrantctl v0.2.3 (current GA). Exactly the opposite as it was before the force push (same error `resource name may not be empty`). I would hesitate merging this if not working on both (once 0.2.4 is GA'd we can merge imo, not sure about the timeline though).

I tried several times, I did not encounter the RLP issue I mentioned (unexpected 429s). It seems that re-creating the HTTPRoute indeed fixed the issue.

pehala:
> I would hesitate merging this if not working on both (once 0.2.4 is GA'd we can merge imo, not sure about the timeline though).

Sadly, that is not easily achievable. There is a breaking change (https://github.com/Kuadrant/kuadrantctl/pull/78) that changes a lot (where exactly the extension is specified) and I do not think it warrants the time investment and complexity needed to make it interchangable. However, we now have https://github.com/Kuadrant/kuadrantctl/releases/tag/v0.3.0, which is now GA

trepel:
> However, we now have https://github.com/Kuadrant/kuadrantctl/releases/tag/v0.3.0, which is now GA

Cool, I am happy to approve once you address the two comments then!

jsmolar:
@azgabur please review. I want you to take a look at this PR.

trepel (approved):
I am far from being and expert but this is good to go afaict. It works with kuadrantctl v0.3.0 (current GA).

testsuite/tests/kuadrantctl/cli/test_simple_auth.py
```diff
@@ -0,0 +1,66 @@
...
+    if stdin:
+        result = kuadrantctl.run("generate", "kuadrant", "authpolicy", "--oas", "-", input=encoded)
+    else:
+        with as_tmp_file(encoded) as file_name:
+            result = kuadrantctl.run("generate", "kuadrant", "authpolicy", "--oas", file_name)
+
+    policy = openshift.apply_from_string(result.stdout, AuthPolicy)
+    request.addfinalizer(policy.delete)
+
+    policy.wait_for_ready()
```
azgabur, line 54: This part is good candidate for a conftest fixture the future.

azgabur (approved):
LGTM, in this state it can be merged.

trepel (dismissed):
I have no idea what you force pushed, but LGTM still, I tried again and it worked (all tests passed).

jsmolar (approved):

pehala:
Removed committed

PR #466, opened 2024-06-27
Remove preexisting authorino

jsmolar:
We don't use preexisting authorino (at least, I am unaware of it). I propose to remove unnecessary code for the `authorino/operator` package and the `authorino` fixture.

I would like to add another topic for discussion. `operator/conftest.py` and `authorino/conftest.py` contain the same "copy-pasted" authorino fixtures, only the scope is different. Is there a good solution for this?

### DIFF
--- a/config/settings.local.yaml.tpl
+++ b/config/settings.local.yaml.tpl
@@ -40,10 +40,6 @@
 #       name: "istio-ingressgateway"
 #    authorino:
 #      image: "quay.io/kuadrant/authorino:latest"  # If specified will override the authorino image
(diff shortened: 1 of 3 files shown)


trepel:
I tried to run test_deploy_authorino.py, passed. So LGTM, just two questions.

Should not `PreexistingAuthorino` class be removed as well? It is no longer used:
https://github.com/Kuadrant/testsuite/blob/main/testsuite/openshift/authorino.py#L113

Here is some suspicious code that looks like containing some leftovers that could be removed? Not 100% though:
https://github.com/Kuadrant/testsuite/blob/main/testsuite/config/__init__.py#L36-L39

PR #468, opened 2024-07-01
Added disruptive make target

jsmolar:


Makefile
```diff
@@ -71,6 +71,10 @@ performance: ## Run performance tests
 performance: poetry-no-dev
 	$(PYTEST) --performance $(flags) testsuite/tests/kuadrant/authorino/performance
 
+disruptive: ## Run disruptive tests
+disruptive: poetry-no-dev
+	$(PYTEST) -m 'disruptive' --enforce $(flags) testsuite
```
pehala, line 76: Not sure how much will `--enforce` matter with this context

pehala (dismissed):

azgabur (approved):
Approved, but with question. Why does the cleanup fixture set the values `replicas` and `resourceRequirements` instead of removing the `limitador` key from kuadrant CR? Would that not apply reasonable defaults as in clean Kuadrant install?

pehala (approved):

azgabur (approved):

PR #473, opened 2024-07-04
Rename tests/kuadrant to tests/singlecluster

pehala:
Since we now have `multicluster` tests, it makes sense to rename our current `kuadrant` tests to `singlecluster`


jsmolar:
My idea is to name this `single` and `multi` respectively. 
WDYT @azgabur @averevki

azgabur:
Yeah I agree with Jakub, if we are already renaming, we can rename the modules to shorter names of `single` and `multi`. I dont think any meaning will get lost.

averevki:
> My idea is to name this `single` (tests/single) and `multi` (tests/multi) respectively.

I like `singlecluster` and `multicluster` better. If I see testsuite for the first time without any knowledge about it, I wouldn't know what it's mean

jsmolar:
`singlecluster` and `multicluster` it is. @averevki had a good point.

jsmolar (approved):

averevki (approved):

PR #483, opened 2024-07-08
Unify ApyProxy clients

pehala:
* Client is required now
   * Fixture will take care of client creation and cleaning up and initial setup (auth + verify)
* Remove verify=false from Mockserver, we use HTTP anyways   
* Small updates to fixtures that use that
* Tracing will get separate PR later with bigger changes

### DIFF
--- a/testsuite/kubernetes/metrics/prometheus.py
+++ b/testsuite/kubernetes/metrics/prometheus.py
@@ -1,12 +1,12 @@
 """Simple client for the Prometheus metrics"""
 
-from typing import Callable
 from datetime import datetime, timezone
+from typing import Callable
 
 import backoff
 from apyproxy import ApyProxy
+from httpx import Client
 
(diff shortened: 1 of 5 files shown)


averevki (dismissed):

averevki (dismissed):

trepel:
This does not work for me:
```
../../../.cache/pypoetry/virtualenvs/kuadrant-testsuite-npzTsQIr-py3.12/lib/python3.12/site-packages/apyproxy/__init__.py:78: in _
    url = os.path.join(self.__url.rstrip("/"), relpath)
E   AttributeError: 'URL' object has no attribute 'rstrip'
```

pehala:
Thanks @trepel, great find, fixed.

averevki (approved):

trepel (approved):

PR #485, opened 2024-07-09
Rework tracing

pehala:
- Create Interface
- Separate Jaeger and Tempo


trepel (approved):
Code looks good to me, tests passed (`flags="-vvv --standalone" make testsuite/tests/singlecluster/authorino/operator/tracing`) as well so I am approving this one. I assume that the tests used Jaeger by default, it would be good to try with Tempo as well but I have no idea how to do it.

PR #486, opened 2024-07-09
Remove typing annotation that were never checked

pehala:
- Mypy no longer throws warning/note
- Included unneeded reorg of imports, not sure if it matter here


trepel (approved):
LGTM, test passes and IDE shows no warnings.

averevki (approved):

PR #489, opened 2024-07-16
Fix multicluster test

averevki:


azgabur (approved):

trepel (approved):
I was lazy to get the full understanding of the test. The test worked for me both with and without this. I seems to me that the bug was related to that hard-coded `[0]` which is fixed and also the piece of code introduced in this PR is much nicer.

PR #490, opened 2024-07-22
Add tests for Defaults

### LINKED ISSUE #375: Atomic defaults tests

averevki:
### Changes

- Add the Defaults&Overrides interface to the testsuite
- Add tests for the Defaults

Closes #375


martinhesko:
Already reviewed in the previous PR, LGTM

martinhesko (approved):

averevki:
signoff commits^

martinhesko (dismissed):

testsuite/tests/singlecluster/authorino/authorization/opa/external_registry/conftest.py
```diff
@@ -24,5 +24,5 @@ def authorization(authorization, opa_policy_expectation):
     Adds OPA policy. Rego query is located on external registry (Mockserver).
     Policy accepts requests that contain `header`.
     """
-    authorization.authorization.add_external_opa_policy("opa", opa_policy_expectation, 1)
+    authorization.rules.authorization.add_external_opa_policy("opa", opa_policy_expectation, 1)
```
pehala, line 27: I would make it backwards compatible  and make `rules` default if nothing is specified. Will result in much smaller PR without any significant sideeffect

testsuite/kuadrant/policy/authorization/auth_policy.py
```diff
@@ -46,3 +45,24 @@ def add_rule(self, when: list["Rule"]):
...
+    def defaults(self):
+        """Add new rule into the `defaults` AuthPolicy section"""
+        self.spec_section = self.model.spec.setdefault("defaults", {})
+        return self
+
+    @property
+    def overrides(self):
+        """Add new rule into the `overrides` AuthPolicy section"""
+        self.spec_section = self.model.spec.setdefault("overrides", {})
+        return self
```
pehala, line 68: This is weird, while it indeed works, wont this cause weird artefacts like this?

```
authorization.identity.oidc() ->rules
authorization.overrides.identity.oidc() -> overrides
authorization.identity.oidc() -> overrides
```

averevki: spec_section resets after new item is added. I have similar setup below 
```py
authorization.defaults.identity.add_oidc("inside-defaults", oidc_provider.well_known["issuer"])
authorization.identity.add_oidc("outside-defaults", oidc_provider.well_known["issuer"])
```

pehala: That is awesome, thanks!

testsuite/tests/singlecluster/conftest.py
```diff
@@ -34,10 +34,12 @@ def authorization_name(blame):
 
 
 @pytest.fixture(scope="module")
-def authorization(kuadrant, route, authorization_name, cluster, label):
+def authorization(request, kuadrant, route, gateway, blame, cluster, label):  # pylint: disable=unused-argument
     """Authorization object (In case of Kuadrant AuthPolicy)"""
+    target_ref = request.getfixturevalue(getattr(request, "param", "route"))
```
pehala, line 39: I do not like that indirect parametrization is so high in the hierarchy. It probably wont affect anything, but I still do not like it. @jsmolar thoughts?

testsuite/tests/singlecluster/defaults/test_rules_exclusivity.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.mark.parametrize(
+    "component_fixture",
+    [
+        pytest.param("authorization", id="AuthPolicy"),
+        pytest.param("rate_limit", id="RateLimitPolicy"),
+    ],
+)
+def test_rules_exclusivity(request, component_fixture):
```
pehala, line 40: This is too confusing. I would rather haev two test cases with creation of the policies inside it as the act of creation is part of the testcase and as such should be done in test function and not in fixtures

pehala: Difference between this test and most other tests is that most other tests expect policy to exist and only test behaviour afterwards, hence creation is not important and is not part of the testfunction. This test tests the creation itself and as such is much more similar to `test_gw_doesnt_exists` and similar reconciliation tests. In these test the creation of policy itself is the tested behaviour and should be present in the test function and not in the fixture, making it hard to understand what is going on for someone who is not familiar with pytest internals.

TLDR: For this test I think it would be better to 1) simplify it 2) not use fixtures for policies and instead create them directly in the test function

averevki: Makes sense. Thanks for the explanation

testsuite/tests/singlecluster/conftest.py
jsmolar: I added exactly the same parametrization to rate limit tests and it is opaque shit. I don't know how to solve parametrization for routes and gateways, so if you have better idea I'm all ears, but if not lets go with this solution.

pehala (approved):

martinhesko (approved):

PR #492, opened 2024-07-24
Remove namespace from prometheus queries

averevki:
A few small prometheus tweaks:
- Remove namespace from prometheus queries
- Change prometheus fixture to the package scope
- Fix pytest.skip after yield (last metrics test was always skipped)


testsuite/tests/singlecluster/authorino/metrics/conftest.py
```diff
@@ -12,7 +12,7 @@
 from testsuite.prometheus import Prometheus
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="package")
```
jsmolar, line 15: did you consider using `session`?

jsmolar (approved):

azgabur (approved):

PR #493, opened 2024-07-25
Add test for limitador metrics

jsmolar:


testsuite/kubernetes/monitoring/pod_monitor.py
```diff
@@ -0,0 +1,37 @@
+"""Module implements Pod Monitor CR"""
+
+from testsuite.kubernetes import KubernetesObject
+from testsuite.kubernetes.client import KubernetesClient
+from testsuite.kubernetes.monitoring import MetricsEndpoint
+from testsuite.utils import asdict
+
+
+class PodMonitor(KubernetesObject):
```
pehala, line 9: Why do we need to use PodMonitor insteaed of ServiceMonitor? 
I would like to see at lůeast some explanation in the PR description

jsmolar: I don't have an answer, I followed documentation: https://github.com/Kuadrant/limitador/tree/main/limitador-server/kubernetes#monitoring

testsuite/tests/singlecluster/limitador/conftest.py
```diff
@@ -1,6 +1,23 @@
 """Conftest for rate limit tests"""
 
 import pytest
+from openshift_client import selector, OpenShiftPythonException
+
+from testsuite.kuadrant.limitador import LimitadorCR
+
+
+@pytest.fixture(scope="session")
+def limitador(cluster, testconfig) -> LimitadorCR:
```
pehala, line 10: You could use `kuadrant.limitador` instead

testsuite/tests/singlecluster/limitador/metrics/test_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.fixture(scope="module", autouse=True)
+def rate_limit(rate_limit):
+    """Add limit to the policy"""
+    rate_limit.add_limit("multiple", [Limit(3, 10)])
+    return rate_limit
+
+
+@pytest.fixture(scope="module", autouse=True)
+def wait_for_scrape(prometheus, pod_monitor, client):
```
pehala, line 16: Misleading name, because it also fires requests, which is not obvious

testsuite/kubernetes/monitoring/pod_monitor.py
pehala: I would prefer using either servicemonitor or podmonitor in all of testsuite

trepel (dismissed):
I am far from being an expert on monitoring but the code looks good to me (just two minor comments) and the test passed. I also validated that the tests are skipped if monitoring is not enabled.

trepel (approved):
LGTM now!

testsuite/tests/singlecluster/limitador/metrics/test_metrics.py
```diff
@@ -0,0 +1,51 @@
+"""Tests for Limitador metrics"""
+
+import pytest
+
+from testsuite.kuadrant.policy.rate_limit import Limit
+
+
+@pytest.fixture(scope="module", autouse=True)
```
averevki, line 8: autouse?

testsuite/kuadrant/__init__.py
```diff
@@ -28,6 +28,10 @@ def deployment(self):
         with self.context:
             return selector("deployment", labels={"app": self.spec_name}).object(cls=Deployment)
 
+    def name(self):
+        """Overrides `name` method from `apiobject` so it returns name of Kuadrant section"""
+        return self.spec_name
```
averevki, line 33: I wouldn't override apiobject properties. I'd use/rename spec_name if there is no other override reason

jsmolar: The idea behind this is that both `kuadrant.limitador` and `limitador` standalone (I know that we don't have it yet). Have the same interface and can be interchangeable in tests (same as authorino). So if I call `authorino.name()` I want to get the name of resource, or the name of subcomponent of kuadrant. Does this make a sense?

PR #499, opened 2024-07-28
Removed standalone-only for authornino metrics tests

jsmolar:


testsuite/kuadrant/authorino.py
```diff
@@ -22,13 +21,9 @@ class TracingOptions:
     insecure: Optional[bool] = None
 
 
-class Authorino(LifecycleObject):
+class Authorino:
     """Authorino interface"""
 
-    @abc.abstractmethod
-    def wait_for_ready(self):
```
pehala, line 29: Why? I see absolutely no reason to do so

testsuite/tests/singlecluster/authorino/metrics/conftest.py
```diff
@@ -40,33 +38,17 @@ def prometheus(cluster):
...
-    """Label Authorino controller-metrics service for the proper discovery"""
-    authorino.metrics_service.label({"app": module_label})
-    return authorino
-
-
-@pytest.fixture(scope="module")
-def service_monitor(cluster, prometheus, blame, module_label):  # pylint: disable=unused-argument
+def service_monitor(cluster, prometheus, blame, module_label, authorino, kuadrant):  # pylint: disable=unused-argument
     """Create ServiceMonitor object to follow Authorino /metrics and /server-metrics endpoints"""
+    if kuadrant:
```
pehala, line 43: This should be incorporated into the Authorino classes somehow, no more ifs

testsuite/tests/singlecluster/authorino/metrics/conftest.py
```diff
@@ -40,33 +38,17 @@ def prometheus(cluster):
...
+        label = {"app": "service_monitor_label"}
```
pehala, line 44: sidenote: if we are also in charge of choosing the label (I suspect we are since this is not a default one), why not use module label as well?

pehala (changes requested):

testsuite/tests/singlecluster/authorino/metrics/conftest.py
jsmolar: Two parallel runs of the same test would be disruptive (two service monitors would target two different labels of the same service). Currently, that is not a problem but if we add more tests (which is not probable rn), different label would make this disruptive.
What do you think, should I use module label?

pehala: And now they would not be?

testsuite/kuadrant/authorino.py
```diff
@@ -133,9 +132,3 @@ def authorization_url(self):
     @property
     def oidc_url(self):
         return self._oidc_url
-
-    def commit(self):
```
pehala, line 137: If you remove this here, remove it from PreexistingAuthorino as well

jsmolar: This is removed from PreexistingAuthorino

pehala (approved):

azgabur (approved):

PR #502, opened 2024-07-31
Add another gateway CR name for cleaning

azgabur:
So we clean the Gateways.
Not sure when this changed but keeping two names for gateways should not be a problem anyways.


pehala (approved):

averevki (approved):
I'm not sure if we ever used `gateways.networking.istio.io` Gateway

Issue #308, opened 2023-12-14
Add test for adding and removing listeners to/from Gateway when DNSPolicy and TLSPolicy are attached


ficap:
**Summary**
When adding or removing listeners to/from  a Gateway (possibly with various hostnames) with attached DNSPolicy and/or TLSPolicy, certs/dns records should be updated accordingly.

**Documentation**

Link any relevant product documentation.

**Test cases**

Split test case into sub-tasks if needed. It should help with tracking progress and navigating more complex tasks.

ficap:
related bug: https://github.com/Kuadrant/multicluster-gateway-controller/issues/601

azgabur:
New issue for the related bug: https://github.com/Kuadrant/kuadrant-operator/issues/794

PR #504, opened 2024-08-09
Add Gateway listener dns test

azgabur:
This PR addes new test that checks for this bug: https://github.com/Kuadrant/kuadrant-operator/issues/794
Part of work on https://github.com/Kuadrant/testsuite/issues/308

### DIFF
--- a/testsuite/tests/singlecluster/gateway/reconciliation/test_gateway_listeners_dns.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/test_gateway_listeners_dns.py
@@ -1,0 +1,77 @@
+"""Testing specific bug that happens when listener hostname in Gateway gets changed while DNSPolicy is applied."""
+
+import pytest
+from testsuite.gateway.gateway_api.hostname import StaticHostname
+from testsuite.utils import is_nxdomain, sleep_ttl
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy, pytest.mark.tlspolicy]
+
+
+@pytest.fixture(scope="module")
+def wildcard_domain(base_domain, blame):
+    """
+    For this test we want specific domain, not wildcard.
+    This will be used in the first iteration of Gateway and HTTPRoute.
+    """
+    return f"{blame('dnsbug1')}.{base_domain}"
+
+
+@pytest.fixture(scope="module")
+def new_domain(base_domain, blame):
+    """In the test the Gateway and HTTPRoute will change their hostnames to this one."""
+    return f"{blame('dnsbug2')}.{base_domain}"
+
+
+@pytest.fixture(scope="module")
+def client(gateway, wildcard_domain):
+    """Make client point to correct domain route domain."""
+    return StaticHostname(wildcard_domain, gateway.get_tls_cert).client()
+
+
+@pytest.fixture(scope="module")
+def client_new(gateway, new_domain):
+    """Second client that will be created after Gateway gets updated pointing to new_domain."""
+
+    def _client_new():
+        return StaticHostname(new_domain, gateway.get_tls_cert).client()
+
+    return _client_new
+
+
+@pytest.fixture(scope="module")
+def route(route, wildcard_domain):
+    """So that route hostname matches the gateway hostname."""
+    route.remove_all_hostnames()
+    route.add_hostname(wildcard_domain)
+    return route
+
+
+@pytest.mark.xfail()
+@pytest.mark.issue("https://github.com/Kuadrant/kuadrant-operator/issues/794")
+def test_change_hostname(client, client_new, auth, gateway, route, new_domain, wildcard_domain):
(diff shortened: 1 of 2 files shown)


testsuite/utils.py
```diff
@@ -176,3 +178,24 @@ def check_condition(condition, condition_type, status, reason=None, message=None
     ):
         return True
     return False
+
+
+def is_nxdomain(hostname: str):
+    """Returns bool value if hostname has `A` record in DNS."""
```
trepel, line 184: My understanding of this function is that it returns `True` if `hostname` cannot be resolved to an IP address. `False` otherwise. But that comment suggests something very different. Please change that comment (if it is wrong) or please explain to me how come that the comment is correct.

azgabur: You are correct the doc string is wrong. I edited it and added useful information.

testsuite/tests/singlecluster/gateway/reconciliation/test_gateway_listeners_dns.py
```diff
@@ -0,0 +1,77 @@
...
+from testsuite.gateway.gateway_api.hostname import StaticHostname
+from testsuite.utils import is_nxdomain, sleep_ttl
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy, pytest.mark.tlspolicy]
+
+
+@pytest.fixture(scope="module")
+def wildcard_domain(base_domain, blame):
+    """
+    For this test we want specific domain, not wildcard.
```
averevki, line 13: It shouldn't be called wildcard_domain then. domain/gateway_domain or whatever else you like

testsuite/utils.py
```diff
@@ -176,3 +178,26 @@ def check_condition(condition, condition_type, status, reason=None, message=None
...
+        return True
+    return False
+
+
+def sleep_ttl(hostname: str):
+    """Sleeps for duration of TTL for given hostname."""
+    try:
+        response = dns.resolver.resolve(hostname, raise_on_no_answer=False)
+    except dns.resolver.NXDOMAIN:
+        return
```
averevki, line 200: reuse `is_nxdomain()` please

testsuite/utils.py
```diff
@@ -176,3 +178,26 @@ def check_condition(condition, condition_type, status, reason=None, message=None
...
+    if response.rrset is None:
+        return
```
averevki, line 202: in what case rrset is none?

averevki (commented):
I couldn't get dnspolicy to the ready state before the test `DNSPolicy has encountered some issues: reconcile DNSRecords error error reconciling dns records for gateway gw-averevki--g81o: DNSRecord.kuadrant.io "gw-averevki--g81o-api" is invalid: [spec.providerRef: Required value, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]`

testsuite/utils.py
trepel: It makes sense now, thanks!

testsuite/utils.py
azgabur: When DNS server returns empty answer but no error. This usually happens when under the domain name there is different type of record than what was requested in the query. The default behavour of `dns.resolver.resolve` is to query for A record. 

This functions is here mostly for MyPy check failing. I will just add type ignore comment here and remove the `raise_on_no_answer` so if the RRset is empty, exception is thrown.

testsuite/tests/singlecluster/gateway/reconciliation/test_gateway_listeners_dns.py
azgabur: That is the name of fixture the gateway fixture uses for the hostname. Alternative would be:
- Redefine the whole gateway fixture in the test with the hostname
- Rename the `wildcard_domain` to `gateway_domain` and check this is not conflicting name in different context
- Change the domain with new listener editing functions which I will create as part of work on the issue and move gateway commit to commit fixture to allow for editing before commit

What do you think is best option?

averevki: The second one. @jsmolar thoughts?

trepel: The truth is that the name is confusing. However, if renaming leads to some non-trivial refactoring I would vote for a separate PR for that work.

azgabur: I agree, we can solve this as another PR.

averevki (approved):

trepel (approved):
If I got it right this test covers what it is expected to cover. I tried to pause the test just before last assertion and removed the dnsbug1 records from AWS manually and the test passed then as expected. LGTM

PR #511, opened 2024-08-14
Add Kuadrant version to pytest report

azgabur:
This change will add image strings of deployed Kuadrants on clusters (including second cluster if defined) in pytest report header when running testsuite with make targets. This string is taken from CatalogSource object and in case the object is not found, no string will be printed. In the future this can be expanded for Kind setup, or doing smarter job in finding the right CatalogSource.

This can make debugging easier and is mostly helpful for figuring out nightly build version as the version string in Kuadrant subscription object is "0.0.0".

### DIFF
--- a/testsuite/capabilities.py
+++ b/testsuite/capabilities.py
@@ -24,3 +24,22 @@ def has_kuadrant():
                 return False, f"Cluster {cluster.api_url} does not have Kuadrant resource in project {project}"
 
     return True, None
+
+
+@functools.cache
+def kuadrant_version():
+    """Returns catalog image tag of deployed Kuadrant if possible."""
+    clusters = [settings["control_plane"]["cluster"]]
+    if cluster2 := settings["control_plane"]["cluster2"]:
+        clusters.append(cluster2)
+    versions = []
+    for cluster in clusters:
+        project = cluster.change_project("openshift-marketplace")
+        if not project.connected:
+            break
+        with project.context:
+            catalog_source = selector("CatalogSource/kuadrant-upstream").object(ignore_not_found=True)
+            if catalog_source is None:
+                break
+            versions.append((catalog_source.as_dict()["spec"]["image"], cluster.api_url))
+    return versions
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -4,10 +4,11 @@ import signal
 from urllib.parse import urlparse
 
 import pytest
+from pytest_metadata.plugin import metadata_key  # type: ignore
 from dynaconf import ValidationError
 from keycloak import KeycloakAuthenticationError
 
-from testsuite.capabilities import has_kuadrant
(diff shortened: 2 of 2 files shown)


averevki (commented):
Will the image show up on report portal?

azgabur:
I tried adding this fixture which adds property info to junit file but report portal is not displaying it. Maybe in the future when report portal starts displaying this additional info we can add this, otherwise I did not find any other simple way.
```py
@pytest.fixture(scope="session", autouse=True)
def log_global_env_facts(record_testsuite_property):
    """Adds Kuadrant version in property of testcase in junit."""
    for i, image in enumerate(kuadrant_version()):
        record_testsuite_property(f"KUADRANT_{i}", image[0])
```

azgabur:
^ I added just the image info to metadata data structure which will be displayed in "Environment" section when generating html report.

trepel:
@azgabur I tried using `html=1 make ./testsuite/tests/singlecluster/gateway/test_basic.py` and `junit=1 make ./testsuite/tests/singlecluster/gateway/test_basic.py`. Former worked (even tried with additional clusters), latter did not work for some reason. Any ideas what might have gone wrong?

azgabur:
> `junit=1 make ./testsuite/tests/singlecluster/gateway/test_basic.py`

In the end I did not implement the junit property carrying the kuadrant image string as the Report portal did not display it so thats why its not included. But just by adding this function (https://github.com/Kuadrant/testsuite/pull/511#issuecomment-2290971355) to top level conftest it can be later added if needed.

trepel (approved):

testsuite/capabilities.py
```diff
@@ -25,3 +25,21 @@ def has_kuadrant():
...
+@functools.cache
+def kuadrant_version():
+    """Returns catalog image tag of deployed Kuadrant if possible."""
+    clusters = weakget(settings)["control_plane"]["additional_clusters"] % []
+    clusters.insert(0, settings["cluster"])
+    versions = []
+    for cluster in clusters:
+        project = cluster.change_project("openshift-marketplace")
+        if not project.connected:
+            break
```
jsmolar, line 39: Why break when a catalog or project is not found? For consistency's sake I would expect that we have relevant information in all reports.

azgabur: When the catalog source under exact name `CatalogSource/kuadrant-upstream` inside `openshift-marketplace` namespace is not found than no version is printed. This is intended. I plan in the future to make the search for version better to work in more circumstances - mainly on Kind. 

This feature should be taken just as convince not as production ready and documented feature of the testsuite.

averevki (approved):

trepel (approved):
Approving again, it works for HTML report.

testsuite/capabilities.py
jsmolar: Please create an issue for this. Otherwise we forgot about it

PR #513, opened 2024-08-16
Log tls policy model when wait for ready fails

jsmolar:


testsuite/kuadrant/policy/tls.py
```diff
@@ -52,4 +52,4 @@ def __getitem__(self, key):
     def wait_for_ready(self):
         """Increase timeout to account for letsEncrypt"""
         success = self.wait_until(has_condition("Enforced", "True"), timelimit=180)
-        assert success, f"{self.kind()} did not get ready in time"
+        assert success, f"{self.kind()} did not get ready in time:\n {self.refresh().model}"
```
azgabur, line 55: Its not just the "model.status" part of the TLSPolicy useful in this situation?

trepel: Agree with @azgabur here, `.model.status` is sufficient in this situation. My question is why for TLSPolicy only. This would be useful to have for each policy, so I would add this to the three assertions here as well: https://github.com/Kuadrant/testsuite/blob/main/testsuite/kuadrant/policy/__init__.py#L29-L43

wdyt?

PR #516, opened 2024-08-20
Add geo load-balancing tests

### LINKED ISSUE #496: Add test for GEO load balancing to multicluster tests
**Summary**

Add test for GEO load balancing to multicluster tests

**Documentation**

- https://docs.kuadrant.io/0.8.0/kuadrant-operator/doc/reference/dnspolicy/
- https://docs.kuadrant.io/0.8.0/multicluster-gateway-controller/docs/how-to/multicluster-loadbalanced-dnspolicy/#geo

**Additional context**

It is multicluster test, so at least two clusters are required. Start with DNS on AWS (Route 53). Use some US and EU geolocations for the test. This test (likely) requires to spin up clusters in AWS in some us an eu region so it might be simpler to wait for https://github.com/Kuadrant/testsuite-pipelines/issues/7 to be finished first (but not technically a blocker, one can spin up the clusters manually).

On each cluster
- 1 Gateway
- 1 DNSPolicy for each gateway with load balanced strategy
- Default GEO set to the same in each policy
- Valid but different Geo annotation added to each gateway
- No additional weighting config 

Success Criteria
- All policies get to ready state
- Gateways should have enforced and affected status
- Validate on cloud provider level that the DNS records were created as expected. This might be simpler/more stable than generating requests from various GEOs and validating that the proper cluster handled them. However, we can potentially create a pod or `oc rsh...` inside existing pod on each cluster and make request from there.

Follow ups

As part of working on this one can also implement (or at least create a follow up issues for):
- Ability to test other cloud providers
- Weighted GEO: same as above but set the same GEO on more than one Gateway and set a custom weight annotation on these clusters. This would require having more than 1 Gateway on each cluster. Validate that both addresses are returned with roughly the expected ratio.

### LINKED ISSUE #515: Test for unsupported GEO Code in DNSPolicy

averevki:
### Changes

- Change the list of additional clusters to only 2 (not required in the near feature)
- Reformat multicluster contest to support only 2 clusters

### Tests

- Add geo load balancing test
- Add unsupported geocode test

Closes #496, closes #515

### DIFF
--- a/testsuite/tests/multicluster/load_balanced/test_unsupported_geocode.py
+++ b/testsuite/tests/multicluster/load_balanced/test_unsupported_geocode.py
@@ -1,0 +1,15 @@
+"""Test not supported geocode in geo load-balancing"""
+
+import pytest
+
+from testsuite.kuadrant.policy import has_condition
+
+pytestmark = [pytest.mark.multicluster]
+
+
+def test_unsupported_geocode(dns_policy):
+    """Change default geocode to not existent one and verify that policy became not enforced"""
+    dns_policy.model.spec.loadBalancing.geo.defaultGeo = "XX"
+    dns_policy.apply()
+
+    assert dns_policy.wait_until(has_condition("Enforced", "False"))
--- a/config/settings.local.yaml.tpl
+++ b/config/settings.local.yaml.tpl
@@ -1,9 +1,5 @@
 #default:
 #  tester: "someuser"                          # Optional: name of the user, who is running the tests, defaults to whoami/uid
-#  cluster:                                    # Primary cluster where tests should run
-#    api_url: "https://api.kubernetes.com"     # Optional: Kubernetes API URL, if None it will use Kubernetes that you are logged in
-#    token: "KUADRANT_RULEZ"                   # Optional: Kubernetes Token, if None it will Kubernetes that you are logged in
-#    kubeconfig_path: "~/.kube/config"         # Optional: Kubeconfig to use, if None the default one is used
 #  kuadrantctl: kuadrantctl
 #  tools:
 #    project: "tools"                          # Optional: Kubernetes project, where external tools are located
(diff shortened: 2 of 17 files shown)


trepel:
@averevki Could you please resolve the conflicts? It would make reviewing easier and you need to that anyway

testsuite/tests/multicluster/load_balanced/test_unsupported_geocode.py
```diff
@@ -0,0 +1,15 @@
...
+
+pytestmark = [pytest.mark.multicluster]
+
+
+def test_unsupported_geocode(dns_policy):
+    """Change default geocode to not existent one and verify that policy became not enforced"""
+    dns_policy.model.spec.loadBalancing.geo.defaultGeo = "XX"
+    dns_policy.apply()
+
+    assert dns_policy.wait_until(has_condition("Enforced", "False"))
```
trepel, line 15: I think that it would be cool if more specific assert is here. Once your https://github.com/Kuadrant/testsuite/pull/520 is merged you can use great approach you used there. Something along the lines of
```
assert dns_policy.wait_until(
        has_record_condition("Ready", "False", "ProviderError", "Cannot find location")
    ), f"DNSPolicy did not reach expected record status, instead it was: {dns_policy.model.status.recordConditions}"
```

trepel:
I don't understand the reasoning behind doing this refactoring to `Change the list of additional clusters to only 2`. Why was it required?

trepel:
I am seeing `The DNS operation timed out.` error while running `resolver.resolve(hostname.hostname)`. Did you come across that?

averevki:
> I don't understand the reasoning behind doing this refactoring to `Change the list of additional clusters to only 2`. Why was it required?

We don't need this functionality for now. Right now, it's committing the same object on each cluster, and I need more flexibility in CR configuration on each cluster

trepel (dismissed):
I tried each test modified in this PR, all good. I tried to modify the geo label to ES and I tried to remove it completely, the geo loadbalancing test failed as expected then. Code looks good to me, my understanding of it is that instead of crafting a request from certain GEO (DE or AU) or looking at DNS records in AWS  the test specifically asks DE and AU DNS nameserver to resolve the same hostname and validates that external IP of right Gateway is returned. Nice!

trepel:
Only thing that confuses me a little is that if I tried to resolve the hostname in https://dnschecker.org/ it took a few minutes till at least one AU server started resolving to "AU" Gateway external IP.

jsmolar (dismissed):

jsmolar (approved):

Issue #427, opened 2024-06-06
Stabilize test_preexisting_auth


pehala:
**Summary**
#421 did not fully fixed the issue, it still fails on nightlies

trepel:
Not sure if this helps but I created a PR adding wait for Gateway to get ready:
https://github.com/Kuadrant/testsuite/pull/523

Not sure what to do next if this does not help. Apart from adding random sleeps :/

PR #523, opened 2024-09-04
Wait for Gateway to be ready before using it

trepel:
## Overview

This is a small try to further stabilze the test_preexisting_auth, see 
https://github.com/Kuadrant/testsuite/issues/427

## Verification Steps
Execute

`flags="-vvv --standalone" make testsuite/tests/singlecluster/authorino/operator/sharding/test_preexisting_auth.py`

and

`flags="-vvv --standalone" make testsuite/tests/singlecluster/authorino/operator/sharding/test_sharding.py`

Both should pass. We will see in nightlies whether this actually helped or not.

### DIFF
--- a/testsuite/tests/singlecluster/authorino/operator/sharding/conftest.py
+++ b/testsuite/tests/singlecluster/authorino/operator/sharding/conftest.py
@@ -22,6 +22,7 @@ def setup_gateway(request, cluster, blame, testconfig, module_label):
         )
         request.addfinalizer(gw.delete)
         gw.commit()
+        gw.wait_for_ready()
         return gw
 
     return _envoy


averevki (approved):

azgabur (approved):
I tried to verify with running more tests concurrently as this is probably the race condition that triggers this bug

```
flags=-n20 make authorino-standalone
```

Sadly the test_preexisting_auth still occasionally fails so the problem must be somewhere deeper. I remember from talks with @pehala that this test is fundamentally wrong so maybe we need to look at the test case itself to determine if it is something we even need to test or if it has really been a bug in Authorino this whole time.

trepel:
@azgabur Thanks for review! I think that this change makes logical sense anyway so I will merge this one. About the test itself please add what you know/recall that might help to https://github.com/Kuadrant/testsuite/issues/427

PR #524, opened 2024-09-05
Remove pytest-asyncio lib

azgabur:
We never used this library and currently its printing many warnings in test report so its good time to remove it.

For verification I would just run the testsuite and see no errors or text search for `asyncio` string in code and see no results.


averevki (approved):

trepel (approved):
Approve. No mentions of asyncio in the test suite. Haven't run the whole testsuite, just one single test, however, nightly pipelines will tell us if there is an issue with this soon

PR #529, opened 2024-09-10
Add gateway listeners scaling test

### LINKED ISSUE #528: DNS Scale test (gateway with 64 listeners)
Gateways can have up to 64 listeners. Test that we can handle this many listeners with distinct domains

- Create multiple gateways with 64 listeners each
- Create DNSPolicies for the gateways
- Verify that DNSPolicies are enforced and gateway listeners are reachable

averevki:
**Changes**
- Overwrite `wait_for_ready` for DNSPolicy with timeout increased to 5 min instead of 1 min, large amount of DNS records take some time to create
- Create 2 gateways with max listeners and see if both dns policies are enforced and some random listeners are reachable

Closes #528

### DIFF
--- a/testsuite/kuadrant/policy/dns.py
+++ b/testsuite/kuadrant/policy/dns.py
@@ -67,3 +67,7 @@ class DNSPolicy(Policy):
             model["spec"]["loadBalancing"] = asdict(load_balancing)
 
         return cls(model, context=cluster.context)
+
+    def wait_for_full_enforced(self, timelimit=300):
+        """Wait for a Policy to be fully Enforced with increased timelimit for DNSPolicy"""
+        super().wait_for_full_enforced(timelimit=timelimit)
--- a/testsuite/tests/singlecluster/gateway/test_scale_listeners.py
+++ b/testsuite/tests/singlecluster/gateway/test_scale_listeners.py
@@ -1,0 +1,76 @@
+"""Create 2 gateways with 64 listeners each and confirm they are reachable with DNSPolicy applied"""
+
+import pytest
+
+from testsuite.httpx import KuadrantClient
+from testsuite.gateway.gateway_api.route import HTTPRoute
+from testsuite.gateway.gateway_api.gateway import KuadrantGateway
+from testsuite.kuadrant.policy.dns import DNSPolicy
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
+
+MAX_GATEWAY_LISTENERS = 64
+
+
+@pytest.fixture(scope="module")
+def gateway(request, cluster, blame, base_domain, module_label):
+    """Create first gateway with 64 listeners"""
+    gw = KuadrantGateway.create_instance(cluster, blame("gw"), f"gw1-api.{base_domain}", {"app": module_label})
+    for i in range(1, MAX_GATEWAY_LISTENERS):
+        gw.add_listener(f"api{i}", f"gw1-api{i}.{base_domain}")
+    request.addfinalizer(gw.delete)
+    gw.commit()
+    gw.wait_for_ready()
+    return gw
+
+
+@pytest.fixture(scope="module")
+def gateway2(request, cluster, blame, base_domain, module_label):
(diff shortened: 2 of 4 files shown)


testsuite/kuadrant/policy/dns.py
```diff
@@ -31,3 +31,10 @@ def create_instance(
         }
 
         return cls(model, context=cluster.context)
+
+    def wait_for_full_enforced(self):
```
jsmolar, line 35: Instead of rewriting the `wait_for_full_enforce` method, why don't we add a parameter to the parent method for timeout:  `Policy.wait_for_full_enforce(self, timeout=60)`?

trepel (approved):
I am approving this one. However, it failed for me twice out of ~8 times so the 300s `timelimit` might need to be increased. We can adjust based on flakiness we observe in nightly pipelines.

testsuite/tests/singlecluster/gateway/test_scale_listeners.py
```diff
@@ -0,0 +1,76 @@
...
+def test_gateway_max_listeners(gateway, gateway2, dns_policy, dns_policy2, base_domain):
+    """Verify that both gateways are affected by DNSPolicy and their listeners are reachable"""
+    assert gateway.refresh().is_affected_by(dns_policy)
+    assert gateway2.refresh().is_affected_by(dns_policy2)
+
+    assert KuadrantClient(base_url=f"http://gw1-api.{base_domain}").get("/get").response.status_code == 200
+    assert KuadrantClient(base_url=f"http://gw1-api63.{base_domain}").get("/get").response.status_code == 200
+
+    assert KuadrantClient(base_url=f"http://gw2-api21.{base_domain}").get("/get").response.status_code == 200
+    assert KuadrantClient(base_url=f"http://gw2-api53.{base_domain}").get("/get").response.status_code == 200
```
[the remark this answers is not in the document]
averevki: first and last listener in the first gateway and random ones in the second, so yes. I think those ones should be enough to verify the gateways scaling

averevki:
> I am approving this one. However, it failed for me twice out of ~8 times so the 300s `timelimit` might need to be increased. We can adjust based on flakiness we observe in nightly pipelines.

I expect some problems with this test due to a large amount of DNS records getting created/deleted. I'm too getting timeouts sometimes when I'm trying to run this test again right after the execution. I hope nightlies will handle a single run of this test in parallel with other ones

testsuite/tests/singlecluster/gateway/test_scale_listeners.py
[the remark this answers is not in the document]
trepel: Ack, thx!

azgabur (dismissed):

testsuite/kuadrant/policy/dns.py
averevki: I think timelimit should be global. It makes sense to me to have different timelimits for different policies, specifically for DNS where creation of a large amount of dnsrecord takes some time @jsmolar

averevki:
rebase^

testsuite/kuadrant/policy/dns.py
trepel: Yes, in my head the best approach is to have default timelimit that is ok for all policies in vast majority of tests and for a few tests where for certain policy longer timelimit is required (eg. dnspolicy wait till lot of dnsrecords are created or tlspolicy wait till Lets Encrypt issues a cert) the longer timelimit is explicitly set -  however, I would personally avoid overriding when implementing this if possible but that's probably just a matter of taste

trepel (dismissed):

trepel:
> * Overwrite `wait_for_ready` for DNSPolicy with timeout increased to 5 min instead of 1 min, large amount of DNS records take some time to create

LGTM, I just can't see any `wait_for_ready` being overridden so I am a bit confused. You probably mean that `wait_for_full_enforced` were overridden (which is used in `wait_for_ready`)

trepel (approved):

jsmolar (approved):

trepel (approved):

Issue #531, opened 2024-09-11
DNSPolicy modification tests


averevki:
**Test-cases**
- [ ] Change strategy
- [x] ~Change default weighting~ (no easy way to test, manual testing/mark as a risk)
- [ ] Change default GEO
- [x] Delete DNSPolicy

PR #532, opened 2024-09-11
Add dnspolicy deletion test

averevki:
Part of the https://github.com/Kuadrant/testsuite/issues/531

### DIFF
--- a/testsuite/tests/singlecluster/gateway/dnspolicy/test_dnspolicy_removal.py
+++ b/testsuite/tests/singlecluster/gateway/dnspolicy/test_dnspolicy_removal.py
@@ -1,0 +1,46 @@
+"""Test that Gateway will behave properly after attached DNSPolicy is deleted"""
+
+from time import sleep
+
+import pytest
+
+from testsuite.gateway.gateway_api.gateway import KuadrantGateway
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
+
+
+@pytest.fixture(scope="module")
+def gateway(request, cluster, blame, wildcard_domain, module_label):
+    """Create gateway without TLS enabled"""
+    gw = KuadrantGateway.create_instance(cluster, blame("gw"), wildcard_domain, {"app": module_label}, tls=False)
+    request.addfinalizer(gw.delete)
+    gw.commit()
+    gw.wait_for_ready()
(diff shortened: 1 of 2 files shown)


testsuite/tests/singlecluster/gateway/dnspolicy/test_dnspolicy_removal.py
```diff
@@ -0,0 +1,43 @@
...
+    - Delete attached DNSPolicy
+    - Verify that Gateway is no longer affected by DNSPolicy and requests are failing
+    """
+    assert gateway.refresh().is_affected_by(dns_policy)
+    response = client.get("/get")
+    assert response.status_code == 200
+
+    dns_policy.delete()
+
+    assert not gateway.refresh().is_affected_by(dns_policy)
```
trepel, line 41: This assertion is not sufficient. I am getting an error:
```
        assert not gateway.refresh().is_affected_by(dns_policy)
        response = client.get("/get")
>       assert response.has_dns_error()
E       assert False
E        +  where False = has_dns_error()
E        +    where has_dns_error = <testsuite.httpx.Result object at 0x7f54b8b28320>.has_dns_error
```

It takes some time for DNS record in cloud provider (I was using AWS) to get removed and the test is too quick so there is no DNS error there. The test works like a charm when debugging or if there is a short sleep added. Apparently `dns_policy.delete()` does not wait for related DNSRecord CR to be deleted (and for related DNS record in cloud provider to be deleted)

trepel (changes requested):

testsuite/tests/singlecluster/gateway/dnspolicy/test_dnspolicy_removal.py
trepel: Not sure if `sleep_ttl` is stable enough, see https://github.com/Kuadrant/testsuite/pull/535#discussion_r1758510522 - but I might be completely wrong on that.

trepel: It is stable now so I approved this one, but I think that it would be better to use the same approach as @azgabur used in his PR (see the link to PR 535 comment above).

trepel (approved):

jsmolar (approved):

PR #533, opened 2024-09-11
Update default settings

averevki:
Update default settings due to the introduced change in https://github.com/Kuadrant/testsuite/pull/516

### DIFF
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -1,6 +1,5 @@
 default:
   dynaconf_merge: true
-  cluster: {}
   kuadrantctl: "kuadrantctl"
   tools:
     project: "tools"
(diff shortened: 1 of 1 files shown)


azgabur (approved):

trepel (approved):
LGTM, but why not to add `cluster2: {}` as well?

azgabur (changes requested):
Please add reasonable defaults for `dns` category as well

averevki:
> Please add reasonable defaults for `dns` category as well

Only defaults I can set for DNS settings are public DNS servers, which would fail on the RH network we all use most of the time. There is an example with public servers in .tpl settings. I don't think we can set anything general for this test, let me know if you have something in mind

azgabur (approved):

azgabur:
If geo tests will be skipped, then no default `dns` is needed
https://github.com/Kuadrant/testsuite/pull/537

PR #534, opened 2024-09-11
Assert IPs returned in round-robin fashion in the multicluster test

### LINKED ISSUE #495: Add test for basic load balancing (Round Robin) to multicluster tests
**Summary**

Add test for basic load balancing (Round Robin) to multicluster tests.

**Documentation**

- https://docs.kuadrant.io/0.8.0/kuadrant-operator/doc/reference/dnspolicy/
- https://docs.kuadrant.io/0.8.0/multicluster-gateway-controller/docs/how-to/multicluster-loadbalanced-dnspolicy/#weighted

**Additional context**

It is multicluster test, so at least two clusters are required. Start with DNS on AWS (Route 53).

On each cluster
- 1 Gateway
- 1 DNSPolicy for each gateway with load balanced strategy 
- Default GEO set to the same in each policy 
- No custom weighting 

Success Criteria
- All policies get to ready state
- Gateways should have enforced and affected status
- Validate on cloud provider level that the DNS records were created as expected. This might be simpler/more stable than generating requests and validating that both clusters are handling them. It takes time for DNS to propagate changes so it won't be as simple as generating 2 requests and asserting than one was handled by cluster1 and other by cluster2. It might take minutes for DNS to swap clusters.
- only one IP address should be returned, not many (as is the case for simple routing strategy, see the test here: https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/multicluster/test_multicluster_dns.py) 

**Follow ups**

As part of working on this one can also implement (or create a follow up issues for):
- 10 Gateways on each cluster instead of 1, each Gateway should have at one shared host and one non-shared host
- Test to timeout after 15 minutes if policies still not ready
- Ability to test other cloud providers

averevki:
DNS resolution in the nightly pipeline should now work the same as your local dns resolver. Therefore assert for round robin strategy should also work in nightlies

Closes #495

### DIFF
--- a/testsuite/tests/multicluster/test_simple_strategy.py
+++ b/testsuite/tests/multicluster/test_simple_strategy.py
@@ -24,6 +24,10 @@ def test_simple_strategy(client, hostname, gateway, gateway2):
     assert not result.has_cert_verify_error(), result.error
     assert result.status_code == 200
 
-    ips = {gateway.external_ip().split(":")[0], gateway2.external_ip().split(":")[0]}
-    dns_ips = {ip.address for ip in dns.resolver.resolve(hostname.hostname)}
-    assert ips == dns_ips, f"Expected IPs and actual IP mismatch, got {dns_ips}, expected {ips}"
+    dns_ip1 = dns.resolver.resolve(hostname.hostname)[0].address
+    dns_ip2 = dns.resolver.resolve(hostname.hostname)[0].address
+    assert dns_ip1 != dns_ip2, "Simple routing strategy should return IPs in a round-robin fashion"
+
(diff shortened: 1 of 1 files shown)


trepel (approved):
Nice! I would maybe try two Robin rounds but I am probably too suspicious and it is overkill.

```
    dns_ip1 = dns.resolver.resolve(hostname.hostname)[0].address
    dns_ip2 = dns.resolver.resolve(hostname.hostname)[0].address
    dns_ip3 = dns.resolver.resolve(hostname.hostname)[0].address
    dns_ip4 = dns.resolver.resolve(hostname.hostname)[0].address
    assert dns_ip1 != dns_ip2, "Simple routing strategy should return IPs in a round-robin fashion"
    assert dns_ip1 == dns_ip3
    assert dns_ip2 == dns_ip4
```

PR #535, opened 2024-09-12
Add Gateway listeners support and new test

azgabur:
Closes https://github.com/Kuadrant/testsuite/issues/308

- Refactors test from previous PR https://github.com/Kuadrant/testsuite/pull/504 to add support for new listeners gateway functions and e2e check for domain missing error.

- Add new test for adding/removing listeners from Gateway
- These current PR's will need to be rebased to use new listeners support @averevki (or I will rebase this PR if these PR's will get merged first)
   - https://github.com/Kuadrant/testsuite/pull/532
   - https://github.com/Kuadrant/testsuite/pull/520


trepel (changes requested):

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Remove second domain
+    route.refresh().remove_hostname(second_domain)
+    route.apply()
+    gateway.refresh().remove_listener(second_domain)
+    gateway.apply()
+
+    # Check the default domain still works and second domain does not exist anymore
+    check_ok_https(wildcard_domain)
+    sleep_ttl(second_domain)
```
trepel, line 50: Is this stable enough? If I got it right the `second_domain` should not exist at this point so there might be no wait at all. I don't really understand the internals of DNS discovery and propagation etc, but I think that it can happen that one DNS resolver does not resolve the domain anymore but some other resolver still might be able to resolve it because it takes some time for deletion to propagate.

When I tried this test I came across folowing error several times:
```
testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py:52: in test_listeners
    assert custom_client(second_domain).get("/get").has_dns_error()
E   AssertionError: assert False
E    +  where False = has_dns_error()
```
This happend when running via `make testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py`. If running via IDE (with or without debugging on) it always went ok.

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_listeners_dns.py
```diff
@@ -0,0 +1,42 @@
...
+def test_change_listener(custom_client, check_ok_https, gateway, route, second_domain, wildcard_domain):
+    """
+    This test checks if after change of listener hostname in a Gateway while having DNSPolicy applied, that
+    the old hostname gets deleted from DNS provider. After editing the hostname in HTTPRoute to the new value
+    this test checks the reconciliation of such procedure.
+
+    WARNING
+        Running this test in unpatched Kuadrant will leave orphaned DNS records in DNS provider if the test fails
+        on the last assert. If you want to delete the records you need to do it manually.
+        The orphaned DNS records will contain string 'dnsbug'
```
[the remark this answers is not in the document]
azgabur: Its true the that the bug was fixed but I want to keep this warning in the test if the bug would ever be reintroduced to inform the tester of the orphaned dns records.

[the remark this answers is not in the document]
trepel: I think that it does not make sense to have such warnings in general because a lot of bugs can potentially be (re)introduced and our code base would be full of such warnings quickly. However, I don't mind keeping this one there, I would maybe rephrase a little. It is easy to understand this as if the bug is still not fixed which is not true. Up to you, I approve this even with current wording.

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
azgabur: I encountered this problem before when running this test from Openshift pod. I am aware that in some circumstances this wait is not enough due to TTL caching of intermediate stub DNS resolvers.
Imagine scenario:
Original TTL of the record in authoritative DNS is 60
First caching DNS resolver queries the Authoritative DNS and stores the record and starts counting 60s
Second caching DNS resolver has a policy that caps TTL to 30 (This is a case on Openshift clusters) and queries First DNS server and starts counting 30
Finally testsuite queries for sleep_ttl() function the Second server and sleeps for 30s.

Now. After testsuite finishes sleeping, second DNS server has expired cache, authoritative DNS server does not contain the record as it was deleted, but first caching DNS server STILL HAS the record in cache! So testsuite now wants to check for dns error so it queries again, second server queries the first and gets an answer with ttl <30 but still valid. This results in the the failed assert. 

There are two ways I found out that can solve this problem. Use stub DNS resolvers that do not have policy capping the TTL or query the Authoritative DNS server for the origin TTL and use that value for sleep_ttl(). In our pipelines that run on Openshift we opted for first option. Implementing the second option can be difficult as some internal networks block port 53 so contacting authoritative DNS server is impossible from the testsuite.

**I will investigate this further.**

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
```diff
@@ -0,0 +1,52 @@
...
+    # Check the default domain works and second domain does not exist yet
+    check_ok_https(wildcard_domain)
+    assert is_nxdomain(second_domain)
+    assert custom_client(second_domain).get("/get").has_dns_error()
+
+    # Add second domain to gateway and route
+    gateway.refresh().add_listener(GatewayListenerTls(hostname=second_domain, gateway_name=gateway.name()))
+    gateway.apply()
+    route.refresh().add_hostname(second_domain)
+    route.apply()
```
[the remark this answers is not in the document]
azgabur: Thanks for pointing this out! `HTTPRoute` class uses `@modify` decorator on the functions that under the hood calls `modify_and_apply` function on the APIObject. I will update listeners functions to use the same!

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
trepel: Thanks for explanation! It is a pity that we just can't specify TTL in DNSPolicy. Nonetheless, would not it be sufficient to simply have 60s sleep hard-coded then? I assume we can reasonably expect that TTL is set to 60s so it would make the test stable and it is super simple solution. Might make more sense than investing time to lengthy investigation.

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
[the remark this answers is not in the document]
trepel: Ahhh, `@modify` is the culprit. I knew I would learn something new if I ask :)

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
azgabur: Hard-coded sleep 60s is an easy option as from my observation this is the default TTL for DNS records created by DNSPolicy. But this can be changed in future. We can settle on this if there are no better solutions.

azgabur: I think I found better solution. We can actually get the authoritative TTL from DNSRecord CR in Openshift. But this must be fetched before the listener gets removed from the Gateway as that will delete the DNSRecord CR. What do you think?

azgabur: I added function `get_listener_dns_ttl()` to Gateway class, take a look.

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_listeners_dns.py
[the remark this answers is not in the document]
azgabur: After a bit of consideration I agree with you. There are many tests in testsuite that can leave leftovers in cluster, SSO provider or now DNS provider. I think it should be rather expected so this warning is a bit out of place. Will remove it.

testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py
trepel: Nice!

trepel (approved):
LGTM, it seems stable now. Good work!

azgabur:
^ Rebased due to https://github.com/Kuadrant/testsuite/pull/520 merging

trepel (dismissed):

azgabur:
^ Add forgotten `name` parameter in `testsuite/tests/singlecluster/conftest.py`

trepel:
> full testsuite run should be performed as well.

Hupst, I missed this.

> ^ Add forgotten `name` parameter in `testsuite/tests/singlecluster/conftest.py`

2nd hupst, I missed this

Anyway, I ran `make test` and there were 1 ERROR at setup of test_letsencrypt (tlspolicy did not get ready in time) and 1 test_limit_for_anonymous_identity test failure. Both seem unrelated to this change, so approving again

trepel (dismissed):

testsuite/gateway/__init__.py
```diff
@@ -177,6 +177,46 @@ def remove_all_backend(self):
...
+
+    hostname: str
+    name: Optional[str] = None
+    port: int = 80
+    protocol: str = "HTTP"
+    allowedRoutes = {"namespaces": {"from": "All"}}
+
+
+@dataclass(kw_only=True)
+class GatewayListenerTls(GatewayListener):
```
jsmolar, line 196:
```suggestion
class TLSGatewayListener(GatewayListener):
```
or
```suggestion
class GatewayListenerTLS(GatewayListener):
```

testsuite/gateway/__init__.py
```diff
@@ -177,6 +177,46 @@ def remove_all_backend(self):
...
+@dataclass
+class GatewayListener:
+    """
+    Dataclass of Gateway listener object
+    If used in Gateway creation `.create_instance()` function the `name` parameter is mandatory.
+    If used in `.add_listener()` function, you can omit the `name` parameter and a unique parameter will be generated.
+    """
+
+    hostname: str
+    name: Optional[str] = None
```
jsmolar, line 189: `name` is not optional

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -1,59 +1,61 @@
...
-                }
-            ]
-
         return cls(model, context=cluster.context)
 
+    @modify
+    def add_listener(self, listener: GatewayListener):
+        """Adds a listener to Gateway. If the argument doesn't contain name, generate new sequential one."""
+        if listener.name is None:
+            listener.name = f"api{self.listener_counter}"
```
jsmolar, line 38: why is this better, than some random value?

testsuite/tests/singlecluster/gateway/reconciliation/listeners/conftest.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    def _client_new(hostname: str):
+        return StaticHostname(hostname, gateway.get_tls_cert).client()
+
+    return _client_new
+
+
+@pytest.fixture(scope="module")
+def check_ok_https(custom_client, auth):
```
averevki, line 45: just a note, we should use this fixture in more places in the future

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -1,59 +1,61 @@
...
+            self.listener_counter += 1
+        self.model.spec.listeners.append(asdict(listener))
+
+    @modify
+    def remove_listener(self, hostname: GatewayListener | str):
+        """Removes a listener based on hostname"""
+        if isinstance(hostname, GatewayListener):
```
jsmolar, line 45: Do we need the option to remove the listener based on `GatewayListener`? I don't like `hostname: GatewayListener`

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -1,59 +1,61 @@
...
+            hostname = hostname.hostname
+        self.model.spec.listeners = list(filter(lambda i: i["hostname"] != hostname, self.model.spec.listeners))
+
+    def get_listener_dns_ttl(self, hostname: Optional[str] = None) -> int:
+        """Returns TTL stored in DNSRecord CR under the default hostname or specific hostname if argument is defined."""
+        listener_name = "api"
```
jsmolar, line 51: Hardcoded `api` for listener name?

testsuite/tests/singlecluster/gateway/reconciliation/listeners/conftest.py
```diff
@@ -0,0 +1,65 @@
...
+
+import pytest
+
+from testsuite.gateway.gateway_api.hostname import StaticHostname
+
+
+@pytest.fixture(scope="module")
+def domain_prefix():
+    """Prefix of default domain."""
+    return "prefix"
```
averevki, line 14: why do you need these custom prefixes?

azgabur:
^ Rebase and incorporate feedback

testsuite/gateway/__init__.py
azgabur: Makes sense

testsuite/gateway/gateway_api/gateway.py
azgabur: I changed the logic, now listener name is used to find listener to remove

testsuite/gateway/gateway_api/gateway.py
azgabur: I removed all hard-coding, I hope

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -1,59 +1,61 @@
...
-            ]
-
         return cls(model, context=cluster.context)
 
+    @modify
+    def add_listener(self, listener: GatewayListener):
+        """Adds a listener to Gateway. If the argument doesn't contain name, generate new sequential one."""
+        if listener.name is None:
+            listener.name = f"api{self.listener_counter}"
+            self.listener_counter += 1
```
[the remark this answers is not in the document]
azgabur: Done

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -1,59 +1,61 @@
...
 
 class KuadrantGateway(KubernetesObject, Gateway):
     """Gateway object for Kuadrant"""
 
+    # Used to make unique listener names
+    listener_counter = 1
+
     @classmethod
-    def create_instance(cls, cluster: KubernetesClient, name, hostname, labels, tls=False):
+    def create_instance(cls, cluster: KubernetesClient, name, listener: GatewayListener, labels):
```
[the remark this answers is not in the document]
azgabur: Is it now better?

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -1,59 +1,61 @@
...
+            listener.name = f"api{self.listener_counter}"
+            self.listener_counter += 1
+        self.model.spec.listeners.append(asdict(listener))
+
+    @modify
+    def remove_listener(self, hostname: GatewayListener | str):
+        """Removes a listener based on hostname"""
+        if isinstance(hostname, GatewayListener):
+            hostname = hostname.hostname
+        self.model.spec.listeners = list(filter(lambda i: i["hostname"] != hostname, self.model.spec.listeners))
```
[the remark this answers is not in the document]
azgabur: Good idea, I agree. Reworked

testsuite/tests/singlecluster/gateway/reconciliation/listeners/conftest.py
azgabur: Just so each test has unique prefix in domain name for easier debugging. This is maybe redundant feature

azgabur:
Acceptance failing because of https://github.com/Kuadrant/testsuite/issues/543

testsuite/gateway/gateway_api/gateway.py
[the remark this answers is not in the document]
averevki: I was thinking more of how we setup route fixture
```py
route = HTTPRoute.create_instance(gateway.cluster, blame("route"), gateway, {"app": module_label})
route.add_hostname(hostname.hostname)
```
e.g.
```py
gw = KuadrantGateway.create_instance(...)
gw.add_listener(Listener(...))
```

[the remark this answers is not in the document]
azgabur: Agreed

testsuite/gateway/__init__.py
azgabur: It is indeed not

testsuite/tests/singlecluster/gateway/reconciliation/listeners/conftest.py
jsmolar: you can just simply do: `f'{blame("prefix1")}.{base_domain}'` Don't need to define fixture

azgabur: Done :+1:

testsuite/gateway/__init__.py
```diff
@@ -177,6 +177,44 @@ def remove_all_backend(self):
         """Sets match for a specific backend"""
 
 
+@dataclass
+class GatewayListener:
+    """
+    Dataclass of Gateway listener object.
+    When used in `add_listener()` function you MUST specify a unique name!
```
jsmolar, line 184: Is this line still needed here and in `TLSGatewayListener`?

azgabur: Yes this warning is still needed as listeners in Gateway must have unique names. I do not generate random names at runtime, this is responsibility of the user if they want to add another listener.

averevki (approved):

jsmolar (approved):

PR #536, opened 2024-09-12
Add flag for html generation

azgabur:
Inpiration from sister testsuite https://github.com/3scale-qe/3scale-tests/blob/376263a7b94faced2698b920f9599c39e1d38b44/Makefile#L33

Makes the html report include CSS instead of creating extra directory to keep the css file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ PYTEST += --junitxml=$(resultsdir)/junit-$(@F).xml -o junit_suite_name=$(@F)
 endif
 
 ifdef html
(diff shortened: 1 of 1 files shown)


trepel (approved):
LGTM. Validated using:
`html=1 make ./testsuite/tests/singlecluster/gateway/test_basic.py`

With this PR the `assets` directory with CSS was not created and the CSS code was included in head HTML tag of HTML report as expected.

averevki (approved):

PR #537, opened 2024-09-16
Skip geo test if dns not configured

averevki:


azgabur (approved):
Due to `make multicluster` target having flag `--enforce` the skip will be treated as error.

trepel (approved):
@azgabur approved this one (he's apparently fine with this) so I am approving as well.

PR #538, opened 2024-09-17
Add test for changing parentRef

Ejdm1:


averevki:
Completed in https://github.com/Kuadrant/testsuite/pull/641

PR #542, opened 2024-09-19
Update DNSPolicy model according to load balance change

jsmolar:


testsuite/kuadrant/policy/dns.py
azgabur, line 1: This change is not enough. Loadbalancing section got refactored as well so `LoadBalancing` class needs to reflect it as well. 
In the `if load_balancing:` the `routingStrategy` should be removed too

azgabur (commented):
See https://github.com/Kuadrant/kuadrant-operator/issues/855

testsuite/kuadrant/policy/dns.py
```diff
@@ -25,15 +26,9 @@ def _check(obj):
...
-
-    def asdict(self):
-        """Custom asdict due to nested structure."""
-        return {
-            "geo": {"defaultGeo": self.default_geo},
-            "weighted": {"defaultWeight": self.default_weight},
-        }
+    defaultGeo: bool
+    geo: str
+    weight: Optional[int]
```
averevki, line 31: i think weight is still a required field

azgabur: According to reference, looks like no :thinking: 
https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/reference/dnspolicy.md#loadbalancingspec

averevki: openshift dashboard says another thing. It might be wrong though

averevki (approved):

PR #554, opened 2024-10-15
Add initial dashboard and metrics

david-martin:
**UPDATED: 18th Oct 2024**

I've added 5 new metrics:
* aggregate cpu & memory by namespace
* aggregate cpu & memory by pod
* calculate the 99th percentile reconcile time by controller (in the kuadrant-system namespace)

There is also an initial grafana dashboard that shows these stats for each job (prep, main, cleanup).
See screenshot for how it looks.

The table/row at the top to show all jobs for the selected time range (so you can easily see and copy a uuid if it aligns to the time range you want to filter to).
The 2nd row shows all job summary info for the selected job UUID.

The 3rd row is actually a repeating row, and shows cpu & memory, and reconcile time for the selected job UUID, with 1 row per job name i.e. scale-test-preparations, scale-test-main, scale-test-cleanup

image

Note the scale-test-0 ns doesn't have any cpu or memory info for the prep job as the ns doesn't exist for that job.

### DIFF
--- a/scale_test/config.yaml
+++ b/scale_test/config.yaml
@@ -48,6 +48,7 @@ jobs:
     jobIterations: 1
     qps: 1
     burst: 1
+    jobPause: 2m
     namespacedIterations: true
     namespace: scale-test
     waitWhenFinished: true
--- a/scale_test/dashboard.json
+++ b/scale_test/dashboard.json
@@ -1,0 +1,1526 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "datasource",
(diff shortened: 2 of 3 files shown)


david-martin:
@trepel updated dashboard to use tables, include per pod aggregations, and fix doubling of values

image

trepel:
@david-martin lgtm, I will approve once you fix that DCO check. I validated by running the kube-burner scale test, running Grafana locally and adding an Elasticsearch data source backed by Opensearch instance hosted in AWS.

trepel (dismissed):

martinhesko (approved):

PR #562, opened 2024-10-18
Add comparison dashboard

david-martin:
image


david-martin:
@trepel I've pushed an update to take the average value for cpu & memory rather than just the most recent value. I think this will give a better representation of usage.
I used this approach for the related comparison script pr #564  as well.

trepel (approved):
I did run this in local Grafana, the Dashboard looks good to me.

averevki (approved):

PR #565, opened 2024-10-23
Fix scale listeners test/Make GatewayListener kw-only

averevki:


azgabur (approved):
For consistency this is good change

jsmolar (approved):

PR #567, opened 2024-10-25
Improve scale gateway listeners test

averevki:
- Make scale gateway listeners test more stable and comprehensive
- Add "affected by policy" condition to use with `.wait_until()`


trepel (approved):
The test passes for me. I like this, much nicer.

jsmolar (approved):

Issue #572, opened 2024-11-06
DNSPolicy health check tests


averevki:
**Documentation**

https://github.com/Kuadrant/dns-operator/issues/141
https://github.com/Kuadrant/dns-operator/pull/280

**Test cases**
- [ ] Healthy endpoint test
- [ ] Unhealthy endpoint test
- [ ] Healthy endpoint becomes unhealthy test
- [ ] Additional headers test

PR #573, opened 2024-11-06
Add DNSPolicy health check tests

averevki:
Test cases:
- Healthy endpoint with TLS
- Healthy endpoint without TLS
- Unhealthy endpoint
- Endpoint removal and return

Also added classes for the health check section of the DNSPolicy and for the DNSHealthCheckProbe CR

Part of the #572

### DIFF
--- a/testsuite/kuadrant/policy/dns.py
+++ b/testsuite/kuadrant/policy/dns.py
@@ -1,9 +1,10 @@
 """Module for DNSPolicy related classes"""
 
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, Literal
 
 from testsuite.gateway import Referencable
+from testsuite.kubernetes import KubernetesObject
 from testsuite.kubernetes.client import KubernetesClient
 from testsuite.kuadrant.policy import Policy
 from testsuite.utils import asdict, check_condition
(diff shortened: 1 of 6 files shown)


trepel:
It looks good to me and it (more or less) worked yesterday for me, but today with a new Kuadrant nightly build it does not work. The reason is that DNSPolicy keeps being successfully enforced despite failing health_check. So this piece:
```
    assert dns_policy.wait_until(has_condition("Enforced", "False"))
    assert dns_policy.wait_until(
        has_record_condition("Ready", "False", "HealthChecksFailed", "None of the healthchecks succeeded")
    )
```
is never true. Not sure if there has been any changes related to this behavior recently and whether it is intended or not but it seems like a bug to me.

trepel:
It happened to me twice that I got following error:
```
testsuite/kuadrant/policy/dns.py:99: in get_dns_health_probe
    dns_probe = self.get_owned("dnsrecords.kuadrant.io")[0].get_owned("DNSHealthCheckProbe")[0]
E   IndexError: list index out of range
```
As if the resource has not been created yet or ownership relationship has not been established quickly enough. I don't think this is something to fix in this PR - we'll see in nightlies whether this is worth investigating or not.

jsmolar (dismissed):

averevki:
> It looks good to me and it (more or less) worked yesterday for me, but today with a new Kuadrant nightly build it does not work. The reason is that DNSPolicy keeps being successfully enforced despite failing health_check. So this piece:
> 
> ```
>     assert dns_policy.wait_until(has_condition("Enforced", "False"))
>     assert dns_policy.wait_until(
>         has_record_condition("Ready", "False", "HealthChecksFailed", "None of the healthchecks succeeded")
>     )
> ```
> 
> is never true. Not sure if there has been any changes related to this behavior recently and whether it is intended or not but it seems like a bug to me.

Thanks, there were indeed some changes in the dnspolicy. Adjusted now :+1:

trepel (approved):
Code looks good to me. Tests passed against November 19 nightly build. Approving this one.

Issue #575, opened 2024-11-07
Refactor RateLimitPolicy to v1beta3


azgabur:
https://github.com/Kuadrant/kuadrant-operator/pull/976

We need to refactor Limitador tests to use CEL predicates instead of `when`
We need change the Limit object `duration` -> `window` and `unit` is removed

PR #577, opened 2024-11-11
Change Limit defition to refelct v1beta3

jsmolar:
Resolves https://github.com/Kuadrant/testsuite/issues/575

- Changes Limit object to reflect the current Limit api
- Swaps currently used Rules from RLP to CEL expressions and predicates

### DIFF
--- a/testsuite/kuadrant/policy/__init__.py
+++ b/testsuite/kuadrant/policy/__init__.py
@@ -1,7 +1,23 @@
 """Contains Base class for policies"""
 
+from dataclasses import dataclass
+
 from testsuite.kubernetes import KubernetesObject
 from testsuite.utils import check_condition
+
+
+@dataclass
+class CelPredicate:
+    """Dataclass that references CEL predicate e.g. auth.identity.anonymous == 'true'"""
+
+    predicate: str
+
+
+@dataclass
+class CelExpression:
+    """Dataclass that references CEL expression"""
+
+    expression: str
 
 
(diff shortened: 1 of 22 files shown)


testsuite/kuadrant/policy/__init__.py
```diff
@@ -1,9 +1,25 @@
...
+    """Dataclass that references CEL predicate e.g. auth.identity.anonymous == 'true'"""
+
+    predicate: str
+
+
+@dataclass
+class CelExpression:
+    """Dataclass that references CEL expression"""
+
+    expression: str
```
averevki, line 20: I'd leave these 2 as a string. Also, they should be under RateLimit module if they are only used there

averevki (approved):

testsuite/kuadrant/policy/__init__.py
jsmolar: what do you mean by leaving them as string?

martinhesko (approved):

testsuite/kuadrant/policy/__init__.py
averevki: I don't see the reason to use dataclasses here

jsmolar:
```
      when:
        - predicate: "request.method == 'POST'" 
```

This is predicate in action. For this you would need to specify "predicate" every time

Issue #579, opened 2024-11-11
AllNamespaces API Key test is now irrelevant


jsmolar:
**Summary**

Kuadrant creates all AuthConfigs in the same namespace where Authorino is running. Therefore we changed default value for `spec.authentication.apiKey.allNamespaces` to `True`. This makes `test_all_namespaces_api_key.py` irrelevant. 

Refactor/remove this test and implement one that creates a secret in the same namespace as Authorino.

PR #580, opened 2024-11-11
Make API Key auth look for all namespaces

jsmolar:
For reason see: https://github.com/Kuadrant/testsuite/issues/579

### DIFF
--- a/testsuite/kuadrant/policy/authorization/sections.py
+++ b/testsuite/kuadrant/policy/authorization/sections.py
@@ -138,7 +138,7 @@ class IdentitySection(Section):
         name,
         selector: Selector,
         *,
-        all_namespaces: bool = False,
+        all_namespaces: bool = True,
         credentials: Credentials = None,
         **common_features,
     ):


martinhesko (approved):

testsuite/kuadrant/policy/authorization/sections.py
```diff
@@ -138,7 +138,7 @@ def add_api_key(
         name,
         selector: Selector,
         *,
-        all_namespaces: bool = False,
+        all_namespaces: bool = True,
```
averevki, line 141: I dont understand the change. Cant you just change it in `test_all_namespaces_api_key.py` test only instead?

jsmolar: Kuadrant creates all AuthConfigs in the namespace where Authorino is running. In our case, they are moved from `kuadrant` to `kuadrant-system` namespaces.
I need either force authorino to look into all namespaces (like I did here), or create all API key secrets in `kuadrant-system` namespaces. What is better?

PR #596, opened 2024-11-15
Fix gateway reconciliation tests

averevki:


trepel (approved):
Both tests passed. Code looks good to me.

jsmolar (approved):

PR #597, opened 2024-11-18
Change prometheus query when getting metrics for limits

jsmolar:
the name of RLP is no longer used in metrics. Now we use `targetRef` name.


averevki (approved):

trepel (approved):
Test passes against Nov 19 nightly build, approving this one.

PR #599, opened 2024-11-20
Update 'identical hostname' tests since the behavior has changed

trepel:
## Overview

The 'identical hostname' limitation is no longer there. It should work the same as if the hostnames were not identical.

## What was done

Removed the `*_desired.py` tests since the desired behavior is now validated by the remaining tests. These tests were moved to `auth` and `rlp` directories and updated so that they validates the expected behavior and not the behavior with 'identical hostname' limitation.

## Verification Steps

`make ./testsuite/tests/singlecluster/identical_hostnames`

All four tests should pass


testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,78 @@
...
+    # 'deny-all' Gateway AuthPolicy affects route2
+    response = client.get("/anything/route2/get")
+    assert response.status_code == 403
+
+    # Deletion of 'allow-all' AuthPolicy should make the 'deny-all' Gateway AuthPolicy enforced on both routes
+    authorization.delete()
+    authorization2.wait_for_ready()
+
+    response = client.get("/anything/route2/get")
+    assert response.status_code == 403
```
averevki, line 75:
```suggestion
    response = client.get("/anything/route1/get")
    assert response.status_code == 403
```

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,78 @@
...
+            request.addfinalizer(auth.delete)
+            auth.commit()
+            auth.wait_for_accepted()
+
+    authorization.wait_for_ready()
+
+    # At this point the 'route2' has not been created yet so authorization2 is completely overridden by authorization1
+    assert authorization2.wait_until(has_condition("Enforced", "False", "Overridden")), (
+        f"'deny-all' AuthPolicy did not reach expected record status, instead it was: "
+        f"{authorization2.model.status.recordConditions}"
```
averevki, line 38:
```suggestion
        f"{authorization2.model.status..conditions}"
```
and in other places

averevki (commented):
tested locally, looks good :+1:. Only few comments

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_gw_and_route.py
trepel: Nice catch! Thanks. Will fix and force push

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_gw_and_route.py
trepel: Good catch again, I found one other occurrence of this. Will force push.

trepel:
@averevki Thanks for review! Your suggestions have been force pushed. Could you please take a look if you get a chance?

testsuite/tests/singlecluster/identical_hostnames/auth/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+
+@pytest.fixture(scope="module")
+def authorization(authorization):
+    """1st 'allow-all' Authorization object"""
+    authorization.authorization.add_opa_policy("rego", "allow = true")
+    return authorization
+
+
+@pytest.fixture(scope="module")
+def rate_limit():
```
jsmolar, line 14: Common practice across testsuite is creating new `commit` fixtures, with objects you want to create. In this case just `authorization`

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_gw_and_route.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.fixture(scope="module")
+def authorization2(gateway, blame, cluster, label):
+    """2nd 'deny-all' Authorization object"""
+    auth_policy = AuthPolicy.create_instance(cluster, blame("authz2"), gateway, labels={"testRun": label})
+    auth_policy.authorization.add_opa_policy("rego", "allow = false")
+    return auth_policy
+
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization, authorization2):
```
jsmolar, line 25: This fixture should be in auth conftest. The same applies for `commit` in `test_identical_hostnames_auth_on_routes.py`

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_routes.py
```diff
@@ -0,0 +1,65 @@
...
+
+import pytest
+
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+
+pytestmark = [pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="module")
+def authorization2(route2, blame, cluster, label):
```
jsmolar, line 16: This should also be in auth conftest

jsmolar (changes requested):

jsmolar:
Great test rewrite! To sum up my comments here. Please move `authorization2`, `rate_limit2`, and `commit` to conftests.
In addition, python package where these tests are located is called `identical_hostnames`. There is no need to include this into test name.

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_gw_and_route.py
trepel: This fixture is slightly different for each test. In 'on route' scenario both auth policies are fully enforced whereas in 'on gw' scenario the 2nd authpolicy (the one attached to GW rather than HTTPRoute) is overridden.
It did not occur to me how to handle this difference nicely in the fixture so that it can be used for both tests, that is the reason why I placed it in the test file rather than conftest. Suggestions welcome!

testsuite/tests/singlecluster/identical_hostnames/auth/test_identical_hostnames_auth_on_routes.py
trepel: Same as above. For one test this auth policy is attached to HTTPRoute, for the other it is attached to GW. Again, not sure how to handle this nicely so that it can be shared in both tests.

testsuite/tests/singlecluster/identical_hostnames/auth/conftest.py
trepel: So in theory this can be removed I assume. I have a `commit` fixture defined in the test files directly and RLP is not committed there.

trepel:
> Great test rewrite! To sum up my comments here. Please move `authorization2`, `rate_limit2`, and `commit` to conftests. In addition, python package where these tests are located is called `identical_hostnames`. There is no need to include this into test name.

The names of test functions are not ideal, a lot of duplicit info there, I agree. The reason I have it this way is that it helps to identify the test in the report portal. RP only shows test function name, that's why it is this long. However, you are right that 'identical_hostnames' is redundant in the name of the test files. I will rename.

testsuite/tests/singlecluster/identical_hostnames/auth/conftest.py
trepel: It seems it indeed can be removed. Will force push and let you know.

trepel: @jsmolar force-pushed now

PR #604, opened 2024-11-21
Fix multicluster tests

averevki:


trepel (approved):
I confirmed with eng what the changes here do is expected behavior. Code looks good to me. Tests are passing against v1. Approving.

jsmolar (approved):

PR #615, opened 2025-01-13
DNSPolicy scale test

mikenairn:
Adds a DNSPolicy specific scale test using kube burner.

Based on the existing [scale test](https://github.com/Kuadrant/testsuite/tree/main/scale_test), but with a focus on DNSPolicy and shared hostnames being updated by multiple dns operator instances.

The workload will create multiple instances of the dns operator in separate namespaces(kuadrant-dns-operator-x), and multiple test namespaces (scale-test-x) that the corresponding dns operator is configured to watch.  The number of dns operator instances and test namespaces created is determined by the `JOB_ITERATIONS` environment variable.
In each test namespace a test app and service is deployed and one or more gateways are created determined by the `NUM_GWS` environment variable.  The number of listeners added to the gateway is determined by the `NUM_LISTENERS` environment variable.
Each listener hostname is generated using the listener number and the `KUADRANT_ZONE_ROOT_DOMAIN` environment variable.  In each test namespace a dns provider credential is created, the type created is determined by the `DNS_PROVIDER` environment variable, additional environment variables may need to be set depending on the provider type.


Makefile
```diff
@@ -102,9 +102,9 @@ reportportal: ## Upload results to reportportal. Appropriate variables for juni2
 reportportal: polish-junit
 	$(RUNSCRIPT)junit2reportportal $(resultsdir)/junit-*.xml
 
-# Check http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
-help: ## Print this help
-	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
mikenairn, line 107: Optional change, just brings it in-line with the help in other repos, you can use `##@ foo` to add sections:

Before:

```
$ make help
commit-acceptance              Runs pre-commit linting checks
reformat                       Reformats testsuite with black
test                           Run all non mgc tests
authorino                      Run only authorino related tests
authorino-standalone           Run only test capable of running with standalone Authorino
limitador                      Run only Limitador related tests
kuadrant                       Run all tests available on Kuadrant
kuadrant-only                  Run Kuadrant-only tests
multicluster                   Run Multicluster only tests
dnstls                         Run DNS and TLS tests
disruptive                     Run disruptive tests
kuadrantctl                    Run Kuadrantctl tests
poetry                         Installs poetry with all dependencies
poetry-no-dev                  Installs poetry without development dependencies
polish-junit                   Remove skipped tests and logs from passing tests
reportportal                   Upload results to reportportal. Appropriate variables for juni2reportportal must be set
help                           Print this help
clean                          Clean all objects on cluster created by running this testsuite. Set the env variable USER to delete after someone else
test-scale-dnspolicy           Run DNSPolicy scale tests.
kube-burner                    Download kube-burner locally if necessary.
```

After:
```
$ make help

Usage:
  make <target>
  commit-acceptance  Runs pre-commit linting checks
  reformat         Reformats testsuite with black
  test             Run all non mgc tests
  authorino        Run only authorino related tests
  authorino-standalone  Run only test capable of running with standalone Authorino
  limitador        Run only Limitador related tests
  kuadrant         Run all tests available on Kuadrant
  kuadrant-only    Run Kuadrant-only tests
  multicluster     Run Multicluster only tests
  dnstls           Run DNS and TLS tests
  disruptive       Run disruptive tests
  kuadrantctl      Run Kuadrantctl tests
  poetry           Installs poetry with all dependencies
  poetry-no-dev    Installs poetry without development dependencies
  polish-junit     Remove skipped tests and logs from passing tests
  reportportal     Upload results to reportportal. Appropriate variables for juni2reportportal must be set
  help             Display this help.
  clean            Clean all objects on cluster created by running this testsuite. Set the env variable USER to delete after someone else

Scale Testing
  test-scale-dnspolicy  Run DNSPolicy scale tests.

Build Dependencies
  kube-burner      Download kube-burner locally if necessary.
```

scale_test/dnspolicy/README.md
```diff
@@ -0,0 +1,144 @@
...
+Scale down the default dns operator:
+```shell
+kubectl scale deployment/dns-operator-controller-manager --replicas=0 -n kuadrant-system
+```
+Note: This is required for the default kubeburner workload (namespaced-dns-operator-deployments.yaml) as it creates its own dns operator deployments.
+
+Deploy the observability stack:
+```shell
+#kubectl apply --server-side -k github.com/mikenairn/dns-operator/config/observability?ref=add_scale_test
+kubectl apply --server-side -k github.com/kuadrant/dns-operator/config/observability?ref=main # Run twice if it fails the first time
```
mikenairn, line 26: This needs to be updated.

I think a single kustomization in the kuadrant operator that can be executed, without the need to pull down the repo, would be useful and easier than having to run this set of steps https://github.com/Kuadrant/kuadrant-operator/tree/main/config/observability#deploying-the-observabilty-stack. Depends how varied the observability setup gets.

Thanos setup should likely be a secondary optional task when working with a single cluster.  

@david-martin Any opinions/thoughts on this?

scale_test/dnspolicy/namespaced-dns-operator-deployments.yaml
```diff
@@ -0,0 +1,162 @@
...
+  {{ if (eq .SKIP_CLEANUP "true") }}
+  # nothing to do if cleanup is skipped
+  {{ else }}
+  - name: dnspolicy-scale-test-cleanup
+    jobType: delete
+    objects:
+      - kind: DNSPolicy
+        apiVersion: kuadrant.io/v1alpha1
+        labelSelector: {kube-burner-job: dnspolicy-scale-test-loadbalanced}
+  {{ end }}
```
mikenairn, line 162: I copied the pattern from the [original scale test](https://github.com/Kuadrant/testsuite/blob/main/scale_test/config.yaml#L139) of adding this cleanup to remove the DNSPolicies. I understand why its here, but it is fairly frustrating that we need it.

I wonder if we should revisit the cleanup of policies/records/secrets and see if there is some reasonable way we can prevent secrets being removed before all resources referencing them are deleted.

@maleck13 I know we have discussed this before, but i think its probably the single most annoying thing about testing anything DNS related.

scale_test/dnspolicy/namespaced-dns-operator-deployments.yaml
```diff
@@ -0,0 +1,162 @@
+metricsEndpoints:
+  - endpoint: {{ .PROMETHEUS_URL }}
+    token: {{ .PROMETHEUS_TOKEN }}
+    metrics:
+      - https://raw.githubusercontent.com/{{.DNS_OPERATOR_GITHUB_ORG}}/dns-operator/refs/heads/{{.DNS_OPERATOR_GITREF}}/test/scale/metrics.yaml
+    alerts:
+      - https://raw.githubusercontent.com/{{.DNS_OPERATOR_GITHUB_ORG}}/dns-operator/refs/heads/{{.DNS_OPERATOR_GITREF}}/test/scale/alerts.yaml
+    indexer:
+      type: local
+      metricsDirectory: ./metrics
```
mikenairn, line 10: I have the alerts and metrics being pulled from the dns operator repo here, but i imagine we could have these being pulled from multiple sources i.e. kuadrant-operator, testsuite repo, other components, where they define their own metrics/alerts specific to the resources they are providing.

The metrics/alerts configured, from what i can gather, are really needed to make the most out of kubeburner runs since alerts firing during the run are what will tell us if things are working or not, and what we would need to improve on if we feel these types of scale tests are useful.

scale_test/dnspolicy/README.md
david-martin: The cmd you have here should be enough to deploy the stack (without thanos or example alerts & dashboards).
You would typically need to run it twice if CRDs don't exist.
Once to just include CRDs so they are registered, and a 2nd time (without CRDs) to avoid errors creating CRs.

scale_test/dnspolicy/namespaced-dns-operator-deployments.yaml
david-martin: I wonder if the alerts and metrics files should be maintained in this repo for easier maintenance in the context of running and maintaining tests.
The alternative could result in extra toil, particularly when working out the details of assertions for a test.

scale_test/dnspolicy/namespaced-dns-operator-deployments.yaml
maleck13: Yes. Well I guess the only real option is to add a finalizer to secret once referenced by a policy as a provider. I guess as the perm there would just be to add fnalizers it wouldn't mean we have full write access to secrets. The more I think about the more reasonable it seems to add a finalizer and remove it if the ref changes. As more than one policy can reference a single secret, we would likely need some form of ref count adding as an annotation for example? 
Anyway I think it makes sense to add a new issue around this given how much hassle it causes people

maleck13 (commented):
I didn't actually try this and would prefer if @trepel or another of the QE team took a look and approved, but the changes look good to me

scale_test/dnspolicy/namespaced-dns-operator-deployments.yaml
mikenairn: Issue to look into this more here https://github.com/Kuadrant/dns-operator/issues/344

trepel:
I tried against OCP cluster and Route53 and it worked as described - except for that missing `github.com/kuadrant/dns-operator/config/observability?ref=main` but that's being discussed. What surprises me was that no matter how many GWs/HTTPRoutes are created there is just NUM_LISTENERS unique hostnames. That's by design if I got it right, and it seems to be working (checked by `dig +short` )

maleck13:
@mikenairn do we want to move this to ready  and get it merged ? @trepel from your comment seems good to merge?

mikenairn:
> What surprises me was that no matter how many GWs/HTTPRoutes are created there is just NUM_LISTENERS unique hostnames.

This is intentional for the workload being added since it's testing dns at scale, part of that is testing that multiple records all contributing to the same dns name works with increasing numbers of owners.

maleck13:
@trepel are you ok to approve and merge this?

scale_test/dnspolicy/namespaced-dns-operator-deployments.yaml
trepel: Resolving the conversation in order to merge this. Let's discuss in slack if needed.

Makefile
trepel: makes sense to me, resolving the conversation.

PR #621, opened 2025-01-20
Fix unsynced objects in multicluster tests

averevki:


trepel:
@averevki lgtm. Do you think that it would be worth going through all the tests and update then to follow this approach (calling refresh() and then asserting the result)?

trepel (approved):
All tests in load_balanced dir passed for me.

averevki:
> @averevki lgtm. Do you think that it would be worth going through all the tests and update then to follow this approach (calling refresh() and then asserting the result)?

I'm honestly still not sure why did this break in the first place. The issue only seems to occur when running tests inside a pod

PR #629, opened 2025-01-30
Increase TLSPolicy timeout for Azure TLS ACME challenges

averevki:
Azure TLS ACME challenges take 7+ minutes on the multiclusters


trepel (approved):
LGTM - downside is that this increases the timeout for TLSPolicy in general, not only for Azure but I agree that it is not worth the effort to try to implement this for Azure only.

Issue #630, opened 2025-02-13
Fix tests with gateway attached Policies


azgabur:
# Description
Tests located in `.../gateway/` module should test gateway attached policies. Currently this is not the case.

# How to check
1. Run `testsuite/tests/singlecluster/gateway/test_basic.py` test
2. Created AuthPolicy should be attached to Gateway and not to HTTPRoute

PR #633, opened 2025-02-17
Fix AuthPolicy attachment to Gateway instead of HTTPRoute

emmaaroche:
**Summary**  
This PR ensures that the `AuthPolicy` is correctly attached to the `Gateway` instead of the `HTTPRoute`, fixing the unintended behavior in some tests.

**Changes Made**  
- Updated `authorization` fixture in `testsuite/tests/singlecluster/conftest.py` to use `gateway` as the target reference.
- Removed an `authorization` fixture that was used to override the AuthPolicy in `test_update_authpolicy_target_ref.py`.

**Related Issue**  
[#630](https://github.com/Kuadrant/testsuite/issues/630)


testsuite/tests/singlecluster/conftest.py
```diff
@@ -41,7 +41,7 @@ def authorization_name(blame):
 @pytest.fixture(scope="module")
 def authorization(request, kuadrant, route, gateway, blame, cluster, label):  # pylint: disable=unused-argument
     """Authorization object (In case of Kuadrant AuthPolicy)"""
-    target_ref = request.getfixturevalue(getattr(request, "param", "route"))
+    target_ref = request.getfixturevalue(getattr(request, "param", "gateway"))
```
azgabur, line 44: This will indeed make all tests in `.../gateway/` module have polices that attach to Gateway but so will all tests in other modules which do not expect this :sweat_smile: 
Only tests in `.../gateway/` module should have the Gateway attachment the default. I would do changes in the `.../gateway/conftest.py` to achieve this

azgabur (changes requested):

averevki (approved):

azgabur (approved):

PR #637, opened 2025-02-18
D&O section target tests

averevki:
Refined part from the https://github.com/Kuadrant/testsuite/pull/608
- Section targeting tests for gateway and route
- Multiple ratelimitpolicies test for overrides


trepel (approved):
Tests passed. Code looks OK-ish enough to me to approve this for merging.

PR #638, opened 2025-02-18
Add tests for D&O merge strategy

### LINKED ISSUE #566: Defaults & Overrides: Merge strategy tests
**Summary**

Investigate and implement the necessary test cases for test coverage of the merge strategy of Defaults & Overrides with regards to both `RateLimitPolicy` and `AuthPolicy`

- Implement test cases based on the provided verification steps in the related PRs.
- Develop test cases based on use cases listed at the bottom of the related PRs.
- Investigate and implement test cases covering other scenarios 

**Documentation**
[D&O RateLimitPolicy Merge strategy PR](https://github.com/Kuadrant/kuadrant-operator/pull/893)
[D&O AuthPolicy Merge strategy PR](https://github.com/Kuadrant/kuadrant-operator/pull/952)
[Original D&O RFC](https://github.com/Kuadrant/architecture/blob/main/rfcs/0009-defaults-and-overrides.md)

**Test cases**

[docs](https://docs.google.com/document/d/131HkIFcgA-XMsat60ja4aWODnUbd-m44tTOWhoIrLS0/edit?usp=sharing)

averevki:
Refined part of https://github.com/Kuadrant/testsuite/pull/608 with merge strategy tests only

Closes #566

### DIFF
--- a/testsuite/kuadrant/policy/authorization/auth_policy.py
+++ b/testsuite/kuadrant/policy/authorization/auth_policy.py
@@ -9,7 +9,7 @@ from testsuite.kubernetes.client import KubernetesClient
 from testsuite.utils import asdict
 from .auth_config import AuthConfig
 from .sections import ResponseSection
-from .. import Policy, CelPredicate
+from .. import Policy, CelPredicate, Strategy
 from . import Pattern
 
 
@@ -48,6 +48,15 @@ class AuthPolicy(Policy, AuthConfig):
         """Add rule for the skip of entire AuthPolicy"""
         self.model.spec.setdefault("when", [])
         self.model.spec["when"].extend([asdict(x) for x in when])
+
+    @modify
+    def strategy(self, strategy: Strategy) -> None:
+        """Add strategy type to default or overrides spec"""
+        if self.spec_section is None:
+            raise TypeError("Strategy can only be set on defaults or overrides")
+
+        self.spec_section["strategy"] = strategy.value
+        self.spec_section = None
 
     @property
     def auth_section(self):
--- a/testsuite/kuadrant/policy/rate_limit.py
+++ b/testsuite/kuadrant/policy/rate_limit.py
@@ -7,7 +7,7 @@ from typing import Iterable
 from testsuite.gateway import Referencable
 from testsuite.kubernetes import modify
 from testsuite.kubernetes.client import KubernetesClient
-from testsuite.kuadrant.policy import Policy, CelPredicate, CelExpression
+from testsuite.kuadrant.policy import Policy, CelPredicate, CelExpression, Strategy
 from testsuite.utils import asdict
 
 
@@ -70,6 +70,15 @@ class RateLimitPolicy(Policy):
             self.spec_section = self.model.spec
 
         self.spec_section.setdefault("limits", {})[name] = limit
+        self.spec_section = None
+
+    @modify
+    def strategy(self, strategy: Strategy) -> None:
+        """Add strategy type to default or overrides spec"""
+        if self.spec_section is None:
+            raise TypeError("Strategy can only be set on defaults or overrides")
+
+        self.spec_section["strategy"] = strategy.value
         self.spec_section = None
 
     @property
--- a/testsuite/tests/singlecluster/defaults/merge/same_target/test_ab_strategy.py
+++ b/testsuite/tests/singlecluster/defaults/merge/same_target/test_ab_strategy.py
@@ -1,0 +1,38 @@
+"""Test defaults policy aimed at the same resource uses the oldest policy."""
+
+import pytest
+
+from testsuite.kuadrant.policy import has_condition
+from .conftest import MERGE_LIMIT, MERGE_LIMIT2
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.limitador]
+
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, route, rate_limit, default_merge_rate_limit):  # pylint: disable=unused-argument
+    """Commits RateLimitPolicy after the HTTPRoute is created"""
+    for policy in [rate_limit, default_merge_rate_limit]:  # Forcing order of creation.
+        request.addfinalizer(policy.delete)
+        policy.commit()
+        policy.wait_for_accepted()
+
+
+def test_multiple_policies_merge_default_ab(client, rate_limit, default_merge_rate_limit):
+    """Test RateLimitPolicy with merge defaults being enforced due to age"""
+    assert rate_limit.wait_until(
+        has_condition(
+            "Enforced",
+            "False",
+            "Overridden",
+            "RateLimitPolicy is overridden by "
+            f"[{default_merge_rate_limit.namespace()}/{default_merge_rate_limit.name()}]",
+        )
+    )
+
+    responses = client.get_many("/get", MERGE_LIMIT.limit)
+    responses.assert_all(status_code=200)
+    assert client.get("/get").status_code == 429
+
+    responses = client.get_many("/anything", MERGE_LIMIT2.limit)
+    responses.assert_all(status_code=200)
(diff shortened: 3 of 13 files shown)


testsuite/tests/singlecluster/defaults/merge/same_target/test_ab_strategy.py
```diff
@@ -0,0 +1,27 @@
+"""Test defaults policy aimed at the same resoure uses oldested policy."""
```
trepel, line 1: resoure / oldested - typos?

testsuite/tests/singlecluster/defaults/merge/same_target/test_ba_startegy.py
```diff
@@ -0,0 +1,27 @@
+"""Test defaults policy aimed at the same resoure uses oldested policy."""
```
trepel, line 1: Typos here too

testsuite/tests/singlecluster/defaults/merge/test_gateway_default_merge.py
```diff
@@ -0,0 +1,51 @@
+"""Test gateway level default merging with and being patrially overriden by another policy."""
```
trepel, line 1: *partially overridden

testsuite/kuadrant/policy/rate_limit.py
```diff
@@ -63,6 +72,15 @@ def add_limit(
...
         self.spec_section = None
 
+    @modify
+    def strategy(self, strategy: Strategy) -> None:
+        """Add strategy type to default or overrides spec"""
+        if self.spec_section is None:
+            raise TypeError("Strategy can only be set on defaults or overrides")
+
+        self.spec_section["strategy"] = strategy.value
+        self.spec_section = None
```
trepel, line 82: Same again, unclear to me why there is this "None" assignment

testsuite/tests/singlecluster/defaults/merge/same_target/test_ab_strategy.py
```diff
@@ -0,0 +1,27 @@
...
+    """Commits RateLimitPolicy after the HTTPRoute is created"""
+    for policy in [rate_limit, default_merge_rate_limit]:  # Forcing order of creation.
+        request.addfinalizer(policy.delete)
+        policy.commit()
+        policy.wait_for_accepted()
+
+
+def test_multiple_policies_merge_default_ab(client):
+    """Test RateLimitPolicy with merge defaults being ignored due to age"""
+    responses = client.get_many("/get", MERGE_LIMIT.limit)
```
trepel, line 21: To be honest I think that the comments need to be swapped between test_ab_strategy and test_ba_strategy. If I got it right "a" means standard policy and "b" means policy with merge defaults.

So in "ab" the "a" policy is created first, and once "b" is created the "a" gets overridden - you can see it in test that "MERGE_LIMIT" is used for both `get_many` calls. However, the comment in "ab" suggests that "b" is ignored which is not true - it is true in the "ba" test where the policies are created in different order.

trepel: I decrypted the policy names.
`dmp` means policy with defauls section and merge strategy (defaults merge policy)
`sp` means standard policy === policy without both defaults and overrides sections
`omp` means policy with overrides section and merge strategy (overrides merge policy)

testsuite/tests/singlecluster/overrides/merge/same_target/test_ab_strategy.py
```diff
@@ -0,0 +1,27 @@
...
+def commit(request, route, rate_limit, override_merge_rate_limit):  # pylint: disable=unused-argument
+    """Commits RateLimitPolicy after the HTTPRoute is created"""
+    for policy in [rate_limit, override_merge_rate_limit]:  # Forcing order of creation.
+        request.addfinalizer(policy.delete)
+        policy.commit()
+        policy.wait_for_accepted()
+
+
+def test_multiple_policies_merge_default_ab(client):
+    """Test RateLimitPolicy with merge overrides being ignored due to age"""
```
trepel, line 20: The RateLimitPolicy with merge overrides is not ignored. On the contrary it is fully enforced

trepel:
I don't like two additional things:
1) In the bodies of test functions the policy status is not asserted (whether policy is indeed overridden/fully and/or partially enforced).
2) The commit fixture is overloaded so that it only check for policy acceptance. I think (this is philosophical and up to wider discussion maybe) that it is better if commit fixture only deals with policies that are fully enforced (I consider this as initial test setup fixtures should be responsible for because that is what they are designed for) and policies that are overridden/partially enforced (or causes some existing policies to get overridden/partially enforced) should be created in test function bodies (I consider this test scenario rather that initial test setup so it does not belong to fixtures anymore).

testsuite/tests/singlecluster/overrides/merge/same_target/test_ab_strategy.py
averevki: I think this behavior is correct, overrides with the merge strategy always take precedence. Some examples from the docs: https://docs.kuadrant.io/1.0.x/architecture/rfcs/0009-defaults-and-overrides/#examples-d-override-policy-rules-merged-into-other-at-lower-level

testsuite/kuadrant/policy/authorization/auth_policy.py
```diff
@@ -46,6 +49,15 @@ def add_rule(self, when: list[CelPredicate]):
...
         self.model.spec["when"].extend([asdict(x) for x in when])
 
+    @modify
+    def strategy(self, strategy: Strategy) -> None:
+        """Add strategy type to default or overrides spec"""
+        if self.spec_section is None:
+            raise TypeError("Strategy can only be set on defaults or overrides")
+
+        self.spec_section["strategy"] = strategy.value
+        self.spec_section = None
```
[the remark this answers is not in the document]
averevki: This function should be called together with `.defaults` or `.overrides` "spec_section", for example `policy.defaults.strategy(Strategy.MERGE)`. When you call `.defaults` or `.overrides` properties, they assign `self.spec_section` to the current policy section (overrides, defaults, none (implicit defaults)) this object adds to. This property must be reset or it also will be used on the next call of `RateLimitPolicy` object, and will add rules to the section from the previous call.

We reset it at the end of add_limit/add_rule, and now we should reset it here too

https://github.com/Kuadrant/testsuite/blob/6533e9a425a0e4fa37dbbcbaac91d47cf2ea8c26/testsuite/kuadrant/policy/rate_limit.py#L69-L73

Just a bad design, we can come up with something better

averevki:
> In the bodies of test functions the policy status is not asserted (whether policy is indeed overridden/fully and/or partially enforced).

good idea, done

> The commit fixture is overloaded so that it only check for policy acceptance. I think (this is philosophical and up to wider discussion maybe) that it is better if commit fixture only deals with policies that are fully enforced (I consider this as initial test setup fixtures should be responsible for because that is what they are designed for) and policies that are overridden/partially enforced (or causes some existing policies to get overridden/partially enforced) should be created in test function bodies (I consider this test scenario rather that initial test setup so it does not belong to fixtures anymore).

agreed, there are a lot of other tests doing that inside the testsuite though. We can do another single PR with all these changes in the future

testsuite/kuadrant/policy/authorization/auth_policy.py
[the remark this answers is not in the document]
trepel: Uh, ok, so my understanding now is: "spec_section" is reference (pointer) to the certain section of spec that is currently being worked with. Once the work with that section is done it needs to be cleared.

Anyway, setting it to None is apparently OK.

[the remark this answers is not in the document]
trepel: Thanks for explanation!

PR #622, opened 2025-01-21
Test for changing parentRef field in policies (AuthPolicy and RateLimitPolicy)

emmaaroche:
Test for changing parentRef field in policies (AuthPolicy and RateLimitPolicy)

Issue #307, opened 2023-12-14
Add test for changing targetRef field in policies


ficap:
**Summary**
Create reconciliation policy tests that focus on targetRef field. 
For example if targetRef of a DNSPolicy points to a gateway with hostname A, try changing targetRef to another gateway with hostname B. DNS records should be updated accordingly. Similarly with TLSPolicy. May also apply to AuthPolicy and RateLimitPolicy.

**Documentation**

**Test cases**

* Change targetRef of a DNSPolicy from Gateway to HTTPRoute and vice versa
  * with unchanged hostname
  * with changed hostname
* Change targetRef of a TLSPolicy from Gateway to HTTPRoute and vice versa
  * with unchanged hostname
  * with changed hostname
* Change targetRef of a DNSPolicy from Gateway A with hostname A to Gateway B with hostname B
* Change targetRef of a TLSPolicy from Gateway A with hostname A to Gateway B with hostname B

PR #641, opened 2025-02-20
Add tests for changing targetRef field in policies

emmaaroche:
**Summary**
This PR adds reconciliation tests focusing on targetRef updates across different policies. 

**Test Cases Covered**
* DNSPolicy: Verifies DNS records are updated accordingly when changing `targetRef` from Gateway 1 (hostname 1) to Gateway 2 (hostname 2).
* TLSPolicy: Verifies TLS records are updated accordingly when changing `targetRef` from Gateway 1 (hostname 1) to Gateway 2 (hostname 2).
* AuthPolicy and RateLimitPolicy were covered in PR #622), but were slightly modified in this PR.

**Related Issue**
Closes [#307](https://github.com/Kuadrant/testsuite/issues/307)

### DIFF
--- a/testsuite/gateway/gateway_api/gateway.py
+++ b/testsuite/gateway/gateway_api/gateway.py
@@ -89,6 +89,19 @@ class KuadrantGateway(KubernetesObject, Gateway):
         return False
 
     def get_tls_cert(self, hostname):
+        tls_cert_secret = self.get_tls_secret(hostname)
+        if tls_cert_secret is None:
+            return None
+
+        tls_cert = Certificate(
+            key=tls_cert_secret["tls.key"],
+            certificate=tls_cert_secret["tls.crt"],
+            chain=tls_cert_secret["ca.crt"] if "ca.crt" in tls_cert_secret else None,
+        )
+        return tls_cert
+
+    def get_tls_secret(self, hostname):
+        """Returns the TLS secret for the matching listener hostname, or None if not found"""
         tls_cert_secret_name = None
         for listener in self.all_tls_listeners():
             if domain_match(hostname, listener.hostname):
@@ -98,17 +111,11 @@ class KuadrantGateway(KubernetesObject, Gateway):
             return None
 
         try:
-            tls_cert_secret = self.cluster.get_secret(tls_cert_secret_name)
+            return self.cluster.get_secret(tls_cert_secret_name)
         except oc.OpenShiftPythonException as e:
             if "Expected a single object, but selected 0" in e.msg:
                 raise oc.OpenShiftPythonException("TLS secret was not created") from None
             raise e
-        tls_cert = Certificate(
-            key=tls_cert_secret["tls.key"],
-            certificate=tls_cert_secret["tls.crt"],
-            chain=tls_cert_secret["ca.crt"] if "ca.crt" in tls_cert_secret else None,
-        )
-        return tls_cert
 
     def all_tls_listeners(self):
         """Yields all listeners in gateway that support 'tls'"""
--- a/testsuite/httpx/__init__.py
+++ b/testsuite/httpx/__init__.py
@@ -59,7 +59,11 @@ class Result:
 
     def has_dns_error(self):
         """True, if the result failed due to DNS failure"""
-        return self.has_error("Name or service not known") or self.has_error("No address associated with hostname")
+        return (
+            self.has_error("nodename nor servname provided, or not known")
+            or self.has_error("Name or service not known")
+            or self.has_error("No address associated with hostname")
+        )
 
     def has_cert_verify_error(self):
         """True, if the result failed due to TLS certificate verification failure"""
--- a/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/conftest.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/conftest.py
@@ -2,9 +2,11 @@
 Conftest for changing targetRef field in policies
 """
 
+import time
+
 import pytest
 
-from testsuite.gateway import GatewayRoute, GatewayListener, Hostname, Exposer
+from testsuite.gateway import GatewayRoute, Hostname, Exposer, GatewayListener
 from testsuite.gateway.gateway_api.gateway import KuadrantGateway
 from testsuite.gateway.gateway_api.hostname import DNSPolicyExposer
 from testsuite.gateway.gateway_api.route import HTTPRoute
@@ -82,6 +84,15 @@ def client2(route2, hostname2):  # pylint: disable=unused-argument
 
 
 @pytest.fixture(scope="module")
+def authorization():
+    """
+    Override the authorization fixture to prevent the creation of an AuthPolicy.
+    This ensures no authentication is enforced during the test
+    """
+    return None
+
+
+@pytest.fixture(scope="module")
 def dns_policy2(blame, gateway2, module_label, dns_provider_secret, request):
     """DNSPolicy fixture for Gateway 2"""
     policy = DNSPolicy.create_instance(
@@ -93,7 +104,13 @@ def dns_policy2(blame, gateway2, module_label, dns_provider_secret, request):
     return policy
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module", autouse=True)
+def ensure_routes_created(route, route2):  # pylint: disable=unused-argument
+    """Ensures both routes exist and are ready before test executes"""
+    return
+
+
+@pytest.fixture(scope="module")
 def change_target_ref():
     """Function that changes targetRef of given policy"""
 
@@ -104,5 +121,6 @@ def change_target_ref():
 
         policy.modify_and_apply(_apply_target_ref)
         policy.wait_for_ready()
+        time.sleep(5)  # Extra wait to avoid inconsistent DNS issues, wait_for_ready isn't always enough
 
     return _change_targetref
--- a/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py
@@ -4,11 +4,21 @@ Test for changing targetRef field in AuthPolicy
 
 import pytest
 
-pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.authorino, pytest.mark.dnspolicy]
+
+
+@pytest.fixture(scope="module")
+def authorization(oidc_provider, gateway, cluster, blame, module_label, route):  # pylint: disable=unused-argument
+    """Overwrite the authorization fixture and attach it to the gateway"""
+    policy = AuthPolicy.create_instance(cluster, blame("authz"), gateway, labels={"testRun": module_label})
+    policy.identity.add_oidc("default", oidc_provider.well_known["issuer"])
+    return policy
 
 
 def test_update_auth_policy_target_ref(
-    gateway2, authorization, client, client2, auth, dns_policy, dns_policy2, change_target_ref
+    route2, gateway2, authorization, client, client2, auth, dns_policy, dns_policy2, change_target_ref
 ):  # pylint: disable=unused-argument
     """Test updating the targetRef of an AuthPolicy from Gateway 1 to Gateway 2"""
     response = client.get("/get", auth=auth)
@@ -27,3 +37,6 @@ def test_update_auth_policy_target_ref(
 
     response = client2.get("/get")
     assert response.status_code == 401
+
+    response = client.get("/get")
+    assert response.status_code == 200
--- a/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
@@ -1,0 +1,42 @@
+"""
+Test for changing targetRef field in DNSPolicy
+"""
+
+from time import sleep
+import pytest
+
+from testsuite.gateway import GatewayListener
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
+
+
+def test_update_dns_policy_target_ref(
+    gateway, gateway2, client, client2, dns_policy, change_target_ref
+):  # pylint: disable=unused-argument
+    """Test updating the targetRef of DNSPolicy from Gateway 1 to Gateway 2"""
+    assert gateway.refresh().is_affected_by(dns_policy)
+    assert not gateway2.refresh().is_affected_by(dns_policy)
+
+    response = client.get("/get")
+    assert not response.has_dns_error()
+    assert response.status_code == 200
+
+    response = client2.get("/get")
+    assert response.has_dns_error()
+
+    dns_ttl = gateway.get_listener_dns_ttl(GatewayListener.name)
+
+    change_target_ref(dns_policy, gateway2)
+
+    # Wait for records deletion/ttl expiration from the previous request
+    sleep(dns_ttl)
+
+    assert not gateway.refresh().is_affected_by(dns_policy)
+    assert gateway2.refresh().is_affected_by(dns_policy)
+
+    response = client2.get("/get")
+    assert not response.has_dns_error()
+    assert response.status_code == 200
+
+    response = client.get("/get")
+    assert response.has_dns_error()
--- a/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_ratelimitpolicy_target_ref.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_ratelimitpolicy_target_ref.py
@@ -6,16 +6,7 @@ import pytest
 
 from testsuite.kuadrant.policy.rate_limit import Limit, RateLimitPolicy
 
-pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
-
-
-@pytest.fixture(scope="module")
-def authorization():
-    """
-    Override the authorization fixture to prevent the creation of an AuthPolicy.
-    This ensures no authentication is enforced during the test
-    """
-    return None
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy, pytest.mark.limitador]
 
 
 @pytest.fixture(scope="module")
@@ -27,7 +18,7 @@ def rate_limit(cluster, blame, module_label, gateway, route):  # pylint: disable
 
 
 def test_update_ratelimit_policy_target_ref(
-    gateway2, rate_limit, client, client2, dns_policy, dns_policy2, change_target_ref
+    route2, gateway2, rate_limit, client, client2, dns_policy, dns_policy2, change_target_ref
 ):  # pylint: disable=unused-argument
     """Test updating the targetRef of a RateLimitPolicy from Gateway 1 to Gateway 2"""
     responses = client.get_many("/get", 2)
@@ -42,3 +33,6 @@ def test_update_ratelimit_policy_target_ref(
     responses = client2.get_many("/get", 2)
     responses.assert_all(status_code=200)
     assert client2.get("/get").status_code == 429
+
+    responses = client.get_many("/get", 3)
+    responses.assert_all(status_code=200)
--- a/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
@@ -1,0 +1,87 @@
+"""
+Test for changing targetRef field in TLSPolicy
+"""
+
+import pytest
+
+from testsuite.gateway import TLSGatewayListener
+from testsuite.gateway.gateway_api.gateway import KuadrantGateway
+from testsuite.gateway.gateway_api.hostname import StaticHostname
+from testsuite.httpx import KuadrantClient
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy, pytest.mark.tlspolicy]
+
+
+@pytest.fixture(scope="module")
+def gateway(request, cluster, blame, wildcard_domain, module_label):
+    """Create Gateway 1 with TLSGatewayListener"""
+    gateway_name = blame("gw")
+    gw = KuadrantGateway.create_instance(
+        cluster,
+        gateway_name,
+        {"app": module_label},
+    )
+    gw.add_listener(TLSGatewayListener(hostname=wildcard_domain, gateway_name=gateway_name))
+    request.addfinalizer(gw.delete)
+    gw.commit()
+    gw.wait_for_ready()
+    return gw
+
+
+@pytest.fixture(scope="module")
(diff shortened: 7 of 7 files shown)


testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,35 @@
...
+    This ensures no authentication is enforced during the test
+    """
+    return None
+
+
+def test_update_dns_policy_target_ref(
+    route2, tls_policy, tls_policy2, gateway, gateway2, dns_policy, client, client2, change_target_ref
+):  # pylint: disable=unused-argument
+    """Test updating the targetRef of an DNSPolicy from Gateway 1 to Gateway 2"""
+    assert gateway.refresh().is_affected_by(dns_policy)
```
averevki, line 23:
```suggestion
    assert gateway.wait_until(lambda obj: not obj.is_affected_by(dns_policy))
```
is more stable, it won't fail immediately if status is not ready yet

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,35 @@
...
+    assert not gateway2.refresh().is_affected_by(dns_policy)
+
+    response = client.get("/get")
```
averevki, line 26: you might benefit from `response.has_dns_error()` in this test

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_update_tls_policy_target_ref(
+    route2, tls_policy, tls_policy2, gateway, gateway2, client, client2, dns_policy, dns_policy2, change_target_ref
+):  # pylint: disable=unused-argument
+    """Test updating the targetRef of an TLSPolicy from Gateway 1 to Gateway 2"""
+    assert gateway.refresh().is_affected_by(tls_policy)
+    assert not gateway2.refresh().is_affected_by(tls_policy)
+
+    response = client.get("/get")
```
averevki, line 26: same here `response.has_cert_verify_error()`

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dns_tls_policies_target_ref.py
```diff
@@ -0,0 +1,74 @@
...
+    assert gateway2.wait_until(lambda obj: obj.is_affected_by(dns_policy))
+
+    client2 = hostname2.client()
+
+    response = client2.get("/get")
+    assert not response.has_cert_verify_error()
+    assert not response.has_dns_error()
+    assert response.status_code == 200
+
+    client2.close()
```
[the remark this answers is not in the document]
emmaaroche: For this, I chose to initialize client2 only after the targetRef change to ensure the TLS secret is available. When using a separate client2 fixture, errors occurred due to missing TLS certificates.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,60 @@
+"""
+Test for changing targetRef field in DNSPolicy
+"""
+
+import pytest
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy, pytest.mark.tlspolicy]
```
averevki, line 7: don't need tlspolicy mark here

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture(scope="module")
+def route2():
+    """
+    Override the route 2 fixture
+    """
+    return None
```
averevki, line 15: Any reason you work with only one route in this test? This should not be a problem to use both as routes are attached straight to gateway

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,100 @@
...
+    tls_policy2.delete()  # Delete tls_policy2 as it was only used to prevent initial TLS secret errors
+
+    assert gateway.wait_until(lambda obj: obj.is_affected_by(tls_policy))
+    assert gateway2.wait_until(lambda obj: not obj.is_affected_by(tls_policy))
+
+    response = client.get("/get")
+    assert not response.has_cert_verify_error()
+    assert response.status_code == 200
+
+    response = KuadrantClient(base_url=f"https://{hostname2.hostname}", verify=True).get("/get")
```
averevki, line 87:
```suggestion
    response = KuadrantClient(base_url=str(client2.base_url), verify=True).get("/get")
```
should save you few fixture requests

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
emmaaroche: The initial reason for using only one route was that, when using route2, a hostname would be created for gateway2/client2, causing the post-switch client2 request to return false instead of true. To work around this, I used the same route for both and switched it accordingly. However, upon further review, I realized that I may have been calling dns_policy2 before making this change. As a result, I’ve reverted to using both routes while making sure dns_policy2 is not created, and the test is passing!

testsuite/httpx/__init__.py
```diff
@@ -59,11 +59,19 @@ def has_error(self, error_msg: str) -> bool:
...
+        )
 
     def has_cert_verify_error(self):
         """True, if the result failed due to TLS certificate verification failure"""
-        return self.has_error("SSL: CERTIFICATE_VERIFY_FAILED")
+        return (
+            self.has_error("SSL: CERTIFICATE_VERIFY_FAILED")
+            or self.has_error("certificate verify failed")
+            or self.has_error("self-signed certificate")
+        )
```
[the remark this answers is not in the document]
emmaaroche: I did this initially for debugging purposes as was having some issues getting gateway2/client2 to return true to  `assert response.has_cert_verify_error()`, however I have undone this to be just `return self.has_error("SSL: CERTIFICATE_VERIFY_FAILED")` and it now seems to be passing fine.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,90 @@
...
+        labels={"app": module_label},
+    )
+    request.addfinalizer(policy.delete)
+    policy.commit()
+    policy.wait_for_ready()
+    return policy
+
+
+def test_update_tls_policy_target_ref(
+    route2, tls_policy, tls_policy2, gateway, gateway2, client, client2, dns_policy, dns_policy2, hostname2, hostname
```
averevki, line 65:
```suggestion
    route2, tls_policy, tls_policy2, gateway, gateway2, client, client2, dns_policy, dns_policy2
```

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py
```diff
@@ -4,11 +4,22 @@
 
 import pytest
 
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+from testsuite.tests.singlecluster.gateway.reconciliation.change_targetref.conftest import change_target_ref
```
averevki, line 8:
```suggestion
from .conftest import change_target_ref
```
will help if there are testsuite structure changes

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,43 @@
...
+
+    response = client.get("/get")
+    assert not response.has_dns_error()
+    assert response.status_code == 200
+
+    response = client2.get("/get")
+    assert response.has_dns_error()
+
+    dns_ttl = gateway.get_listener_dns_ttl(GatewayListener.name)
+    dns_policy.delete()
```
averevki, line 29: You don't need to delete policy you're testing with here :laughing: What I sent you is just an example how you can extract DNSRecod TTL and then wait for this time after you switch the gateway

Wonder how it didn't fail when I tried to execute it (already know, modify_and_apply uses apply() which creates the object if it doesn't exists, crazy)

averevki (dismissed):
LGTM!

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/conftest.py
```diff
@@ -93,16 +104,13 @@ def dns_policy2(blame, gateway2, module_label, dns_provider_secret, request):
     return policy
 
 
-@pytest.fixture(scope="session")
```
azgabur, line 96: Dont remove the fixture. If the scope is too broad you can change it to "module" so it fits. If you want to run the fixture body later in the test you can create a function and return it like it was before.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py
```diff
@@ -4,11 +4,22 @@
 
 import pytest
 
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+from .conftest import change_target_ref
```
azgabur, line 8: We want to prevent imports like these. See my first comment.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py
```diff
@@ -4,11 +4,22 @@
...
+
 pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
```
azgabur, line 10: Add `pytest.mark.authorino`

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,43 @@
...
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
+
+
+def test_update_dns_policy_target_ref(
+    route, route2, gateway, gateway2, client, client2, dns_policy
+):  # pylint: disable=unused-argument
+    """Test updating the targetRef of DNSPolicy from Gateway 1 to Gateway 2"""
+    assert gateway.wait_until(lambda obj: obj.is_affected_by(dns_policy))
+    assert gateway2.wait_until(lambda obj: not obj.is_affected_by(dns_policy))
```
azgabur, line 19: These waits should happen (or already happen) in creation fixtures, whats the reason for putting them here? If you want just to check if the gateway is affected or not just simple `gateway.is_affected_by(dns_policy)` should be enough. What do you think?

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
```diff
@@ -0,0 +1,43 @@
...
+
+    dns_ttl = gateway.get_listener_dns_ttl(GatewayListener.name)
+
+    change_target_ref(dns_policy, gateway2)
+
+    # Wait for records deletion/ttl expiration from the previous request
+    sleep(dns_ttl)
+
+    assert gateway.wait_until(lambda obj: not obj.is_affected_by(dns_policy))
+    assert gateway2.wait_until(lambda obj: obj.is_affected_by(dns_policy))
```
azgabur, line 36: Same thing here, no need for the `.wait_until()`

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_ratelimitpolicy_target_ref.py
```diff
@@ -5,19 +5,11 @@
 import pytest
 
 from testsuite.kuadrant.policy.rate_limit import Limit, RateLimitPolicy
+from .conftest import change_target_ref
 
 pytestmark = [pytest.mark.kuadrant_only, pytest.mark.dnspolicy]
```
azgabur, line 10: Add `pytest.mark.limitador`

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_update_tls_policy_target_ref(
+    route2, tls_policy, tls_policy2, gateway, gateway2, client, client2, dns_policy, dns_policy2
+):  # pylint: disable=unused-argument
+    """Test updating the targetRef of TLSPolicy from Gateway 1 to Gateway 2"""
+    tls_policy2.delete()  # Delete tls_policy2 as it was only used to prevent initial TLS secret errors
+
+    assert gateway.wait_until(lambda obj: obj.is_affected_by(tls_policy))
+    assert gateway2.wait_until(lambda obj: not obj.is_affected_by(tls_policy))
```
azgabur, line 71: No need for `.wait_until()` here

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,90 @@
...
+
+    response = client.get("/get")
+    assert not response.has_cert_verify_error()
+    assert response.status_code == 200
+
+    response = KuadrantClient(base_url=str(client2.base_url), verify=True).get("/get")
```
azgabur, line 77: The problem with `verify=True` is that in our environment where all our certificates are untrusted this will always fail. I would maybe try to check this differently but I dont know how.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,90 @@
...
+    policy.commit()
+    policy.wait_for_ready()
+    return policy
+
+
+def test_update_tls_policy_target_ref(
+    route2, tls_policy, tls_policy2, gateway, gateway2, client, client2, dns_policy, dns_policy2
+):  # pylint: disable=unused-argument
+    """Test updating the targetRef of TLSPolicy from Gateway 1 to Gateway 2"""
+    tls_policy2.delete()  # Delete tls_policy2 as it was only used to prevent initial TLS secret errors
```
azgabur, line 68: Is this really necessary?

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py
```diff
@@ -4,11 +4,22 @@
...
 
 
+@pytest.fixture(scope="module")
+def authorization(oidc_provider, gateway, cluster, blame, module_label, route):  # pylint: disable=unused-argument
+    """Overwrite the authorization fixture and attach it to the gateway"""
+    policy = AuthPolicy.create_instance(cluster, blame("authz"), gateway, labels={"testRun": module_label})
+    policy.identity.add_oidc("default", oidc_provider.well_known["issuer"])
```
azgabur, line 17: Small tip: If you dont want to bother with `auth` fixture here, you can add "anonymous" identity which does not require auth :) But this is correct way too :+1:

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
emmaaroche: Yeah, I can remove them and add gateway.is_affected_by(dns_policy) instead, I saw them in some other dns/tls tests which is why I added them into this test

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
emmaaroche: Yes, I was having issues with getting the tests to run due to tls secret errors without setting up a tls_policy2 so when discussing with Alex this seems to be the best work around for that.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
emmaaroche: I encountered challenges in getting this check to assert the cert errors when it should have been, and after a call with Alex, this approach appeared to be the most reliable workaround. That said, I'm happy to explore alternative solutions if needed.

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/conftest.py
```diff
@@ -104,5 +115,6 @@ def _apply_target_ref(apiobj):
 
         policy.modify_and_apply(_apply_target_ref)
         policy.wait_for_ready()
+        time.sleep(5)
```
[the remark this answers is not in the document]
emmaaroche: Yes you are right, I will add a comment for context to this. The reason I added is for some reason, when testing with gcp and azure credentials, if I do not add this sometimes it does not wait for the switch to complete and gateway2/client2 requests fail with a hostname not known error.

testsuite/gateway/gateway_api/gateway.py
```diff
@@ -106,6 +106,23 @@ def get_tls_cert(self, hostname):
         )
         return tls_cert
 
+    def get_tls_secret(self, hostname):
```
azgabur, line 109: Please make method `get_tls_cert()` call this method to get its secret to prevent duplicated code

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,90 @@
...
+    assert not response.has_cert_verify_error()
+    assert response.status_code == 200
+
+    response = KuadrantClient(base_url=f"https://{hostname2.hostname}", verify=False).get("/get")
+    assert response.has_error("[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol")
+
+    change_target_ref(tls_policy, gateway2)
+
+    assert gateway2.refresh().is_affected_by(tls_policy)
+    assert not gateway.refresh().is_affected_by(tls_policy)
```
azgabur, line 76: The sudden change of order is bugging me a little :laughing:

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,90 @@
...
+    response = custom_client(hostname2.hostname, gateway2).get("/get")
+    assert not response.has_cert_verify_error()
+    assert response.status_code == 200
+
+    # Delete TLS secret to verify gateway1 no longer serves valid TLS traffic
+    tls_secret = gateway.get_tls_secret(hostname.hostname)
+    assert tls_secret is not None, "TLS secret was not found after targetRef change"
+    secret_name = tls_secret.model.metadata.name
+    namespace = tls_secret.model.metadata.namespace
+    gateway.cluster.do_action("delete", "secret", secret_name, "-n", namespace)
```
azgabur, line 87: You can just call `tls_secret.delete()` and it will throw exception if it wont find it. And its also less verbose :+1:

azgabur (commented):
I think after addressing these last changes I will approve this PR. Looks good :+1:

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/conftest.py
```diff
@@ -104,5 +121,6 @@ def _apply_target_ref(apiobj):
 
         policy.modify_and_apply(_apply_target_ref)
         policy.wait_for_ready()
+        time.sleep(5)  # Extra wait to avoid inconsistent DNS issues, wait_for_ready isn't always enough
```
azgabur, line 124: If it is needed and it gets merged, please add the occurance to https://github.com/Kuadrant/testsuite/issues/660
No need to address this now

azgabur (approved):
Good job :+1: 
Please wait for a second review from @averevki to merge

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
averevki: these statuses won't be set in time for tests sometimes, that's why it is combined with a wait_until in tests where it is used

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py
```diff
@@ -0,0 +1,87 @@
...
+    """Test updating the targetRef of TLSPolicy from Gateway 1 to Gateway 2"""
+    assert gateway.refresh().is_affected_by(tls_policy)
+    assert not gateway2.refresh().is_affected_by(tls_policy)
+
+    response = custom_client(hostname.hostname, gateway).get("/get")
+    assert not response.has_cert_verify_error()
+    assert response.status_code == 200
+
+    response = KuadrantClient(base_url=f"https://{hostname2.hostname}", verify=False).get("/get")
+    assert response.has_error("[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol")
```
averevki, line 71: This request will be backing off based on the error returned
https://github.com/Kuadrant/testsuite/blob/f518b3e060bddf403004fab9f2563f1070d2f38d/testsuite/httpx/__init__.py#L46-L54

I liked the previous approach more. Previously it was verifying if valid/invalid certificate was provided with the request. Error now to me looks like https wasn't configured properly

azgabur: And thats indeed what we want to test, that the https is not working due to TlsPolicy not being applied to that gateway/hostname. Its true the backoff will be triggered and because of that there will be additional wait, but I think we can redo that behaivor in https://github.com/Kuadrant/testsuite/issues/660 as I think configuring the client to expect some kind of error is beneficial when testing "non happy path error" scenarios.

The verification of the certificate is done when we expect it to return 200 with the custom_client fixture,

testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_dnspolicy_target_ref.py
azgabur: I think at least the first wait_until is not needed due to it being called in `commit` fixture, the subsequent ones are indeed needed yes.

PR #643, opened 2025-02-27
Add slow_loadbalancers and refactor MockserverBackend

azgabur:
After trying testing on ROSA I noticed tests failing from loadbalancers not being ready due to them being attached to service but not yet started in AWS. I have maybe not a nice solution but I dont have any better idea to just add sleep in wait_for_ready in Gateway and Service methods. I put a toggle in settings.yaml in which this can be enabled as this happens only on AWS and it is pointless to have such long wait in other scenarios.

Additionally I refactored MockserverBackend to be better + factor out the hardcoded mockserver image.


testsuite/backend/mockserver.py
```diff
@@ -35,9 +35,7 @@ def commit(self):
         )
         self.service.commit()
 
-    def wait_for_ready(self, timeout=300):
-        """Waits until Deployment is marked as ready"""
-        success = self.service.wait_until(
-            lambda obj: "ip" in self.service.refresh().model.status.loadBalancer.ingress[0], timelimit=timeout
-        )
-        assert success, f"Service {self.name} did not get ready in time"
+    def wait_for_ready(self, timeout=60 * 5):
```
trepel, line 38: where do you actually call this method? I can see that you removed the `self.deployment.wait_for_ready()` call (was on line 26) and move it into this method but I am struggling to see where this method is actually used.

azgabur: The `MockserverBackend` is used only this this one tests and the `wait_for_ready()` is called here: https://github.com/azgabur/testsuite/blob/refactor-mockserver/testsuite/tests/singlecluster/gateway/dnspolicy/health_check/test_additional_headers.py#L48

trepel: I must have been blinded by that bloody sun shining directly at my monitor :) All clear now, thanks!

averevki (commented):
Not very pretty. We try to use settings preferably only for any external information testsuite needs. Surely there is a way to detect if cluster is running on ROSA, right? Don't want to block though

(2 unrelated changes in the single commit :goat:)

Issue #647, opened 2025-03-11
Add custom unassigned return code test when issue fixed


averevki:
If after https://github.com/Kuadrant/kuadrant-operator/issues/1022 / https://issues.redhat.com/browse/CONNLINK-287 is solved, and it will be possible to return custom [unassigned](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) return codes with authpolicy, create test returning this custom return code

Tests for valid return codes located at `testsuite/tests/singlecluster/authorino/response/test_deny_with.py`

PR #648, opened 2025-03-11
Change to valid return codes in deny_with tests

averevki:
Link moved to a separate issue https://github.com/Kuadrant/testsuite/issues/647


trepel (approved):
If I got it right you use standard (valid) return codes in the test. We can move to using non-standard ones but only once it is decided what the expected behavior should be (you linked the issue to watch)
LGTM

azgabur (approved):

PR #649, opened 2025-03-11
Add backoff to the health-check tests

averevki:
Some healthcheck tests failing inside pods, I'm not sure why they are failing yet. Trying to add a backoff for each failing action to see if resource creation timing is the problem

### DIFF
--- a/testsuite/kuadrant/policy/dns.py
+++ b/testsuite/kuadrant/policy/dns.py
@@ -3,6 +3,8 @@
 from dataclasses import dataclass
 from typing import Optional, Literal
 
+import openshift_client as oc
+
 from testsuite.gateway import Referencable
 from testsuite.kubernetes import KubernetesObject
 from testsuite.kubernetes.client import KubernetesClient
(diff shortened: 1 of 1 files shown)


testsuite/kuadrant/policy/dns.py
```diff
@@ -93,10 +100,23 @@ def set_health_check(self, health_check: HealthCheck):
...
+
     def get_dns_health_probe(self) -> DNSHealthCheckProbe:
         """Returns DNSHealthCheckProbe object for the created DNSPolicy"""
+        dns_record = self._get_dns_record()
         with self.context:
-            dns_probe = self.get_owned("dnsrecords.kuadrant.io")[0].get_owned("DNSHealthCheckProbe")[0]
+            assert dns_record.wait_until(
+                lambda obj: len(obj.get_owned("DNSHealthCheckProbe")) > 0
+            ), "The corresponding DNSHealthCheckProbe object wasn't created in time"
+            dns_probe = dns_record.get_owned("DNSHealthCheckProbe")[0]
```
trepel, line 119: This is definitely better than `dns_probe = self.get_owned("dnsrecords.kuadrant.io")[0].get_owned("DNSHealthCheckProbe")[0]` I still think that it is not optimal, what if the DNSHealthCheckProbe appears and the disappear quickly. In such case it would pass the assert on 118 but would fail on the 119 because would not be available.

But let's merge this and see if this helps (it should). There is a good chance that this is good enough for our use  case.

averevki: I don't think a probe object can be created and then disappear quickly without any client side changes. Never saw that happening at least. 

What I have encountered is a probe object not created at all when it should've been. With this PR I just want to see if this is not the problem with nightly runs and just some responsiveness under the load issue

trepel (approved):
Validated by `make testsuite/tests/singlecluster/gateway/dnspolicy/health_check`. All tests passed.

Issue #631, opened 2025-02-14
Make .apply() fail with exception


azgabur:
Currently when doing changes to APIobject-s via the `.apply()` or `.modify_and_apply()` methods these can fail silently. This does not apply to class methods with `@modify` decorator as those will fail with assertion error (https://github.com/Kuadrant/testsuite/blob/main/testsuite/kubernetes/__init__.py#L90)

To fix this there are two solutions:

a) Rewrite all `.apply()` calls to class methods with `@modify` decorator
b) Override `.apply()` or `.modify_and_apply()` methods in class `KubernetesObject` to fail with assert if unsuccessful

PR #654, opened 2025-03-21
Add assertions to apply() and modify_and_apply() methods

emmaaroche:
**Summary**
This PR adds assertions to the `.apply()` and `.modify_and_apply()` methods to ensure errors are raised when resource updates fail, preventing silent failures.

**Description of changes**
Following solution (b) from issue #631, the `.apply()` and `.modify_and_apply()` methods have been overridden in `KubernetesObject` to raise an assertion error when:

- A change was expected but not applied (i.e. no modification occurred)
- The apply command itself failed (returned a non-zero exit code)

**Related Issue**
Closes [#631](https://github.com/Kuadrant/testsuite/issues/631)

### DIFF
--- a/testsuite/kubernetes/__init__.py
+++ b/testsuite/kubernetes/__init__.py
@@ -34,6 +34,23 @@ class KubernetesObject(APIObject, LifecycleObject):
         self._committed = True
         return self.refresh()
 
+    def apply(self, *args, **kwargs):
+        """
+        Overrides apply() to fail with assertion error if unsuccessful
+        """
+        result = super().apply(*args, **kwargs)
+        assert result.status() == 0, f"Apply returned non-zero exit code for {self.kind()} {self.name()}"
+        return result
+
+    def modify_and_apply(self, *args, **kwargs):
+        """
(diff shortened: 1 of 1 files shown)


testsuite/kubernetes/__init__.py
azgabur, line 1: This will work if we expect the library API wont change, but to make this future-proof I would use `**kwargs` as the only argument which will pass all arguments to parent function. That way if the number, order of arguments change, this override functions dont need to change.

testsuite/kubernetes/__init__.py
```diff
@@ -34,6 +34,24 @@ def commit(self):
         self._committed = True
         return self.refresh()
 
+    def apply(self, cmd_args=None):
+        """
+        Overrides apply() to fail with assertion error if unsuccessful
+        """
+        result, status = self.modify_and_apply(lambda _: True, retries=0, cmd_args=cmd_args)
```
azgabur, line 41: You should call `super().apply(**kwags)` here

azgabur (approved):
LGTM!

PR #658, opened 2025-03-28
Change the listener hostname directly - fix test_change_listener test

azgabur:
## Description
Gateway cant have 0 listeners. This change wont change the test case as this is what I wanted to test in the first place. ie. Changing the hostname of the single listener.

## Verification steps
Run the test
```sh
make testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_listeners_dns.py
```


trepel (approved):
Test passed for me against Kuadrant 1.1.0-rc2

Issue #571, opened 2024-11-06
Secondary AuthPolicy/RateLimitPolicy attached to same HttpRoute/Gateway collision test


azgabur:
**Summary**

We want to check that only one AuthPolicy/RateLimitPolicy can be attached to a HttpRoute at any given time. The first one attached should be the one working.

**Documentation**
...

**Additional context**

We already created test for AuthConfig collision: testsuite/tests/singlecluster/authorino/operator/clusterwide/test_wildcard_collision.py

**Test cases**

Test that two AuthPolicy on same HTTPRoute will result in one having status "Conflicted"
Same for RateLimitPolicy

crstrn13:
Issue addressed in #661

PR #661, opened 2025-04-01
Collision tests for policies.

crstrn13:
This PR aims to fix issue: #571 

Added collision test for both auth policies and rate limits when attached to the same target.

### DIFF
--- a/testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
+++ b/testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
@@ -1,13 +1,17 @@
+"""Test auth policies collisions when attached to the same target"""
+
 import pytest
 
 from testsuite.httpx.auth import HeaderApiKeyAuth
 from testsuite.kuadrant.policy import has_condition
 from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
 
+
 @pytest.fixture(scope="module")
 def api_key(create_api_key, module_label):
     """Creates API key Secret."""
     return create_api_key("api-key", module_label, "api_key_value")
+
 
 # pylint: disable = unused-argument
 @pytest.fixture(scope="module")
@@ -19,6 +23,7 @@ def authorization(request, cluster, blame, module_label, route, label, oidc_prov
     auth.identity.add_oidc("first", oidc_provider.well_known["issuer"])
     return auth
 
+
 @pytest.fixture(scope="module")
 def authorization2(request, cluster, blame, module_label, route, label, api_key):
     """Create a RateLimitPolicy with a basic limit with target coming from test parameter"""
@@ -27,6 +32,7 @@ def authorization2(request, cluster, blame, module_label, route, label, api_key)
     auth = AuthPolicy.create_instance(cluster, blame("sp"), target_ref, labels={"testRun": label})
     auth.identity.add_api_key("second", selector=api_key.selector)
     return auth
+
 
 @pytest.fixture(scope="module")
 def auth2(api_key):
@@ -42,16 +48,10 @@ def commit(request, authorization, authorization2):
         policy.commit()
         policy.wait_for_ready()
 
+
 @pytest.mark.parametrize("authorization", ["gateway", "route"], indirect=True)
-def test_collision(client, authorization, authorization2, auth, auth2):
+def test_collision_auth_policy(client, authorization, authorization2, auth, auth2):
     """Test first policy is being overridden when another policy with the same target is created."""
-    assert authorization.wait_until(
-        has_condition(
-            "Enforced",
-            "False",
-            "Overridden",
-            "AuthPolicy is overridden"
-        )
-    )
+    assert authorization.wait_until(has_condition("Enforced", "False", "Overridden", "AuthPolicy is overridden"))
     assert client.get("/get", auth=auth).status_code == 401
     assert client.get("/get", auth=auth2).status_code == 200
--- a/testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
+++ b/testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
@@ -1,0 +1,57 @@
+import pytest
+
+from testsuite.httpx.auth import HeaderApiKeyAuth
+from testsuite.kuadrant.policy import has_condition
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+
+@pytest.fixture(scope="module")
+def api_key(create_api_key, module_label):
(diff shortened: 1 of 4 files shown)


testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
```diff
@@ -0,0 +1,57 @@
...
+
+    auth = AuthPolicy.create_instance(cluster, blame("fp"), target_ref, labels={"testRun": label})
+    auth.identity.add_oidc("first", oidc_provider.well_known["issuer"])
+    return auth
+
+
+@pytest.fixture(scope="module")
+def authorization2(request, cluster, blame, module_label, route, label, api_key):
+    """Create a RateLimitPolicy with a basic limit with target coming from test parameter"""
+    target_ref = request.getfixturevalue(getattr(request, "param", "route"))
```
averevki, line 30: I think paremetrization here won't work as you call it only for the first authpolicy `@pytest.mark.parametrize("authorization", ["gateway", "route"], indirect=True)`

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
```diff
@@ -0,0 +1,57 @@
...
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization, authorization2):
+    """Commits AuthPolicy after the target is created"""
+    for policy in [authorization, authorization2]:  # Forcing order of creation.
+        request.addfinalizer(policy.delete)
+        policy.commit()
+        policy.wait_for_ready()
+
+
+@pytest.mark.parametrize("authorization", ["gateway", "route"], indirect=True)
```
averevki, line 52: Issue mentions only the httproute collision tests.

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
```diff
@@ -0,0 +1,57 @@
...
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+
+
+@pytest.fixture(scope="module")
+def api_key(create_api_key, module_label):
+    """Creates API key Secret."""
+    return create_api_key("api-key", module_label, "api_key_value")
+
+
+# pylint: disable = unused-argument
```
averevki, line 16: Disables placed like that will likely cover the whole file. Try to use them on the same line you need disable on

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
```diff
@@ -0,0 +1,57 @@
+"""Test auth policies collisions when attached to the same target"""
+
+import pytest
+
+from testsuite.httpx.auth import HeaderApiKeyAuth
+from testsuite.kuadrant.policy import has_condition
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+
```
averevki, line 8: See if you can place appropriate `pytestmarks` for both of these files

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
averevki, line 1: we already have authorino and limitador tests separated by tests/singlecluster/authorino and tests/singlecluster/limitador paths. I think it would be better to not thread them further

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
crstrn13: oh, you are totally right. Let me see what I can do about it.

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
crstrn13: ok, that would make things much easier.

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
crstrn13: added.

testsuite/tests/singlecluster/authorino/collisions/auth_policy/test_collisions.py
crstrn13: moved.

testsuite/tests/singlecluster/authorino/collisions/test_collisions.py
```diff
@@ -0,0 +1,62 @@
...
+    return HeaderApiKeyAuth(api_key)
+
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization, authorization2):
+    """Commits both AuthPolicies after the target is created"""
+    for policy in [authorization, authorization2]:  # Forcing order of creation.
+        request.addfinalizer(policy.delete)
+        policy.commit()
+        policy.wait_for_ready()
```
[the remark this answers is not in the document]
crstrn13: Could you point me to the exact test? In the merge, overrides tests I believe the behavior is similar, but I could be mistaken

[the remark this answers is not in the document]
azgabur: Its this test: https://github.com/Kuadrant/testsuite/blob/4eb61a64ced5aa2d6210c6f151988500ef8bf500/testsuite/tests/singlecluster/authorino/operator/clusterwide/test_wildcard_collision.py

azgabur (approved):
LGTM!

testsuite/tests/singlecluster/authorino/collisions/test_collisions.py
```diff
@@ -0,0 +1,62 @@
...
+    """Returns the test target(gateway or route)"""
+    return request.getfixturevalue(request.param)
+
+
+@pytest.fixture(scope="module")
+def authorization(
+    request, gateway, route, cluster, blame, target, label, oidc_provider
+):  # pylint: disable=unused-argument
+    """Create an AuthPolicy with either gateway or route as target reference"""
+    auth = AuthPolicy.create_instance(cluster, blame("fp"), target, labels={"testRun": label})
```
azgabur, line 29: I forgot this before merging. This is on me...
The standalone tests are not passing.
You cant create AuthPolicy here as Authorino knows only AuthConfig.

azgabur: Quick fix for this would be to add `kuadrant_only` in marks or use `authorization` fixture already provided - which handles the changing from AuthPolicy to AuthConfig depending on if the tests are run as standalone or no.

Issue #664, opened 2025-04-25
Investigate failing change_targetref tests


emmaaroche:
**The following change_targetref tests are failing consistently on the nightly's:**

- testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py

- testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_authpolicy_target_ref.py

**Error logs:** 

test_update_tlspolicy_target_ref.py
```
E   AssertionError: assert False
E    +  where False = has_error('[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol')
E    +    where has_error = <testsuite.httpx.Result object at 0x7f1d7742fb10>.has_error
```
test_update_authpolicy_target_ref.py

```
E   AssertionError: authpolicy did not get fully enforced in time
```

emmaaroche:
Update - since the merge of [#665](https://github.com/Kuadrant/testsuite/pull/665) these tests are passing on the nightly's. Will keep reviewing these tests on report portal for the coming days just to make sure they continue to pass. If failure occurs, will investigate if it is related or unrelated to these changes.

Issue #660, opened 2025-03-31
Investigate smarter waits


azgabur:
There are currently 17 invocations of `sleep` in our testsuite which is enough for investigation on how many are actually needed to be done. 
The replacements could be some sort of smarter wait. Either a backoff or other more reliable checking if the resource is in expected state. 

Calling `sleep` should be a last resort and the time value must be chosen with a reason, not at random.

Current places with `sleep`:
```
./testsuite/utils.py (1 occurrences):
211:    sleep(dns.resolver.resolve(hostname).rrset.ttl)  # type: ignore
./testsuite/gateway/envoy/__init__.py (1 occurrences):
57:        time.sleep(3)  # or some reason wait_for_ready is not enough, needs more investigation
./testsuite/gateway/gateway_api/gateway.py (1 occurrences):
75:            sleep(60)
./testsuite/tests/multicluster/load_balanced/test_change_default_geo.py (1 occurrences):
28:    sleep(300)  # wait for DNS propagation on providers
./testsuite/tests/multicluster/load_balanced/test_change_strategy.py (1 occurrences):
35:    sleep(300)  # wait for DNS propagation on providers
./testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_listeners_dns.py (1 occurrences):
31:    sleep(wildcard_domain_ttl)
./testsuite/tests/singlecluster/gateway/reconciliation/listeners/test_gateway_basic_listeners.py (1 occurrences):
43:    sleep(second_domain_ttl)
./testsuite/tests/singlecluster/gateway/dnspolicy/test_dnspolicy_removal.py (1 occurrences):
45:    sleep(dns_ttl)
./testsuite/tests/singlecluster/limitador/test_multiple_iterations.py (1 occurrences):
28:        sleep(10)
./testsuite/tests/singlecluster/identical_hostnames/rlp/test_rlp_on_routes.py (1 occurrences):
71:    sleep(15)
./testsuite/tests/singlecluster/identical_hostnames/rlp/test_rlp_on_gw_and_route.py (1 occurrences):
80:    sleep(15)
./testsuite/tests/singlecluster/authorino/authorization/opa/external_registry/test_cache.py (2 occurrences):
24:    sleep(2)  # waits for cache to reset because of ttl=1
44:    sleep(2)
./testsuite/tests/singlecluster/authorino/caching/metadata/test_ttl.py (1 occurrences):
31:    sleep(cache_ttl)
[...]

PR #665, opened 2025-04-25
Update failing change_targetref tests

emmaaroche:
## Description

This PR updates the failing `change_targetref` tests as described in issue #664.

**Changes:**
- Solution for now is to add `time.sleep(10)` to both tests, comments in code explain reasoning. In the future, when issue #660 is addressed, there could be a better solution in place of this.

- Added `has_tls_error() `method to `testsuite/httpx/__init__.py`

- Updated assertions in `testsuite/tests/singlecluster/gateway/reconciliation/change_targetref/test_update_tlspolicy_target_ref.py` to use `assert response.has_tls_error()` instead of directly checking for a specific SSL error message via `response.has_error("[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol")`.

### DIFF
--- a/testsuite/httpx/__init__.py
+++ b/testsuite/httpx/__init__.py
@@ -65,6 +65,10 @@ class Result:
             or self.has_error("No address associated with hostname")
         )
 
+    def has_tls_error(self):
+        """True, if the result failed due to TLS failure"""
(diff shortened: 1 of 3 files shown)


averevki (approved):
LGTM. Let's see if this helps

PR #666, opened 2025-04-28
Fix/refactor authorino metadata tests

averevki:
These changes aimed to improve nightlies execution of cached authorino tests. 
- Refactor authorino metadata test to verify 200 response code
- Fix response extract utility function to return valid weakget object

### Verification steps

```sh
make testsuite/tests/singlecluster/authorino/caching/metadata/
```

### DIFF
--- a/testsuite/tests/singlecluster/authorino/caching/metadata/test_caching.py
+++ b/testsuite/tests/singlecluster/authorino/caching/metadata/test_caching.py
@@ -22,12 +22,15 @@ def test_cached(client, auth, module_label, mockserver):
         - both requests return the same result
         - only single external value evaluation occurs. The second response contains cached (in-memory) value
     """
-    response = client.get("/get", auth=auth)
-    data = extract_response(response)[module_label]["uuid"] % None
-    response = client.get("/get", auth=auth)
-    cached_data = extract_response(response)[module_label]["uuid"] % None
-
-    assert cached_data is not None
+    response1 = client.get("/get", auth=auth)
+    assert response1.status_code == 200
+    data = extract_response(response1)[module_label]["uuid"] % None
     assert data is not None
+
+    response2 = client.get("/get", auth=auth)
+    assert response2.status_code == 200
+    cached_data = extract_response(response2)[module_label]["uuid"] % None
+    assert cached_data is not None
+
     assert data == cached_data
     assert len(mockserver.retrieve_requests(module_label)) == 1
--- a/testsuite/tests/singlecluster/authorino/caching/metadata/test_ttl.py
+++ b/testsuite/tests/singlecluster/authorino/caching/metadata/test_ttl.py
@@ -9,30 +9,28 @@ from testsuite.utils import extract_response
 
 pytestmark = [pytest.mark.authorino]
 
-
-@pytest.fixture(scope="module")
-def cache_ttl():
-    """Returns TTL in seconds for Cached Metadata"""
-    return 5
+CACHE_TTL = 5
 
 
 @pytest.fixture(scope="module")
-def authorization(authorization, module_label, expectation_path, cache_ttl):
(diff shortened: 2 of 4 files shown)


testsuite/tests/singlecluster/authorino/caching/metadata/test_caching.py
```diff
@@ -22,12 +22,15 @@ def test_cached(client, auth, module_label, mockserver):
         - both requests return the same result
         - only single external value evaluation occurs. The second response contains cached (in-memory) value
     """
-    response = client.get("/get", auth=auth)
-    data = extract_response(response)[module_label]["uuid"] % None
-    response = client.get("/get", auth=auth)
-    cached_data = extract_response(response)[module_label]["uuid"] % None
+    response1 = client.get("/get", auth=auth)
```
trepel, line 25: why this usage of two variables (`response1` and `response2`) instead of using just `response` variable here, is that due to easier debugging?

testsuite/tests/singlecluster/authorino/caching/metadata/test_ttl.py
```diff
@@ -9,30 +9,28 @@
 
 pytestmark = [pytest.mark.authorino]
 
-
-@pytest.fixture(scope="module")
-def cache_ttl():
-    """Returns TTL in seconds for Cached Metadata"""
-    return 5
+CACHE_TTL = 5
```
trepel, line 12: Just to educate myself, what is the reason for using const instead of fixture here? I assume that fixture processing requires a bit more CPU/RAM and const works equally well. Was there any other reason?

testsuite/utils.py
```diff
@@ -125,7 +125,7 @@ def extract_response(response, header="Simple", key="data"):
 
     # Returning None if content is empty, this typically happens for non-200 responses
     if len(response.content) == 0:
-        return weakget(None)
+        return weakget({})
```
trepel, line 128: I don't understand this change either. I did some reading but it would behave the same afaict. Could you please educate me why using '{}' is better here?

trepel (approved):
Very nice! I have a few questions (mosty to educate myself so no blockers for merging) but the code looks good to me. Also the tests passed for me.
LGTM

testsuite/tests/singlecluster/authorino/caching/metadata/test_ttl.py
azgabur: I thinking in general in our testsuite we use global constants if they are used just in one test file and constant in fixture if it would be used in multiple files. At least thats how I try to write code.

testsuite/utils.py
azgabur: ~I think that this is the "default return" from the weakget if it does not find the key. I would be careful changing this as some tests maybe depend on it.~ No no, this is the object being checked so if we want to always return "missing key" from the weak get it should be an empty dictionary. I know that in python `{}` is by default a dictionary but maybe for some it can be confusing. Maybe using `dict()` would be better option? In any case its okay.

Ignore this:
This always had a problem when you **expect** `None` being returned but you can actually override it with the `% something` syntax. I used this for example in this test: https://github.com/Kuadrant/testsuite/blob/12948ee5e3836919c2ad5937f07cb1941827a94d/testsuite/tests/singlecluster/authorino/identity/extended_properties/test_overwriting.py

azgabur (approved):
Extra test cases and nice weakget fix :+1:

testsuite/tests/singlecluster/authorino/caching/metadata/test_caching.py
averevki: I changed it to make it similar to the other test, not cached. But overall, I try to avoid reassigning variables in Python. I think it can cause some unintended bugs

PR #672, opened 2025-05-05
Added more resilience on update policies.

crstrn13:
Added check function that will assert the generation has been incremented before confirming it the policy enforcement.

### DIFF
--- a/testsuite/tests/singlecluster/authorino/identity/anonymous/test_anonymous_identity.py
+++ b/testsuite/tests/singlecluster/authorino/identity/anonymous/test_anonymous_identity.py
@@ -2,6 +2,7 @@
 
 import pytest
 
+
 pytestmark = [pytest.mark.authorino]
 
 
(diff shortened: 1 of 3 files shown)


testsuite/tests/singlecluster/authorino/identity/anonymous/test_anonymous_identity.py
```diff
@@ -31,8 +37,9 @@ def test_anonymous_identity(client, auth, authorization):
     response = client.get("/get")
     assert response.status_code == 401
 
+    generation = authorization.generation
     authorization.identity.add_anonymous("anonymous")
-    authorization.wait_for_ready()
+    authorization.wait_until(has_observed_generation(generation + 1))
```
azgabur, line 42: You still need the `wait_for_ready` as just waiting for Authorino to pick up the the AuthPolicy is not enough. The observedGeneration can be incremented but the Policy can be in status "Not Enforced". `wait_for_ready` makes sure the Policy is also "Enforced"

azgabur, line 42: I would put the wait for `generation == observedGeneration` inside `wait_for_ready` in class `Policy` as it conceptually fits there.

azgabur (approved):

PR #679, opened 2025-05-22
Add CA issuer for multicluster tests

azgabur:
Currently multicluster tests use as TLS issuer the Letsencrypt. There are connected quotas and long delays for the issuing process making the tests take much longer time to finish. 

Additionally the Letsencrypt issuer should only be used in test `test_external_ca` as in any other case we do not care about the type of issuer.

Adding a known CA that will issue test certificates we can simplify TlsPolicy testing. Maybe in the future the selfsinged-issuer can be fully replaced by this CA issuer and the logic in `Hostname` class that fetches certificate from cluster at runtime will no longer be needed. For now I only changed the multicluster tests where it was mainly needed.

A testing CA is part of helm-install since https://github.com/azgabur/kuadrant-helm-install/commit/67dafa7ff5fb6cde64a5c6d1f4c0fc35513ffde4


testsuite/certificates.py
azgabur, line 1: I found a bug in this test as well. It was checking for signature algorithm of signing CA not algorithm of public key of the certificate.

azgabur:
> Will it require any change inside the testsuite pipelines?

Nope, I added the CA in https://github.com/azgabur/kuadrant-helm-install/commit/b9128fbd843f5000b01076c0df82f0aa428eaeec to helm install and pipeline always uses most recent version. Only additional thing to do is to add the CA issuer on clusters that had not been deployed with the newer version of helm chart. I will inform on Slack

Issue #673, opened 2025-05-05
Add JWT plain identity test


averevki:
**Summary**

Basic [plain identity test](https://github.com/Kuadrant/testsuite/pull/366) we have in our testsuite covers the functionality of plain identity, but doesn't test any real-world scenario that a Kuadrant customer might want. 

Create the plain identity test with authentication handled in Envoy leveraging the Envoy JWT Authentication filter (decoded JWT injected as 'metadata_context' into the response). 

**Documentation**

Kuadrant user guide on JWT authn with plain identity https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/envoy-jwt-authn-and-authorino.md.

Path to existing test for plain identity inside the testsuite: `testsuite/tests/singlecluster/authorino/identity/plain/test_plain_identity.py` (for the reference)

**Additional context**

* The Kuadrant user guide for plain identity might have covered more functionality than we need for the test. 
* Test plain identity with three types of asserts: user have assigned a role, user doesn't have role assigned and with invalid token.
* Parametrize test for both Kuadrant and standalone Authorino with pytest marks

**Test cases**

Test case from the Kuadrant user guide

PR #686, opened 2025-06-10
JWT plain identity test

silvi-t:
## Description
This PR addresses part of issue #673. It tests two cases of authentication in Envoy using the Envoy JWT authentication filter. The first case is when the user has a role assigned, and the second case is when the user does not have a role assigned.

### DIFF
--- a/testsuite/tests/singlecluster/authorino/identity/plain/test_jwt_plain_identity.py
+++ b/testsuite/tests/singlecluster/authorino/identity/plain/test_jwt_plain_identity.py
@@ -1,0 +1,91 @@
+"""
+Plain identity test with automation handled by Envoy leveraging the Envoy JWT Authentication filter
+(decoded JWT injected as 'metadata_context' into the response).
+https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/envoy-jwt-authn-and-authorino.md
+"""
+
+import pytest
+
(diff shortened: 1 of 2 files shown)


testsuite/tests/singlecluster/authorino/identity/plain/test_jwt_plain_identity.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.fixture(scope="module")
+def authorization(authorization, realm_role, blame):
+    """
+    Setup AuthConfig to retrieve identity from given path.
+    Adds rule, that requires user to be part of realm_role to be allowed access.
+    """
+    authorization.identity.add_plain(
+        "plain", "context.metadata_context.filter_metadata.envoy\\.filters\\.http\\.jwt_authn|verified_jwt"
+    )
+    authorization.responses.add_simple("context")
```
averevki, line 41: You might benefit from adding a little bit bigger path here and have fewer keys used with the extract_response function, it will make it more readable I'd say
```suggestion
    authorization.responses.add_simple("context.metadata_context.filter_metadata.envoy.filters")
```

azgabur (commented):
Not happy about creating new class but it seems to be how the different Envoy configs are done in testsuite so I am okay with it

PR #692, opened 2025-06-20
Add JWT test implementing user story from user-guides

silvi-t:
## Description
This PR implements the user story from issue #689 into the testsuite.
As a user you can access from anywhere if you have a realm role assigned. But if user does not have assigned realm role than the user can access only from allowed path parameter "/anythig/SK/" or when accessing through "/anything/global" path parameter that does not trigger external authorization by Envoy.

### DIFF
--- a/testsuite/gateway/envoy/config.py
+++ b/testsuite/gateway/envoy/config.py
@@ -114,7 +114,7 @@ class EnvoyConfig(ConfigMap):
             "route_config"
         ]["virtual_hosts"][0]["routes"]
         for host in virtual_hosts:
-            if host["match"]["prefix"] == prefix and host["route"]["cluster"] == backend.url:
+            if host["match"].get("prefix") == prefix and host["route"]["cluster"] == backend.url:
                 return True
         return False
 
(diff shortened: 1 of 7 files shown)


testsuite/tests/singlecluster/authorino/identity/plain/test_jwt_user_story.py
```diff
@@ -0,0 +1,80 @@
...
+    Setup AuthConfig to retrieve identity from given path.
+    Adds authorization rule to start geofence when user doesn't have realm role assign.
+    In case of geofence user without role can only access from /anything/SK path parameter.
+    """
+    authorization.identity.add_plain(
+        "plain", "context.metadata_context.filter_metadata.envoy\\.filters\\.http\\.jwt_authn|verified_jwt"
+    )
+    authorization.authorization.add_auth_rules(
+        "geofence",
+        [Pattern("context.request.http.path", "eq", "/anything/SK")],
```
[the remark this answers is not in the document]
averevki: I see that they can be used interchangeably with other template languages, so no problem using them together with mustache templates inside the testsuite :+1: 

There is a simple if condition in mustache I've found, maybe you can figure out how to adapt this https://stackoverflow.com/questions/8548943/how-to-handle-an-if-statement-in-a-mustache-template

testsuite/gateway/envoy/config.py
```diff
@@ -140,3 +140,24 @@ def remove_all_backends(self):
             "virtual_hosts"
         ][0]["routes"] = {}
         self["envoy.yaml"] = yaml.dump(config)
+
+    @modify
+    def add_routes_for_jwt(self, backend: Backend, path_separated_prefix: str):
```
averevki, line 145: We want to have reusable methods, this looks like an extracted copy of what you had. It needs to be possible to use this same method on a test with wristband envoy and maybe others.

Have it add config components based on the parameters passed to the method. We do that a lot in other constructing classes

testsuite/gateway/envoy/jwt_plain_identity.py
```diff
@@ -7,15 +7,23 @@
...
+        keycloak_realm,
+        keycloak_url,
+        backend_url,
+        labels: dict[str, str],
     ):
         super().__init__(cluster, gw_name, authorino, envoy_image, labels)
         self.server_url = keycloak_url
         self.realm = keycloak_realm
-        self.tools = tools
+        self.backend_url = backend_url
```
averevki, line 26: I can't find where this is used

PR #695, opened 2025-06-26
Add tests for observability configuration and functional behavior

emmaaroche:
### Description:

This PR introduces tests that validate Kuadrant's observability functionality.

**Observability Tests Validate:** 

- ServiceMonitor and PodMonitor resources exist and are labeled with 'kuadrant.io/observability=true'
- Prometheus successfully scrapes metrics from each monitor’s configured endpoints
- All expected ServiceMonitor metrics (e.g., controller_runtime, workqueue) are present per service
- All expected PodMonitor metrics (e.g., Istio metrics) are present
- Predictable PodMonitor metrics (like istio_requests_total) have exactly 2 source-side samples with value 1 each

_**Note:**_ These tests currently cover observability integration with Istio-based gateways. Coverage for Envoy Gateway is not included at this time.

#### Related Issues / PRs:

* kuadrant/kuadrant-operator#1254
* kuadrant/kuadrant-operator#1164

### DIFF
--- a/testsuite/kuadrant/__init__.py
+++ b/testsuite/kuadrant/__init__.py
@@ -5,7 +5,7 @@ import dataclasses
 from openshift_client import selector
 
 from testsuite.kuadrant.authorino import Authorino
-from testsuite.kubernetes import CustomResource
+from testsuite.kubernetes import CustomResource, modify
 from testsuite.kubernetes.deployment import Deployment
 from testsuite.utils import asdict
 
@@ -101,3 +101,8 @@ class KuadrantCR(CustomResource):
         """Returns spec.limitador from Kuadrant object"""
         self.model.spec.setdefault("limitador", {})
         return LimitadorSection(self, "limitador")
+
+    @modify
+    def set_observability(self, enabled: bool):
+        """Enable observability"""
+        self.model.spec["observability"] = {"enable": enabled} if enabled else None
--- a/testsuite/prometheus.py
+++ b/testsuite/prometheus.py
@@ -94,3 +94,15 @@ class Prometheus:
             return False
 
         assert _wait_for_scrape(), "Scrape wasn't done in time"
+
+    @backoff.on_predicate(backoff.constant, interval=5, max_tries=12, jitter=None)
+    def verify_no_observability_targets(self, label_filters: dict[str, list[str]]) -> bool:
+        """Verify that no observability targets are active in Prometheus"""
+        targets = self.get_active_targets()
+
+        for t in targets:
+            labels = t.get("labels", {})
+            for label_key, expected_values in label_filters.items():
+                if labels.get(label_key) in expected_values:
+                    return False
+        return True
--- a/testsuite/tests/singlecluster/observability/conftest.py
+++ b/testsuite/tests/singlecluster/observability/conftest.py
@@ -1,0 +1,108 @@
+"""Conftest for observability tests"""
+
+import backoff
+import pytest
+from openshift_client import selector
+
+from testsuite.kubernetes.monitoring.pod_monitor import PodMonitor
+from testsuite.kubernetes.monitoring.service_monitor import ServiceMonitor
+
+
+SERVICE_MONITOR_SERVICES = [
+    "kuadrant-operator-metrics",
+    "dns-operator-controller-manager-metrics-service",
+    "limitador-operator-metrics",
+    "authorino-operator-metrics",
+]
+
+POD_MONITOR_JOB = [
+    "kuadrant/istio-pod-monitor",
+]
+
+
+@pytest.fixture(scope="module")
+def commit():
+    """Override the commit fixture to do nothing"""
+    return None
+
+
+@pytest.fixture(scope="module", autouse=True)
+def enable_observability(kuadrant, request, prometheus):
+    """Enable / reset observability and verify that no relevant observability targets are active"""
+
+    def _reset():
+        kuadrant.set_observability(False)
+
+    request.addfinalizer(_reset)
+
+    assert prometheus.verify_no_observability_targets(
+        label_filters={
+            "service": SERVICE_MONITOR_SERVICES,
+            "job": POD_MONITOR_JOB,
+        }
+    ), "Observability targets still present in Prometheus before enabling observability"
+
+    kuadrant.set_observability(True)
+    kuadrant.wait_for_ready()
+
+
+@pytest.fixture(scope="module")
+def service_monitors(cluster, testconfig):
+    """Return all 4 expected ServiceMonitors created by enabling observability"""
+    context = cluster.change_project(testconfig["service_protection"]["system_project"]).context
+
+    @backoff.on_predicate(backoff.constant, lambda x: len(x) == 4, interval=5, max_tries=12, jitter=None)
+    def wait_for_monitors():
+        return selector("ServiceMonitor", labels={"kuadrant.io/observability": "true"}, static_context=context).objects(
+            cls=ServiceMonitor
+        )
+
+    all_monitors = wait_for_monitors()
+    assert len(all_monitors) == 4, "Expected ServiceMonitors were not found"
+
+    return all_monitors
+
+
+@pytest.fixture(scope="module")
+def pod_monitor(cluster):
+    """Return PodMonitor created by enabling observability"""
+
+    @backoff.on_predicate(backoff.constant, lambda x: len(x) == 1, interval=5, max_tries=12, jitter=None)
+    def wait_for_monitor():
+        return selector(
+            "PodMonitor", labels={"kuadrant.io/observability": "true"}, static_context=cluster.context
+        ).objects(cls=PodMonitor)
+
+    monitor = wait_for_monitor()
+    assert len(monitor) == 1, "PodMonitor 'istio-pod-monitor' not found"
+
+    return monitor[0]
+
+
+@pytest.fixture(scope="module")
+def service_monitor_metrics_by_service(service_monitors, prometheus):
+    """Return a dictionary, for each expected service, showing which metrics were collected"""
+    metrics_by_service = {}
+
+    for sm in service_monitors:
+        assert prometheus.is_reconciled(sm), f"{sm.name()} not reconciled in Prometheus"
+        prometheus.wait_for_scrape(sm, "/metrics")
+
+    for service_name in SERVICE_MONITOR_SERVICES:
+        metrics = prometheus.get_metrics(labels={"service": service_name})
+        metrics_by_service[service_name] = set(metrics.names)
+
+    return metrics_by_service
+
+
+@pytest.fixture(scope="module")
+def pod_monitor_metrics(client, pod_monitor, prometheus):
+    """Return metrics from PodMonitor"""
+    result = client.get("/get")  # Trigger request traffic to generate istio metrics
+    assert result.status_code == 200
+
+    assert prometheus.is_reconciled(pod_monitor), f"{pod_monitor.name()} not reconciled in Prometheus"
+    prometheus.wait_for_scrape(pod_monitor, "/stats/prometheus")
+
+    metrics = prometheus.get_metrics(labels={"job": f"{pod_monitor.namespace()}/{pod_monitor.name()}"})
+    return metrics
--- a/testsuite/tests/singlecluster/observability/test_observability.py
+++ b/testsuite/tests/singlecluster/observability/test_observability.py
@@ -1,0 +1,65 @@
+"""Tests the enabling of observability configuration via the Kuadrant CR"""
+
+import pytest
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.disruptive]
+
+SERVICE_MONITOR_METRICS = [
+    "controller_runtime_active_workers",
+    "controller_runtime_reconcile_total",
+    "controller_runtime_reconcile_errors_total",
+    "workqueue_work_duration_seconds_bucket",
+    "workqueue_work_duration_seconds_count",
+    "workqueue_queue_duration_seconds_bucket",
+    "workqueue_queue_duration_seconds_count",
+    "workqueue_retries_total",
+    "workqueue_unfinished_work_seconds",
+    "workqueue_adds_total",
+    "workqueue_depth",
+    "workqueue_longest_running_processor_seconds",
+]
+
+POD_MONITOR_METRICS = [
+    "istio_requests_total",
+    "istio_response_bytes_count",
+    "istio_response_bytes_sum",
+    "istio_agent_cert_expiry_seconds",
+    "istio_agent_outgoing_latency",
+    "istio_agent_num_outgoing_requests",
+    "istio_agent_process_resident_memory_bytes",
+    "istio_agent_process_cpu_seconds_total",
+    "istio_agent_scrapes_total",
+    "istio_agent_startup_duration_seconds",
+]
+
+PREDICTABLE_METRICS = [
+    "istio_requests_total",
+    "istio_response_bytes_count",
+]
+
+
+@pytest.mark.parametrize("metric", SERVICE_MONITOR_METRICS)
+def test_service_monitor_metrics_per_service(metric, service_monitor_metrics_by_service):
+    """Tests that all expected metrics appear in the data collected from each ServiceMonitor"""
+    for service_name, metrics in service_monitor_metrics_by_service.items():
+        assert metric in metrics, f"Expected metric '{metric}' in '{service_name}'. Metrics found: {metrics}"
+
+
+@pytest.mark.parametrize("metric", POD_MONITOR_METRICS)
+def test_pod_monitor_metrics(metric, pod_monitor_metrics):
+    """Tests that each expected PodMonitor metric is present"""
+    assert metric in pod_monitor_metrics.names
+
+
+@pytest.mark.parametrize("metric", PREDICTABLE_METRICS)
+def test_pod_monitor_metrics_values(metric, pod_monitor_metrics):
+    """Tests that PodMonitor metrics have expected source entries"""
+    filtered_metrics = pod_monitor_metrics.filter(lambda x: x["metric"]["__name__"] == metric)
+
+    source_metrics = filtered_metrics.filter(lambda x: x["metric"].get("reporter") == "source")
+
+    assert len(source_metrics.values) == 2, f"{metric} expected 2 source metrics, got {len(source_metrics.values)}"
+
+    # Each metric entry should have value of 1
+    for i, value in enumerate(source_metrics.values):
+        assert value == 1, f"{metric} source metric {i} expected value 1, got {value}"


emmaaroche:
_Note:_ refactoring these tests further based on feedback in call with @averevki.

testsuite/prometheus.py
```diff
@@ -82,7 +82,7 @@ def wait_for_scrape(self, monitor: ServiceMonitor | PodMonitor, metrics_path: st
         """Wait before next metrics scrape on service is finished"""
         call_time = datetime.now(timezone.utc)
 
-        @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=4)
+        @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=24)
```
emmaaroche, line 85: Had to up this to 24 as it kept failing with "Scrape wasn't done in time" with the original or similar values.

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+@pytest.fixture(scope="module")
+def generate_requests(client):
+    """Send traffic to generate metrics and verify requests succeed"""
+
+    def _send(n=5, delay=0.5):
+        for _ in range(n):
+            response = client.get("/get")
+            assert response.status_code == 200
+            if delay:
+                time.sleep(delay)
```
averevki, line 91: Are there any generated istio/kuadrant metrics you can predict? For example you send 5 requests here, and some istio metric like `istio_requests_total` equals 5? It would be good if we can match those metrics with the actual requests we sending inside the test, and not to just check that they appear with some random number

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.fixture(scope="function")
+def observability_enabled(kuadrant, request):
+    """Enable observability before the test and reset it after"""
+
+    kuadrant.refresh().model.spec.observability = {"enable": True}
+    kuadrant.apply()
+    kuadrant.wait_for_ready()
+    kuadrant.refresh()
```
averevki, line 29: I think you don't need to refresh objects after `wait_for_ready`. It's doing it for you

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+def rate_limit():
+    """Override the rate limit fixture to prevent the creation of an RateLimitPolicy"""
+    return None
+
+
+@pytest.fixture(scope="function")
+def observability_enabled(kuadrant, request):
+    """Enable observability before the test and reset it after"""
+
+    kuadrant.refresh().model.spec.observability = {"enable": True}
```
averevki, line 26: Looks like a new method for `KuadrantCR` class

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+@pytest.fixture(scope="module")
+def authorization():
+    """Override the authorization fixture to prevent the creation of an AuthPolicy"""
+    return None
+
+
+@pytest.fixture(scope="module")
+def rate_limit():
+    """Override the rate limit fixture to prevent the creation of an RateLimitPolicy"""
+    return None
```
averevki, line 19: If you look at `pytest --setup-plan`, you'll see that `commit` fixture is executed either way because it is on autouse. And it's doing nothing because it supposed to commit authpolicy and ratelimitpolicy, but you set them as None here. 

Just override commit fixture to do nothing here, instead of overwriting both policies. That way 1 fixture will be executed instead of 3. Also you can put some other fixtures, you don't know where to request at, as dependencies to the new `commit`

testsuite/prometheus.py
averevki: You probably mixing waits here, as far as I remember there were 2 main waits in prometheus - for Monitor object reconciliation and assigning it as one of the active targets for Prometheus scrape, and other one that actually waits for the new scrape every few seconds. Check `Prometheus.is_reconciled` method

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,104 @@
...
+    assert wait_for_monitors(present=True)
+
+    service_monitors, pod_monitors = get_all_monitors()
+    all_monitors = service_monitors + pod_monitors
+    assert all_monitors, "No ServiceMonitor or PodMonitor resources found"
+
+    assert {
+        testconfig["service_protection"]["system_project"],
+        testconfig["service_protection"]["project"],
+    } <= {m.namespace() for m in all_monitors}, "Missing monitors in expected namespaces"
```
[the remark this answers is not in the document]
emmaaroche: I thought so too, I tried using cursor for the first time and it gave me this as a solution to a make commit-acceptance local variable issue I was getting with the previous method I was using :sweat_smile: 

It basically asserts that the monitors exist in the expected namespaces.

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,104 @@
...
+    servicemonitors, podmonitors = get_all_monitors()
+
+    assert not servicemonitors, "ServiceMonitor resources exist"
+    assert not podmonitors, "PodMonitor resources exist"
+
+    for monitor in servicemonitors + podmonitors:
+        labels = monitor.model.metadata.labels or {}
+        assert (
+            "kuadrant.io/observability" not in labels
+        ), f"Unexpected observability label on {monitor.kind()} {monitor.name()} in {monitor.namespace()}"
```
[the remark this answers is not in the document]
emmaaroche: I felt it was worth checking both, not just whether the monitors exist or not, but also that they don’t accidentally carry the observability label. Maybe this is not necessary though and happy to remove this.

testsuite/tests/singlecluster/observability/test_observability.py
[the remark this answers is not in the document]
emmaaroche: Maybe an unnecessary check, should I remove?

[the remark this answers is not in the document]
averevki: > I thought so too, I tried using cursor for the first time and it gave me this as a solution to a make commit-acceptance local variable issue I was getting with the previous method I was using 😅
> 
> It basically asserts that the monitors exist in the expected namespaces.

Comparison between sets is not so intuitive for me :smile:

testsuite/tests/singlecluster/observability/test_observability.py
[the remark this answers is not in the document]
emmaaroche: yeah, that is true, I think I liked it more for extra security, which is not necessary 😅

testsuite/tests/singlecluster/observability/conftest.py
emmaaroche: Yes, `istio_requests_total` is predictable, it goes up by the number of requests we send (so if we send 5, it increases by 5). `istio_response_bytes_count` also seems to match the number of responses. These two are good to check against the traffic we generate. The others I currently have listed aren’t as consistent as far as I can see.

averevki: That's nice, see if you can match that inside the test for istio metrics

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,77 @@
...
+
+@pytest.fixture(scope="module")
+def service_monitor_metrics(observability_setup, prometheus):
+    """Return metrics from ServiceMonitors"""
+    service_monitors, _ = observability_setup
+    for sm in service_monitors:
+        assert prometheus.is_reconciled(sm), f"{sm.name()} not reconciled in Prometheus"
+        prometheus.wait_for_scrape(sm, "/metrics")
+
+    return prometheus.get_metrics(labels={"job": "metrics"}).names
```
averevki, line 36: More specific label if possible. There were 18783 metrics fetched for me

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+
+    request.addfinalizer(_reset)
+    return service_monitors, pod_monitors
+
+
+@pytest.fixture(scope="module")
+def wait_for_monitors(cluster, testconfig):
+    """Wait for ServiceMonitor and PodMonitor resources"""
+
+    def _wait(present=True):
```
averevki, line 50: Probably a leftover, but it seems like you using this function only once, you don't need nested fixture with parameter for this.

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+        @backoff.on_predicate(backoff.constant, interval=5, max_tries=12, jitter=None)
```
averevki, line 51: Can be handled with the testsuite kubernetes client and `.wait_until()` or similar

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@pytest.fixture(scope="module")
+def pod_monitor_metrics(client, observability_setup, prometheus):
+    """Return metrics from PodMonitors"""
+    for _ in range(5):
+        response = client.get("/get")
+        assert response.status_code == 200
+        time.sleep(0.5)
```
averevki, line 45: what this sleep for?

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,77 @@
...
+        assert prometheus.is_reconciled(sm), f"{sm.name()} not reconciled in Prometheus"
+        prometheus.wait_for_scrape(sm, "/metrics")
+
+    return prometheus.get_metrics(labels={"job": "metrics"}).names
+
+
+@pytest.fixture(scope="module")
+def pod_monitor_metrics(client, observability_setup, prometheus):
+    """Return metrics from PodMonitors"""
+    for _ in range(5):
```
averevki, line 42: `client.get_many()`

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,77 @@
...
+        response = client.get("/get")
+        assert response.status_code == 200
+        time.sleep(0.5)
+
+    _, pod_monitors = observability_setup
+    for pm in pod_monitors:
+        assert prometheus.is_reconciled(pm), f"{pm.name()} not reconciled in Prometheus"
+        prometheus.wait_for_scrape(pm, "/stats/prometheus")
```
averevki, line 50: What's on `/stats/prometheus`? There were only  `/metrics` endpoints scraped in podmonitors I saw

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,77 @@
...
+    "istio_response_bytes_sum",
+    "istio_agent_cert_expiry_seconds",
+    "istio_agent_outgoing_latency",
+]
+
+
+@pytest.fixture(scope="module")
+def service_monitor_metrics(observability_setup, prometheus):
+    """Return metrics from ServiceMonitors"""
+    service_monitors, _ = observability_setup
```
averevki, line 31: 2 fixtures, and you wont need to make stunts like this :bicyclist:

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,77 @@
...
+Tests the enabling and disabling of observability configuration via the Kuadrant CR
+"""
+
+import time
+
+import pytest
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.disruptive]
+
+SERVICE_MONITOR_METRICS = [
```
averevki, line 11: are these only metrics 4 kuadrant operators expose?

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+            break
+        time.sleep(5)
+    else:
+        raise AssertionError("Observability targets still present in Prometheus before enabling observability")
+
+
+@pytest.fixture(scope="module")
+def observability_setup(gateway, kuadrant, request, wait_for_monitors):  # pylint: disable=unused-argument
+    """Enable observability before the test and wait to retrieve expected monitors"""
+    kuadrant.set_observability(True)
```
averevki, line 34: better as extension of kuadrant fixture. Also, why create broad fixtures like `observability_setup` when you can make a lot of dedicated small ones, each doing 1-2 things, and assign dependencies between them. Everything else here can be assigned to other fixtures it seems

testsuite/prometheus.py
averevki: How much time it takes to wait for istio and kuadrant monitors to appear in active targets? And how much time it takes for scrape to be made on these targets right after they were added as active?

Because, I think, the only thing Kuadrant observability will do is create those monitors, after that prometheus is the one responsible about adding metrics to his metrics pool. We are already doing all these waits with another metrics test, which is stable, so I'm not sure why it wants more time here :confused:

testsuite/kuadrant/__init__.py
```diff
@@ -101,3 +101,11 @@ def limitador(self) -> LimitadorSection:
...
         self.model.spec.setdefault("limitador", {})
         return LimitadorSection(self, "limitador")
+
+    def set_observability(self, enabled: bool):
+        """Enable or disable observability and wait for readiness"""
+        self.refresh().model.spec["observability"] = {"enable": enabled} if enabled else None
+        self.apply()
+        self.wait_for_ready()
+        if enabled:
+            self.wait_until(lambda obj: obj.model.spec.get("observability", {}).get("enable") is True)
```
averevki, line 111: This is handled by kubernetes. Check if apply operation was successful instead

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.fixture(scope="module")
+def commit():
+    """Override the commit fixture to do nothing"""
+    return None
+
+
+@pytest.fixture(scope="module", autouse=True)
+def no_observability_metrics_before_enable(prometheus):
```
averevki, line 20: how much time it take for targets to get removed after monitors are removed? Are there any metric properties that can indicate when they will be removed?

testsuite/tests/singlecluster/observability/test_observability.py
emmaaroche: No, I just selected 5 to not have a large number of metrics, but I can add more if you want?

testsuite/tests/singlecluster/observability/test_observability.py
emmaaroche: The path is `/stats/prometheus` because that’s the endpoint set in the PodMonitor YAML:

```
spec:
  namespaceSelector: {}
  podMetricsEndpoints:
  - bearerTokenSecret:
      key: ""
      name: ""
    interval: 30s
    path: /stats/prometheus
```

This is how it’s configured to scrape metrics from the istio-proxy sidecar.

testsuite/prometheus.py
emmaaroche: The monitors appear almost instantly once observability is enabled, so I am not too sure either, but any time I lesson this number it fails with the "Scrape wasn't done in time" error, so not sure what else to do with this 🫠

testsuite/tests/singlecluster/observability/conftest.py
emmaaroche: I tried using `.wait_until()` with the testsuite Kubernetes client, but it gave lint errors like `"Instance of 'APIObject' has no 'wait_until' member"`. I didn’t look for other similar methods because I decided to remove the wait logic instead. The monitors show up right away after enabling observability, so I don’t think the wait is needed really. I have instead added other asserts that will not let test progress if monitors are not present. What do you think?

testsuite/tests/singlecluster/observability/test_observability.py
emmaaroche: job = kuadrant/istio-pod-monitor

testsuite/tests/singlecluster/observability/test_observability.py
emmaaroche: I just checked in my environment and I’m not seeing a fifth service monitor in the istio-system namespace with the kuadrant.io/observability=true label , only the expected 4 in kuadrant-system

emmaaroche: Update to above comment - we discussed as to why this was being created for you, so all good now!

testsuite/prometheus.py
emmaaroche: I have added a wait for monitors backoff back to `service_monitors` and `pod_monitor`, which has seemed to fix this issue occurring, I have reverted this backoff to it's original value of 4.

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.fixture(scope="module")
+def gateway(request, cluster, blame, wildcard_domain, module_label):
+    """Create and configure Gateway with module scope"""
+    gw = KuadrantGateway.create_instance(cluster, blame("gw"), {"app": module_label})
+    gw.add_listener(GatewayListener(hostname=wildcard_domain))
+    request.addfinalizer(gw.delete)
+    gw.commit()
+    gw.wait_for_ready()
+    return gw
```
emmaaroche, line 27: Context for adding this `gateway`: since adding the `kuadrant` extension tests failed instantly with: 

```
ScopeMismatch: You tried to access the module scoped fixture kuadrant with a session scoped request object. Requesting fixture stack:
testsuite/tests/singlecluster/conftest.py:130:  def gateway(request, kuadrant, cluster, blame, label, testconfig, wildcard_domain) -> testsuite.gateway.Gateway
```

Adding this gateway is how I fixed this issue, but maybe there is a better way.

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,122 @@
...
+        "istio_requests_total": 10,
+        "istio_response_bytes_count": 10,
+    }
+
+    if metric in predictable_metrics:
+        result = prometheus.get_metrics(metric)
+        total = sum(result.values) if result and result.values else 0
+        assert (
+            total == predictable_metrics[metric]
+        ), f"{metric} expected to be {predictable_metrics[metric]}, got {total}"
```
averevki, line 122: this can be a separate test

testsuite/tests/singlecluster/observability/conftest.py
averevki: yeah i guess because only the client fixture in test wants gateway that wants kuadrant. In case that this provokes us to add the gateway fixture here, I'd come back to a separate fixture to setup the kuadrant observability (`setup_observability` or what was it before). We will have only 1 fixture instead of 2, which I think even ok to be on autouse for this test, wyt?

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+    """Return PodMonitor created by enabling observability"""
+
+    @backoff.on_predicate(backoff.constant, interval=5, max_tries=12, jitter=None)
+    def wait_for_monitor():
+        result = selector("podmonitor/istio-pod-monitor", static_context=cluster.context).objects(cls=PodMonitor)
+        return result[0] if len(result) == 1 else None
+
+    monitor = wait_for_monitor()
+    if not monitor:
+        raise AssertionError("PodMonitor 'istio-pod-monitor' not found")
```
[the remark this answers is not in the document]
emmaaroche: updated these to assert 👍

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,122 @@
...
+
+    for sm in service_monitors:
+        assert prometheus.is_reconciled(sm), f"{sm.name()} not reconciled in Prometheus"
+        prometheus.wait_for_scrape(sm, "/metrics")
+
+        for metric in SERVICE_MONITOR_METRICS:
+            data = prometheus.get_metrics(metric, labels={"job": f"{sm.name()}"})
+            if data:
+                # Collect the same metric from each ServiceMonitor that has it
+                results.setdefault(metric, []).extend(data.names)
```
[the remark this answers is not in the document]
emmaaroche: The idea was to confirm each expected metric appeared at least once across all monitors

testsuite/kuadrant/__init__.py
```diff
@@ -101,3 +101,11 @@ def limitador(self) -> LimitadorSection:
         """Returns spec.limitador from Kuadrant object"""
         self.model.spec.setdefault("limitador", {})
         return LimitadorSection(self, "limitador")
+
+    @modify
+    def set_observability(self, enabled: bool):
+        """Enable or disable observability and wait for readiness"""
+        self.refresh().model.spec["observability"] = {"enable": enabled} if enabled else None
+        res = self.apply()
+        assert res.status() == 0, res.err()
```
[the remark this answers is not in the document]
emmaaroche: `@modify` must not be working for `set_observability` beacuse when I remove

```
res = self.apply()
assert res.status() == 0, res.err()
```
observability does not enable as expected.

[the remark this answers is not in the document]
emmaaroche: should I just remove `@modify ` in this case?

testsuite/tests/singlecluster/observability/conftest.py
emmaaroche: Yeah makes sense. I removed the `gateway` fixture and added the `enable_observability` one back with `autouse=True` in updates.

testsuite/tests/singlecluster/observability/test_observability.py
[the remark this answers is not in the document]
emmaaroche: In the new changes, this fixture collects metrics separately for each service monitor, using the related service label, and returns them as a dictionary. 

I also updated the service monitor test to: `test_service_monitor_metrics_per_service`, this test verifies that all expected metrics from `SERVICE_MONITOR_METRICS` appear in the data collected from each individual service monitor, ensuring coverage per service monitor rather than mixing everything together.

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+def pod_monitor_metrics(client, pod_monitor, prometheus):
+    """Return metrics from PodMonitor"""
+    result = client.get("/get")  # Trigger request traffic to generate istio metrics
+    assert result.status_code == 200
+
+    assert prometheus.is_reconciled(pod_monitor), f"{pod_monitor.name()} not reconciled in Prometheus"
+    prometheus.wait_for_scrape(pod_monitor, "/stats/prometheus")
+
+    metrics = prometheus.get_metrics(labels={"job": POD_MONITOR_JOB[0]})
+    return set(metrics.names) if metrics and metrics.names else set()
```
averevki, line 120: overthinking here :cat:
```suggestion
    return set(metrics.names)
```

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.mark.parametrize("metric", POD_MONITOR_METRICS)
+def test_pod_monitor_metrics(metric, pod_monitor_metrics):
+    """Tests that each expected PodMonitor metric is present"""
+    assert metric in pod_monitor_metrics
+
+
+@pytest.mark.parametrize("metric", PREDICTABLE_METRICS)
+def test_pod_monitor_metrics_values(metric, prometheus):
+    """Tests that predictable PodMonitor metrics have the expected values"""
+    # Both source and destination proxies emit metrics for each request, so 1 request results in 2 metric entries total
```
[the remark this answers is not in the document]
emmaaroche: https://docs.sysdig.com/en/docs/sysdig-monitor/integrations/integration-library/istio/

https://isar-nasimov.medium.com/unlock-the-secrets-of-istio-master-monitoring-with-these-expert-tips-175197aab4e9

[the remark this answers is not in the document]
emmaaroche: Yeah, I noticed both source and destination proxies emit metrics, so each request ends up counted twice, that’s why I set the expected value to 2 here.

I kept it that way to make sure both sides are emitting metrics as expected, but let me know if it would make more sense to just filter by one reporter (like destination) instead?

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,68 @@
...
+    predictable_metrics = {
+        "istio_requests_total": 2,
+        "istio_response_bytes_count": 2,
+    }
+
+    if metric in predictable_metrics:
```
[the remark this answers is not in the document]
emmaaroche: Good point 😅  I’ve removed the if check in new updates and switched to using .filter() on pod_monitor_metrics to grab just the matching metric by name and sum its values.

testsuite/prometheus.py
```diff
@@ -94,3 +94,21 @@ def _wait_for_scrape():
...
+            for t in targets:
+                labels = t.get("labels", {})
+                if labels.get("service") in service_monitor_services:
+                    return False
+                if labels.get("job") in pod_monitor_jobs:
+                    return False
+
+            return True
+
+        return targets_cleared()
```
averevki, line 114: You need to generalize class methods more, their intention is to be reused. Seems like you just copied the fixture here. See if you can make this method here working only for a single servicemonitor/podmonitor (also ideally generalized (not just 2 different parameters but decision based on a value this method got)) :sweat_smile:

testsuite/tests/singlecluster/observability/test_observability.py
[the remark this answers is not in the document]
averevki: I put better flow for this test inside of the comments in my last review. That's nice that you check for both, but I see there only `source` istio metrics and I think those are different metrics (health checks, prometheus queries).

testsuite/tests/singlecluster/observability/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+@pytest.fixture(scope="module")
+def pod_monitor_metrics(client, pod_monitor, prometheus):
+    """Return metrics from PodMonitor"""
+    result = client.get("/get")  # Trigger request traffic to generate istio metrics
+    assert result.status_code == 200
+
+    assert prometheus.is_reconciled(pod_monitor), f"{pod_monitor.name()} not reconciled in Prometheus"
+    prometheus.wait_for_scrape(pod_monitor, "/stats/prometheus")
+
+    metrics = prometheus.get_metrics(labels={"job": POD_MONITOR_JOB[0]})
```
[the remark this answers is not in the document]
averevki: It'll prevent possible bugs from appearing in the future. But if it makes sense for your code to have them both as list (e.g. you reuse them in a method that can only accept lists), go ahead

testsuite/tests/singlecluster/observability/test_observability.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@pytest.mark.parametrize("metric", PREDICTABLE_METRICS_EXPECTED)
+def test_pod_monitor_metrics_values(metric, pod_monitor_metrics):
+    """Tests that predictable PodMonitor metrics have the expected values"""
+    filtered_metrics = pod_monitor_metrics.filter(lambda x: x["metric"]["__name__"] == metric)
+    total = sum(filtered_metrics.values)
+    assert (
+        total == PREDICTABLE_METRICS_EXPECTED[metric]
+    ), f"{metric} expected to be {PREDICTABLE_METRICS_EXPECTED[metric]}, got {total}"
```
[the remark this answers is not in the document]
emmaaroche: Yeah, it is just the source istio metrics appearing, no destination, so I updated the test to check for 2 source entries in total with a value of 1 each.

testsuite/prometheus.py
```diff
@@ -94,3 +94,19 @@ def _wait_for_scrape():
...
+            targets = self.get_active_targets()
+
+            for t in targets:
+                labels = t.get("labels", {})
+                for label_key, expected_values in label_filters.items():
+                    if labels.get(label_key) in expected_values:
+                        return False
+            return True
+
+        return targets_cleared()
```
averevki, line 112: Can backoff be moved to the method level here? There is assert inside the method above, where backoff is nested, but here you are executing the nested function right after defining it, and that's it

emmaaroche: Updated this 👍

averevki (approved):

azgabur (approved):

PR #699, opened 2025-07-02
Fix invalid issuer test

averevki:
## Description
https://github.com/Kuadrant/kuadrant-operator/pull/1393 changed message returned if an invalid issuer kind has been set (like 'Gateway' in this test). Validation also was moved to the object creation

## Verification steps
`pytest /home/averevki/work/testsuite/testsuite/tests/singlecluster/gateway/reconciliation/test_invalid_issuer_reference.py`

### DIFF
--- a/testsuite/tests/singlecluster/gateway/reconciliation/test_invalid_issuer_reference.py
+++ b/testsuite/tests/singlecluster/gateway/reconciliation/test_invalid_issuer_reference.py
@@ -15,7 +15,7 @@ def commit():
     return None
 
 
-def test_wrong_issuer_type(request, gateway, blame, module_label):
+def test_wrong_issuer_type(request, gateway, blame, module_label, cluster):
     """Tests that TLSPolicy is rejected if issuer does not have a correct type"""
 
     policy = TLSPolicy.create_instance(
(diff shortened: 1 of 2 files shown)


testsuite/tests/singlecluster/gateway/reconciliation/test_invalid_issuer_reference.py
```diff
@@ -30,18 +30,12 @@ def test_wrong_issuer_type(request, gateway, blame, module_label):
...
-            'issuerRef.kind. Must be empty, "Issuer" or "ClusterIssuer"',
-        ),
-        timelimit=20,
-    ), f"Policy did not reach expected status, instead it was: {policy.refresh().model.status.conditions}"
+    res = cluster.do_action("create", "-f", "-", stdin_str=policy.as_json(), auto_raise=False)
+    assert res.status() == 1, "Policy should not be created with invalid issuerRef.kind"
+    assert res.err().strip() == (
+        f'The TLSPolicy "{policy.model.metadata.name}" is invalid: spec.issuerRef: Invalid value: "object": '
+        f"Invalid issuerRef.kind. The only supported values are blank, 'Issuer' and 'ClusterIssuer'"
+    )
```
emmaaroche, line 38: Would it make sense to change this to something like `assert "Invalid issuerRef.kind" in res.err()`? Just thinking it might make the test a bit more flexible if the full error message ever changes slightly in the future. I am not totally sure if this would be better, just a thought, so curious what you think 🤔

emmaaroche (approved):
LGTM, test is passing. I left one comment with a small suggestion, but it is not a blocker. I am happy to approve it as it is.

testsuite/tests/singlecluster/gateway/reconciliation/test_invalid_issuer_reference.py
averevki: @emmaaroche I think error message is a part of the test here. Even though you're right I aligned this test with how validation is handled on TLSPolicy a few times already, I think it's better to have it more strict here in case something breaks and only part of the error message is shown. 

Also I think only second line is added by Kuadrant and first one is automatically generated by k8s. So we can ***expect*** first line to always be correct ig

averevki: It'll show object creation error. I need the message where validation is failed, extracted by oc client

PR #708, opened 2025-07-14
Token Rate Limit Policy tests

crstrn13:
## Description
Add Token Rate Limit Policies tests.

### DIFF
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -11,6 +11,8 @@ default:
       password: "testPassword"
   httpbin:
     image: "quay.io/trepel/httpbin:jsmadis"
+  llm_sim:
+    image: "ghcr.io/llm-d/llm-d-inference-sim:v0.1.1"
   mockserver:
     image: "quay.io/mganisin/mockserver:latest"
   service_protection:
--- a/testsuite/backend/llm_sim.py
+++ b/testsuite/backend/llm_sim.py
@@ -1,0 +1,36 @@
+from testsuite.backend import Backend
+from testsuite.kubernetes import Selector
+from testsuite.kubernetes.client import KubernetesClient
+from testsuite.kubernetes.deployment import Deployment
+from testsuite.kubernetes.service import Service, ServicePort
+
+class LlmSim(Backend):
+    """LlmSim deployed in Kubernetes as Backend"""
+
+    def __init__(self, cluster: KubernetesClient, name, label, image, replicas=1) -> None:
+        super().__init__(cluster, name, label)
+        self.replicas = replicas
+        self.image = image
+
+    def commit(self):
+        match_labels = {"app": self.label, "deployment": self.name}
+        self.deployment = Deployment.create_instance(
+            self.cluster,
+            self.name,
+            container_name="llm-sim",
+            image=self.image,
+            ports={"api": 8080},
+            selector=Selector(matchLabels=match_labels),
+            labels={"app": self.label},
+            command_args=["--model", "meta-llama/Llama-3.1-8B-Instruct", "--port", "8080"],
+        )
+        self.deployment.commit()
+        self.deployment.wait_for_ready()
+
+        self.service = Service.create_instance(
+            self.cluster,
+            self.name,
+            selector=match_labels,
+            ports=[ServicePort(name="http", port=8080, targetPort="api")],
+        )
+        self.service.commit()
--- a/testsuite/kuadrant/policy/token_rate_limit.py
+++ b/testsuite/kuadrant/policy/token_rate_limit.py
@@ -1,0 +1,30 @@
+from testsuite.gateway import Referencable
+from testsuite.kuadrant.policy.rate_limit import RateLimitPolicy
+from testsuite.kubernetes.client import KubernetesClient
+
+
+class TokenRateLimitPolicy(RateLimitPolicy):
+    """TokenRateLimitPolicy object, used for applying token rate limiting rules to a Gateway/HTTPRoute"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.spec_section = None
+
+    @classmethod
+    def create_instance(
+        cls,
+        cluster: KubernetesClient,
+        name,
+        target: Referencable,
+        section_name: str = None,
+        labels: dict[str, str] = None,
+    ):
+        """Creates new instance of TokenRateLimitPolicy"""
+        model: dict = {
+            "apiVersion": "kuadrant.io/v1alpha1",
+            "kind": "TokenRateLimitPolicy",
+            "metadata": {"name": name, "labels": labels},
+        }
+        if section_name:
+            model["spec"]["targetRef"]["sectionName"] = section_name
+        return cls(model, context=cluster.context)
--- a/testsuite/tests/singlecluster/llm-d/tokens_rate_limit/conftest.py
+++ b/testsuite/tests/singlecluster/llm-d/tokens_rate_limit/conftest.py
@@ -1,0 +1,50 @@
+import pytest
+
+from testsuite.kuadrant.policy.authorization import JsonResponse, ValueFrom
+
+@pytest.fixture(scope="module")
+def user_label(blame):
+    """Creates a label prefixed as user"""
+    return blame("user")
+
+@pytest.fixture(scope="module")
+def free_user_api_key(create_api_key, user_label):
+    """Creates API key Secret for a free user"""
+    annotations = {
+        "kuadrant.io/groups": "free",
+        "secret.kuadrant.io/user-id": "user-1"
+    }
+    secret = create_api_key("api-key", user_label, "iamafreeuser", annotations=annotations)
+    return secret
+
+@pytest.fixture(scope="module")
+def paid_user_api_key(create_api_key, user_label):
+    """Creates API key Secret for a paid user"""
+    annotations = {
+        "kuadrant.io/groups": "paid",
+        "secret.kuadrant.io/user-id": "user-2"
+    }
+    secret = create_api_key("api-key", user_label, "iamapaiduser", annotations=annotations)
+    return secret
+
+@pytest.fixture(scope="module")
+def authorization(authorization, user_label, free_user_api_key):
+    authorization.identity.add_api_key("api-key", selector=free_user_api_key.selector)
+    authorization.responses.add_success_dynamic(
+        "identity", JsonResponse({"anonymous": ValueFrom("{auth.identity.anonymous}")})
+    )
+    authorization.authorization.add_opa_policy(
+        "allow-groups",
+        """
+        groups := split(object.get(input.auth.identity.metadata.annotations, "kuadrant.io/groups", ""), ",")
+        allow { groups[_] = "free" }
+        allow { groups[_] = "paid" }
+        """
+    )
+    return authorization
+
+@pytest.fixture(scope="module", autouse=True)
+def commit(request, authorization):
+    request.addfinalizer(authorization.delete)
+    authorization.commit()
+    authorization.wait_for_ready()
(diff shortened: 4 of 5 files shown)


testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,142 @@
...
+
+@pytest.mark.parametrize("token_rate_limit", ["route", "gateway"], indirect=True)
+def test_missing_api_key(client):
+    """Ensures requests without an API key are rejected with 401 Unauthorized"""
+    response = client.post(
+        "/v1/chat/completions",
+        json=BASIC_REQUEST,
+    )
+    assert response.status_code == 401, f"Expected status code 401, but got {response.status_code}"
+
```
averevki, line 55: Wouldn't these tests only test authpolicy authentication?

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,142 @@
...
+        if response.status_code == 429:
+            break
+
+        assert response.status_code == 200, f"Expected status code 200, but got {response.status_code}"
+
+        tokens_used = response.json().get("usage", {}).get("total_tokens", 0)
+        total_tokens += tokens_used
+
+    else:
+        assert False, f"Rate limit not triggered after {total_tokens} tokens"
```
averevki, line 104: Put this inside methods please. I can suggest creating `post_many` if you need to send requests with same parameters. Also `.extract_response()`

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+    assert response.status_code == 429, f"Follow-up request should be 429, but got {response.status_code}"
+
+
+@pytest.mark.parametrize("token_rate_limit", ["route", "gateway"], indirect=True)
+def test_token_rate_limit_paid_user(client, paid_user_auth):
+    """Ensures paid-tier users are rate limited after exceeding their token quota"""
+    total_tokens = 0
+
+    for _ in range(20):
```
averevki, line 120: Where do the numbers 10 and 20 come from?

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,142 @@
...
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.limitador]
+
+
+BASIC_REQUEST = {
+    "model": "meta-llama/Llama-3.1-8B-Instruct",
+    "messages": [{"role": "user", "content": "What is Kubernetes?"}],
+}
+
+
+@pytest.mark.parametrize("token_rate_limit", ["route", "gateway"], indirect=True)
```
averevki, line 14: You don't need indirect parametrization in these tests. It is used to granularly set parameters for each test, but you have the same line on every test, which we trying to avoid. Less code. 
```py
@pytest.fixture(scope="module", params=["route", "gateway"])
```
on trlp fixture or something similar will be enough

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,142 @@
...
+def test_token_rate_limit_paid_user(client, paid_user_auth):
+    """Ensures paid-tier users are rate limited after exceeding their token quota"""
+    total_tokens = 0
+
+    for _ in range(20):
+        response = client.post(
+            "/v1/chat/completions",
+            auth=paid_user_auth,
+            json={**BASIC_REQUEST, "max_tokens": 30, "stream": False, "usage": True},
```
averevki, line 124: what these parameters do, how am I supposed to know this here? What stream and usage do? We need to better document this or improve test so it is clear from the code

config/settings.yaml
```diff
@@ -11,6 +11,8 @@ default:
       password: "testPassword"
   httpbin:
     image: "quay.io/rhn_support_azgabur/httpbin:latest"
+  llm_sim:
+    image: "ghcr.io/llm-d/llm-d-inference-sim:v0.1.1"
```
averevki, line 15: please also document what this config do in settings.yaml.tpl, thanks

testsuite/backend/llm_sim.py
```diff
@@ -0,0 +1,39 @@
...
+        match_labels = {"app": self.label, "deployment": self.name}
+        self.deployment = Deployment.create_instance(
+            self.cluster,
+            self.name,
+            container_name="llm-sim",
+            image=self.image,
+            ports={"api": 8080},
+            selector=Selector(matchLabels=match_labels),
+            labels={"app": self.label},
+            command_args=["--model", "meta-llama/Llama-3.1-8B-Instruct", "--port", "8080"],
```
averevki, line 28: this model is not specific to the default image, is it?

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,142 @@
+"""Test TokenRateLimitPolicy functionality"""
+
+import pytest
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.limitador]
```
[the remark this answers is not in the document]
crstrn13: Yes, limitador is the one managing the token rate limit policy. We should add the mark for authorino as well.

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
crstrn13: Well, basically yes. But we also want to test that the integration with the authorino does not break, hence we still do these checks.

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
crstrn13: I believe that without them, the `llm-sim` would return 400. Some comments regarding those would be helpful indeed.

testsuite/kuadrant/policy/token_rate_limit.py
```diff
@@ -0,0 +1,35 @@
...
+        model: dict = {
+            "apiVersion": "kuadrant.io/v1alpha1",
+            "kind": "TokenRateLimitPolicy",
+            "metadata": {"name": name, "labels": labels},
+            "spec": {
+                "targetRef": target.reference,
+            },
+        }
+        if section_name:
+            model["spec"]["targetRef"]["sectionName"] = section_name
```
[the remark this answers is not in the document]
averevki: https://github.com/Kuadrant/testsuite/issues/732

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
averevki: I'd test integration with limitador processing llm requests after authorino is done with its flow. It won't even get to the trlp here? It will get rejected on the said authentication stage.

testsuite/tests/singlecluster/limitador/tokens_rate_limit/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+    target_ref = request.getfixturevalue(getattr(request, "param", "route"))
+
+    policy = TokenRateLimitPolicy.create_instance(
+        cluster, blame(f"trlp-{request.param}"), target_ref, labels={"testRun": module_label}
+    )
+
+    # Free user limit - 50 tokens per 24h
+    policy.add_limit(
+        name="free",
+        limits=[Limit(limit=50, window="24h")],
```
averevki, line 93: This is one of the main features of trlp right? Any plans on testing the tokens limit more extensively?

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
averevki: but you're sending requests without these parameters above, and 200 is returned

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
emmaaroche: Yeah, these don’t hit TRLP logic directly since Authorino blocks them first. However, the reason they are here is to make sure TRLP being present doesn’t interfere with auth, and that unauthenticated users don’t consume quota or get the wrong error (e.g. 429 instead of 401).

testsuite/tests/singlecluster/limitador/tokens_rate_limit/conftest.py
emmaaroche: We’ve covered the basic happy paths for now, but we’ve also reached out to someone actively using TRLP to get feedback on what scenarios would also be valuable to test. Once we have that input, we plan to expand the tests more extensively in the near future.

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
averevki: Sure, divide those tests properly in that case. I can think of hundreds "basic" usecases like that.

Basic use cases should only utilize minimal setup needed for trlp to work. These 7 tests all use the same one

testsuite/tests/singlecluster/limitador/tokens_rate_limit/test_trlp.py
```diff
@@ -0,0 +1,78 @@
+"""Test TokenRateLimitPolicy functionality"""
+
+import pytest
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.limitador, pytest.mark.authorino]
+
+
+@pytest.fixture(autouse=True)
+def add_opa_policy(authorization):
```
averevki, line 9: wrong module for opa policy. And if you'll name it the same as authorization fixture, it will be picked up without autouse

testsuite/backend/llm_sim.py
averevki: Is the setting for the model important in this case when it is basically used as a name, can it be some `kuadrant-qe-model` static string? Or does it depend on the llm?

testsuite/httpx/__init__.py
```diff
@@ -92,6 +92,15 @@ def __str__(self):
...
         return f"Result[error={self.error}]"
 
+    def extract_tokens(self) -> int:
+        """Extract total_tokens from an LLM response, or 0 if not present"""
+        if self.response is not None:
+            try:
+                return self.response.json().get("usage", {}).get("total_tokens", 0)
+            except (ValueError, TypeError, AttributeError):
+                return 0
+        return 0
```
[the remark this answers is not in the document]
emmaaroche: The token usage is only in the response body (usage.total_tokens) and not in the headers, so `extract_response` doesn’t work for this case.

testsuite/backend/llm_sim.py
crstrn13: it can be anything you wish. The only thing you need to take into account is that when you make the request and specify the `model` it needs to be the same as the one you give when running the deployment.

testsuite/tests/singlecluster/limitador/tokens_rate_limit/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+    target_ref = request.getfixturevalue(getattr(request, "param", "route"))
+
+    policy = TokenRateLimitPolicy.create_instance(
+        cluster, blame(f"trlp-{request.param}"), target_ref, labels={"testRun": module_label}
+    )
+
+    # Free user limit - 50 tokens per 30s
+    policy.add_limit(
+        name="free",
+        limits=[Limit(limit=50, window="30s")],
```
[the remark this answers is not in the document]
emmaaroche: Have updated paid user to have a 60s limit

PR #709, opened 2025-07-14
Add kube-burner to dockerfile

averevki:
For the possibility to run scale tests inside image


azgabur (approved):
It builds. Maybe we could also update versions of the other images to newer ones?

averevki:
> It builds. Maybe we could also update versions of the other images to newer ones?

Sure, what images? I think we should updated them only when needed/there was a big jump between versions

PR #718, opened 2025-07-31
Improve apply() logic

averevki:
Also removed related asserts from dns tests, leftover from https://github.com/Kuadrant/testsuite/pull/654

#### Verification steps
`make dnstls`, `make testsuite/tests/singlecluster/gateway/reconciliation/change_targetref`


azgabur:
Most important place where `modify_and_apply` is still used is here: https://github.com/Kuadrant/testsuite/blob/8a3813adf7119fb46ada32928e6f65dde6086272/testsuite/kubernetes/__init__.py#L116
was it migrated to `apply`?

PR #724, opened 2025-08-12
Refactor Wristband test to use add_custom_routes_match from EnvoyConfig

silvi-t:
## Description
This PR updates the Wristband test to use the add_custom_routes_match method from EnvoyConfig.

Previously, the test used the Wristband class to add a route match. This approach was overly complex, as the class was only needed for this one function.

Now, we use the add_custom_routes_match method, which is already used in other test. This makes our code more consistent, readable, and easier to maintain.


testsuite/tests/singlecluster/authorino/wristband/conftest.py
```diff
@@ -90,6 +74,31 @@ def wristband_hostname(exposer, gateway, blame):
...
+                    "value": '%DYNAMIC_METADATA(["envoy.filters.http.ext_authz", "wristband"])%',
+                }
+            }
+        ],
+    }
+    return route_dictionary
+
+
+@pytest.fixture(scope="module")
+def route(route, route_dictionary):
```
averevki, line 96: Isn't it enough to just edit the envoy config once? Both these routes point to the same gateway, so I'd think they have a single config for both

PR #740, opened 2025-09-02
OIDC policy

crstrn13:
**Summary**
Added OIDC policy extension tests for Authorino, covering different authentication flows and hostname configurations.

**Test Coverage**
- Public Client (PKCE flow)
- Service Client (M2M flow)
- Confidential Client (Auth code flow)

Each flow tested against:
- Gateway/Route targets
- Wildcard/FQDN hostnames
- Auth/Unauth scenarios

**Requirements**
- Requires Authorino with extensions support (`@pytest.mark.extensions`)
- Keycloak as OIDC provider


testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy.py
```diff
@@ -0,0 +1,85 @@
...
+@pytest.mark.parametrize("gateway", ["fully_qualified_domain_name", "wildcard_domain"], indirect=True)
+@pytest.mark.parametrize("oidc_client", ["public_client"], indirect=True)
+def test_public_client_flow(client, auth, oidc_client, gateway):
+    """Test public client with PKCE flow"""
+    # No auth -> redirect to login with PKCE
+    response = client.get("/")
+    assert response.status_code == 302
+    assert "Location" in response.headers
+    location = response.headers["Location"]
+    assert "response_type=code" in location  # Uses authorization  code flow
```
trepel, line 27: `authorization  code` - double space there

trepel (changes requested):
@crstrn13 Great work! The tests passed and there are only small typos there. I am just not sure about that `to_dict` methods so I pinged Aleks to see what he thinks about it.

testsuite/kuadrant/extensions/oidc_policy.py
```diff
@@ -0,0 +1,192 @@
...
+from dataclasses import dataclass
+from enum import Enum, auto
+from typing import Dict, Optional
+
+from enum import Enum, auto
+from testsuite.gateway import Referencable
+from testsuite.kubernetes import modify
+from testsuite.kubernetes.client import KubernetesClient
+from testsuite.kuadrant.policy import Policy
+from testsuite.utils import asdict
```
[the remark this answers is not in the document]
crstrn13: removed. :+1:

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy.py
crstrn13: adapted.

testsuite/kuadrant/extensions/oidc_policy.py
```diff
@@ -0,0 +1,192 @@
...
+    """Provider defines the settings related to the Identity Provider (IDP)"""
+
+    issuer_url: str
+    client_id: str
+    client_secret: Optional[str] = None
+    authorization_endpoint: Optional[str] = None
+    redirect_uri: Optional[str] = None
+    token_endpoint: Optional[str] = None
+
+    def to_dict(self) -> Dict:
```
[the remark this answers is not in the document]
crstrn13: ok, should be good now.

trepel (dismissed):
Nice! LGTM now.

@crstrn13 I leave merging this to you since we need to make sure that extensions are installed as part of nightlies first. Feel free to ping me for review for that work too.

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy.py
```diff
@@ -0,0 +1,85 @@
...
+    response = client.get("/", auth=auth)
+    assert response.status_code == 200
+    token = jwt.decode(auth.token.access_token, options={"verify_signature": False})
+    assert "openid" in token["scope"]  # Has OpenID scope
+    assert token["azp"] == oidc_client.client_id  # Correct client ID
+
+
+@pytest.mark.parametrize("oidc_policy", ["gateway", "route"], indirect=True)
+@pytest.mark.parametrize("gateway", ["fully_qualified_domain_name", "wildcard_domain"], indirect=True)
+@pytest.mark.parametrize("oidc_client", ["confidential_client"], indirect=True)
```
averevki, line 65: Does it really make sense to make oidc policy tests test every client with both hostname types and both gateway/route? I would guess there is enough code you test is overlapping in these 12 generated tests

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy.py
```diff
@@ -0,0 +1,85 @@
...
+- Service client (machine-to-machine)
+- Confidential client with authorization code
+"""
+
+from urllib.parse import quote
+
+import jwt
+import pytest
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.authorino, pytest.mark.extensions]
```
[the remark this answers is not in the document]
crstrn13: Correct, but you can also disable them. I think it would be beneficial to not run some tests are dependant on kuadrant-operator flag to enable the extensions.

Makefile
```diff
@@ -43,23 +43,35 @@ smoke: poetry-no-dev
...
+	$(PYTEST) -n4 -m 'authorino and not multicluster and not extensions' --dist loadfile --enforce $(flags) testsuite/tests/singlecluster
+
+authorino-extensions: ## Run only authorino related tests
+authorino-extensions: poetry-no-dev
+	$(PYTEST) -n4 -m 'authorino and not multicluster and extensions' --dist loadfile --enforce $(flags) testsuite/tests/singlecluster
 
 authorino-standalone: ## Run only test capable of running with standalone Authorino
 authorino-standalone: poetry-no-dev
-	$(PYTEST) -n4 -m 'authorino and not kuadrant_only' --dist loadfile --enforce --standalone $(flags) testsuite/tests/singlecluster/authorino
+	$(PYTEST) -n4 -m 'authorino and not kuadrant_only and not extensions' --dist loadfile --enforce --standalone $(flags) testsuite/tests/singlecluster/authorino
```
averevki: I still don't think we should run extensions as a separate mark right now, they are part of the kuadrant operator and this is just a redundant separation for me. I plan to refactor how we use marks in the testsuite either way, so it's good for now

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy_confidential_client.py
```diff
@@ -0,0 +1,222 @@
...
+    """Test confidential client authorization code flow with secrets and enhanced features."""
+    # Test unauthenticated request redirects
+    response = client.get("/")
+    assert response.status_code == 302, "Unauthenticated request should redirect"
+    assert "Location" in response.headers, "Redirect must include Location header"
+
+    location = response.headers["Location"]
+
+    # Validate OAuth2 redirect parameters
+    validate_oauth2_redirect(location, confidential_client, gateway)
```
averevki: I'm not a fan of hiding "complex" checks in tests behind helper functions

Makefile
crstrn13: well, you can run the kuadrant-operator without the extensions altogether. Thing is that we enable them by default. A mechanism to not trigger 'extensions' test is needed, as we need to test kuadrant in both modes - with and without extensions.

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy_service_client.py
```diff
@@ -0,0 +1,98 @@
...
+from testsuite.oidc.test_client import OIDCTestClient
+
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.authorino, pytest.mark.extensions]
+
+
+@pytest.fixture(scope="module")
+def test_client(request, keycloak, hostname):
+    """Create service OIDC test client"""
+    client, keycloak_client = OIDCTestClient.create_service_client(keycloak, hostname.hostname)
+    request.addfinalizer(lambda: keycloak.realm.delete_client(keycloak_client.client_id))
```
averevki: you usually set finalizers before the command you need finalizer for

testsuite/oidc/test_client.py
```diff
@@ -0,0 +1,187 @@
...
+from testsuite.oidc import Token
+from testsuite.kuadrant.extensions.oidc_policy import Provider
+
+
+@dataclass
+class ClientConfig:  # pylint: disable=too-many-instance-attributes
+    """Configuration for creating OIDC test clients"""
+
+    client_id: str
+    client_type: str  # "confidential", "public", "service"
```
averevki:
```suggestion
    client_type: Literal["confidential", "public", "service"]
```
if these are the only options

testsuite/oidc/test_client.py
```diff
@@ -0,0 +1,187 @@
...
+        keycloak_client = keycloak.realm.create_client(**config.to_keycloak_payload())
+
+        oidc_client = KeycloakOpenID(
+            server_url=keycloak.server_url,
+            client_id=keycloak_client.auth_id,
+            realm_name=keycloak.realm_name,
+            client_secret_key=keycloak_client.secret,
+        )
+
+        return cls(oidc_client, client_id, keycloak_client.secret), keycloak_client
```
averevki: why does it return 2 clients in a tuple?

testsuite/oidc/test_client.py
```diff
@@ -0,0 +1,187 @@
...
```
averevki: I see some nice patterns here. Maybe you can create a class for each of the clients with it's own initialization method

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy.py
```diff
@@ -0,0 +1,85 @@
...
+    assert response.status_code == 302
+    assert "x-ext-auth-reason" in response.headers
+    assert response.headers["x-ext-auth-reason"] == "credential not found"
+    location = response.headers["Location"]
+    assert "client_secret" not in location  # No sensitive info in redirect
+
+    # Valid auth -> success with correct token
+    response = client.get("/", auth=auth)
+    assert response.status_code == 200
+    token = jwt.decode(auth.token.access_token, options={"verify_signature": False})
```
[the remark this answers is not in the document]
averevki: I think you missed this one @crstrn13

testsuite/tests/singlecluster/extensions/oidc_policy/test_oidc_policy_service_client.py
crstrn13: we delete the realm at the end anyway. Removed the finalizer.

testsuite/oidc/test_client.py
averevki: I think you have more duplication in methods here than if you had one base class with a single client function that will be reused by client classes that inherit the base one. 

I think it is a better structure, don't need to do this though

PR #756, opened 2025-09-12
Fix kuadrantctl tests

### LINKED ISSUE #514: Fix kuadrantctl tests
Currently kuadrantctl tests do not work due to inaccessable fixtures. 

Missing fixtures are: backend, hostname, gateway

One solution is to move fixtures from `testsuite/tests/singlecluster/conftest.py` up a level to be accessable for tests in `testsuite/tests/kuadrantctl/` path or some other solution.

averevki:
Fix kuadrantctl tests. Added new test for shell completions.

Closes #514 

### Verification steps

Don't need to verify, it generates policies with old manifests. kuadrantctl is not being developed or updated currently. 

Basic commands and route generation tests still work though.

### DIFF
--- a/testsuite/kuadrantctl.py
+++ b/testsuite/kuadrantctl.py
@@ -1,6 +1,5 @@
-# pylint: disable=line-too-long
 """
-Help as of 0.2.3
+kuadrantctl v0.5.0
 Kuadrant configuration command line utility
 
 Usage:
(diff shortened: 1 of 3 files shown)


emmaaroche (approved):
Tests you said are passing are passing so LGTM 👍

PR #768, opened 2025-09-24
Add class for TelemetryPolicy

silvi-t:
## Description
Adds TelemetryPolicy class into testsuite that allows configuring telemetry metrics labels on Gateway resources.

Known Issues
- TODO: Delete wait_for_ready(self) method when TelemetryPolicy status is fixed.


crstrn13 (approved):
LGTM :tada: :shipit:

PR #791, opened 2025-10-14
Add model label support and improve TRLP metric test logic

emmaaroche:
### Description

This PR updates the TRLP metric tests to include validation for the new `model` label introduced in Kuadrant v1.3.0. It also adds relevant pytest markers and refactors test logic to reduce duplication of test code.

Closes https://github.com/Kuadrant/kuadrant-operator/issues/1626

### DIFF
--- a/testsuite/tests/singlecluster/limitador/metrics/trlp_metrics/test_trlp_metrics.py
+++ b/testsuite/tests/singlecluster/limitador/metrics/trlp_metrics/test_trlp_metrics.py
@@ -7,9 +7,13 @@ https://github.com/Kuadrant/kuadrant-operator/blob/a1ad64a2fb9230985230b57695ffa
 
 import pytest
 
+pytestmark = [pytest.mark.kuadrant_only, pytest.mark.limitador]
+
+USERS = ("free", "paid")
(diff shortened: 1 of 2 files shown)


testsuite/tests/singlecluster/limitador/metrics/trlp_metrics/test_trlp_metrics.py
```diff
@@ -18,46 +22,61 @@
...
     metrics = prometheus.get_metrics(
         labels={
             "pod": limitador.pod.name(),
             "limitador_namespace": f"{route.namespace()}/{route.name()}",
             "job": f"{pod_monitor.namespace()}/{pod_monitor.name()}",
-            "user": free_user_api_key.model.metadata.annotations["secret.kuadrant.io/user-id"],
-            "group": "free",
+            "user": user_info["user_id"],
+            "group": user_info["group"],
+            "model": MODEL_NAME,
```
trepel: You added this model here, should the docstring (comment) of this test function be updated to smt like `...for users/groups/models`. If I got it right this function also validates that `model` label was present with certain value now.

trepel (approved):
Great work! Tests passed for me and code looks good.

PR #796, opened 2025-10-21
refactor: Fix TokenRateLimitPolicy tests failing in parallel execution

### LINKED ISSUE #769: TRLP tests fail when run in parallel
### Description

In the nightly testsuite pipeline runs, some of the TokenRateLimitPolicy tests fail intermittently when executed in parallel (the pipeline runs with `-n4` to use 4 workers). However, when the same tests are run sequentially, they pass consistently.

### Observed Behavior

Requests are not being rate-limited at the expected point when multiple TRLP tests are active simultaneously.

Example failure from `test_multiple_trlp_limit_iterations`:

```
Iteration 6/10: Expected 429 after 52/50 tokens, but got 200
E   AssertionError: Iteration 6/10: Expected 429 after 52/50 tokens, but got 200
```

How it should work is that requests starting below the limit and then pushing just past it are still accepted, and only the following requests get rate limited. From the error messages it looks like the 429 (rate-limited) response is only returned at the moment the token limit is crossed, when it should actually trigger once the limit has been met or passed. This inconsistent behavior only occurs when multiple TRLP tests are running at the same time, even though each test is designed to send the “over-limit” request only after the token quota has been fully consumed.

### Steps to Reproduce

1. Run tests sequentially:

   ```bash
   poetry run pytest testsuite/tests/singlecluster/limitador/tokens_rate_limit
   ```

   ✅ All tests pass.

2. Run tests in parallel:

   ```bash
   poetry run pytest -n4 testsuite/tests/singlecluster/limitador/tokens_rate_limit
   ```

   ❌ Failures appear intermittently.

### Expected Behavior

Tests should pass consistently regardless of whether they are run sequentially or in parallel.

Anton-Fil:
## Description
closes #769 

**Problem Description**
The TokenRateLimitPolicy tests in both `basic/ and multiple_iterations/ directories `were failing when executed in parallel, while working correctly when run sequentially. The issue occurred because:

1. The basic tests `(test_trlp.py, test_trlp_streaming.py)` use shared authentication fixtures with identical user IDs, while multiple iterations tests use anonymous requests but still compete for the same rate limit quota
2. The TokenRateLimitPolicy is configured to track rate limits per user ID using `counters=[CelExpression("auth.identity.userid")]` for basic tests, and per IP/request for multiple iterations tests
3. When tests run in parallel, they simultaneously consume tokens from shared limits (50 tokens/30s for basic tests, 50 tokens/20s for multiple iterations), causing `unexpected 200 `responses instead of the `expected 429` rate limit responses, as the tests were interfering with each other's token consumption patterns

**Attempted Solutions (That Didn't Work)**
Several approaches were considered but not implemented:

- Attempted to use `pytest-xdist` worker IDs to create unique user IDs per worker, but this approach failed
- Tried creating separate user fixtures for each test file, but this required changes to the shared conftest.py and authorization configuration that weren't feasible
- Attempted to set up different backends for each test module to isolate rate limiting, but this approach gave the same failures and added complexity to the setup

**Final Solution**
Modified the test logic to be more resilient to parallel execution by:

- Changed the while loops to bounded for loops so the test now stops when a `429` response is received, instead of waiting to reach the exact token limit. The original while `total_tokens < LIMIT` loop relied on precise token accounting, but in parallel runs the shared rate-limit counters caused small timing differences between workers. This led to inconsistent token tracking, where some threads consumed tokens faster or slower, causing the loop to exit early before the actual limit was reached.
- Set reasonable upper bounds (`max_requests = 20` for most tests, `max_requests = 30` for paid user tests) to prevent infinite loops
- Tests now break the loop immediately when rate limit (429) is encountered, rather than continuing to consume tokens
- Added `pytest.fail()` with descriptive messages when rate limits aren't hit within the expected request count
- Added a readiness check to make sure the route is ready before committing the policy, preventing “route not attached to existing path” errors.
- Reduced the iteration range (from 10→5 and 5→3) to speed up test execution while still running enough cycles to verify TokenRateLimitPolicy reset and enforcement functionality

### DIFF
--- a/testsuite/tests/singlecluster/limitador/tokens_rate_limit/basic/test_trlp.py
+++ b/testsuite/tests/singlecluster/limitador/tokens_rate_limit/basic/test_trlp.py
@@ -48,10 +48,19 @@ def test_trlp_limit_and_reset_free_user(client, free_user_auth):
     assert tokens_used > 0 and tokens_used == usage["prompt_tokens"] + usage["completion_tokens"]
     total_tokens += tokens_used
 
-    # Keep sending requests while within token quota
-    while total_tokens < FREE_USER_LIMIT.limit:
+    # Keep sending requests until the rate limit is hit
+    hit_limit = False  # Track if the rate limit has been hit (received 429)
+    max_requests = 10  # Safety limit to prevent infinite loop
+    for request_num in range(max_requests):
         response = client.post("/v1/chat/completions", auth=free_user_auth, json={**basic_request})
-        assert response.status_code == 200
+        if response.status_code == 429:
+            hit_limit = True
+            break
+        assert (
+            response.status_code == 200
+        ), f"Request {request_num + 1} got {response.status_code} at {total_tokens}/{FREE_USER_LIMIT.limit} tokens"
+
+        # Track token usage
         json_data = response.json()
         usage = json_data.get("usage")
         assert usage and all(key in usage for key in ("total_tokens", "prompt_tokens", "completion_tokens"))
@@ -59,11 +68,8 @@ def test_trlp_limit_and_reset_free_user(client, free_user_auth):
         assert tokens_used > 0 and tokens_used == usage["prompt_tokens"] + usage["completion_tokens"]
         total_tokens += tokens_used
 
-    # Next request should be 429
-    response = client.post("/v1/chat/completions", auth=free_user_auth, json={**basic_request})
-    assert (
-        response.status_code == 429
-    ), f"Expected 429 after {total_tokens}/{FREE_USER_LIMIT.limit} tokens, but got {response.status_code}"
+    # Verify the rate limit was hit
+    assert hit_limit, f"Rate limit was never hit after {total_tokens}/{FREE_USER_LIMIT.limit} tokens"
 
     # Assert quota resets after wait period
     sleep(30)
@@ -84,9 +90,17 @@ def test_trlp_limit_and_reset_paid_user(client, paid_user_auth):
     assert tokens_used > 0 and tokens_used == usage["prompt_tokens"] + usage["completion_tokens"]
     total_tokens += tokens_used
 
-    while total_tokens < PAID_USER_LIMIT.limit:
+    hit_limit = False
+    max_requests = 20
+    for request_num in range(max_requests):
         response = client.post("/v1/chat/completions", auth=paid_user_auth, json={**basic_request})
-        assert response.status_code == 200
+        if response.status_code == 429:
+            hit_limit = True
+            break
+        assert (
+            response.status_code == 200
+        ), f"Request {request_num + 1} got {response.status_code} at {total_tokens}/{PAID_USER_LIMIT.limit} tokens"
+
         json_data = response.json()
         usage = json_data.get("usage")
         assert usage and all(key in usage for key in ("total_tokens", "prompt_tokens", "completion_tokens"))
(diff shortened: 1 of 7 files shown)


testsuite/tests/singlecluster/limitador/tokens_rate_limit/multiple_iterations/test_multiple_trlp_iterations.py
```diff
@@ -20,28 +19,39 @@
...
+        else:
+            # Loop completed without hitting 429
+            pytest.fail(f"Iteration {i+1}/5: Never hit rate limit after {max_requests} requests")
 
+        # Verify requests are rate limited
         response = client.post("/v1/chat/completions", json={**basic_request})
         assert (
             response.status_code == 429
-        ), f"Iteration {i+1}/10: Expected 429 after {total_tokens}/{LIMIT.limit} tokens, but got {response.status_code}"
+        ), f"Iteration {i+1}/5: Expected 429 to still be enforced, but got {response.status_code}"
```
emmaaroche:
``` suggestion
        ), f"Iteration {i+1}/5: Expected 429 after {total_tokens}/{LIMIT.limit} tokens, but got {response.status_code}"

**Comment by Anton-Fil** [comment]
Hey @emmaaroche , I applied the suggested changes but the test fails with:

`Failed: Never hit rate limit after 15 requests.`

I also refined tests further for better stability. Here’s what I added:

- `module_label` to user IDs to isolate rate limit counters per worker — this should prevent counter conflicts in parallel runs.
- also i added a` hit_limit` flag to explicitly verify that 429 responses were received, ensuring more deterministic checks for rate limit enforcement.

**Review by azgabur — COMMENTED** [review]
I think the whole problem comes down to the `token_rate_limit` fixture and the structure of test modules. The fixture is `module` scoped. This can be problematic for rate limit tests as multiple test functions inside one module will use the same TRLP which obviously will cause interference. Thats why most limitador tests contain only one test function per module. 
To keep the current setup I suggest changing the scope for `token_rate_limit` and `commit` fixtures to `function`. That will force re-creation of TRLP and AuthPolicy for each test function effectively re-setting the counter but at the cost of re-creating and waiting for reconcile.

Another solution is to separate test functions to multiple modules, that will cause the xdist library to run these in parallel as well as each having their own TRLP which may be in the end faster as it divides the work among workers per-module basis. I will leave that decision on you. But especially if you use sleeps than parallelization should be preferred. 

See how the changed scope for RLP and commit fixture was done in: `testsuite/tests/singlecluster/limitador/storage`

I tested this simple change on `main` but I still got some test fails so its probably not this simple.

**Comment by emmaaroche** [comment]
Thanks for the suggestions @azgabur. Following your feedback, @Anton-Fil and I experimented with several approaches:

- Created separate test modules for each test case
- Defined a unique TRLP per module
- Used different limit configurations per TRLP
- Tried switching to function-scoped fixtures
- Tested separate TRLPs for route- and gateway-level setups

Unfortunately, the same TRLP issues are still showing up after making these changes 😢

**Comment by crstrn13** [comment]
@emmaaroche @Anton-Fil 
I have noticed that in the `multiple-iterations/conftest.py` we do not set any counter on the TRLP. Is this by design or is it interfering with the test logic itself if we added a counter there? The counter doesn't have to be auth-specific, I think we could add a header to all of the requests in that test for instance and count on that, for instance.

Thing is I liked more the way we asserted for roughly the number of tokens, rather that sending requests until we run out of quota. Does that make any sense? Let me know

File: testsuite/tests/singlecluster/limitador/tokens_rate_limit/basic/test_trlp.py [hunk 2]
```diff
@@ -48,22 +48,30 @@ def test_trlp_limit_and_reset_free_user(client, free_user_auth):
...
+
+        # Track token usage
         json_data = response.json()
         usage = json_data.get("usage")
         assert usage and all(key in usage for key in ("total_tokens", "prompt_tokens", "completion_tokens"))
         tokens_used = usage["total_tokens"]
         assert tokens_used > 0 and tokens_used == usage["prompt_tokens"] + usage["completion_tokens"]
         total_tokens += tokens_used
+    else:
+        pytest.fail(f"Never hit rate limit after {max_requests} requests")
```
emmaaroche: I think you could remove the `else` after the request loop in each test as the `assert hit_limit` already covers the failure case if the rate limit isn’t reached.

PR #799, opened 2025-10-22
Refactor gateway listener in dns section target test

azgabur:
The test worked previously too, but the gateway setup had one flaw. Resulting gateway had the default wildcard listener and the additional managed and unmanaged listeners, but these were http only. The client fixtures were using https protocol so it actually connected to the wildcard listener instead, not the managed or unamanaged listener. 

I know this is a small change and it has no impact on the test result but for the future if the test would be changing it would solve some debugging time..

to verify:
```
make testsuite/tests/singlecluster/gateway/dnspolicy/test_dns_policy_section_targeting_gateway.py
```

### DIFF
--- a/testsuite/tests/singlecluster/gateway/dnspolicy/test_dns_policy_section_targeting_gateway.py
+++ b/testsuite/tests/singlecluster/gateway/dnspolicy/test_dns_policy_section_targeting_gateway.py
@@ -15,7 +15,6 @@ from testsuite.gateway import CustomReference, GatewayListener
 from testsuite.gateway.gateway_api.gateway import KuadrantGateway
 from testsuite.gateway.gateway_api.hostname import StaticHostname
 from testsuite.gateway.gateway_api.route import HTTPRoute
-from testsuite.httpx import KuadrantClient
 from testsuite.kuadrant.policy.dns import DNSPolicy
 from testsuite.utils import is_nxdomain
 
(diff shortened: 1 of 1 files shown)


testsuite/tests/singlecluster/gateway/dnspolicy/test_dns_policy_section_targeting_gateway.py
```diff
@@ -68,8 +68,12 @@ def gateway(gateway: KuadrantGateway, managed_domain, unmanaged_domain):
...
     """
 
-    gateway.add_listener(GatewayListener(name=MANAGED_LISTENER_NAME, hostname=managed_domain))
-    gateway.add_listener(GatewayListener(name=UNMANAGED_LISTENER_NAME, hostname=unmanaged_domain))
+    gateway.add_listener(
+        TLSGatewayListener(name=MANAGED_LISTENER_NAME, hostname=managed_domain, gateway_name=gateway.name())
+    )
+    gateway.add_listener(
+        TLSGatewayListener(name=UNMANAGED_LISTENER_NAME, hostname=unmanaged_domain, gateway_name=gateway.name())
+    )
```
averevki: Shouldn't this test NOT have wildcard listener at all in the first place?

averevki (commented):
Good catch, but why do you want to use tlspolicy together with dnspolicy in this test? It should only test the section name change in dnspolicy. More _atomic_, if you will :grin:

azgabur:
Yes another solution is to make the clients http, which is maybe better solution come to think of it :thinking: 

The test can have wildcard listener as the more specific hostname listener will be used instead of the wildcard when a request comes, of course not if they are on different ports as happened here.. 

You could remove the default listener and have only the managed and unmanaged ones but it makes little difference. I will add it regardless

fabikova:
@azgabur I'm looking at the last CI run and I don't quite understand the failure - dns.resolver.NoAnswer. Could you explain what this failure means, please?

fabikova:
thanks for explanation!

PR #810, opened 2025-11-12
feat(ci): add conventional pr title validation

crstrn13:
## Description
Added conventional PR title checker that validates if the title is following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format

## Changes
- Added Github workflow
- Modified pull_request_template with the PR title guidelines
- Updated Claude.md

## Verification
- Title in unconventional format
- Title in conventional format.

### DIFF
--- a/.github/workflows/validate-pr-title.yml
+++ b/.github/workflows/validate-pr-title.yml
@@ -37,23 +37,7 @@ jobs:
             chore
             revert
 
-          # Configure which scopes are allowed
-          scopes: |
-            authorino
-            chore
-            ci
-            dns
-            docs
-            gateway
-            limitador
-            multicluster
-            perf
-            refactor
-            style
-            test
-            tls
-
-          # Require a scope (set to true to make scopes mandatory)
+          # Scope requirement: currently optional (set to true to make mandatory
           requireScope: false
 
           validateSingleCommit: false


CLAUDE.md
```diff
@@ -51,6 +51,42 @@ make mypy                # Run type checking
 make clean               # Delete all test-created resources (uses USER env var)
 ```
 
+## Pull Request Guidelines
+
+### PR Title Format (Conventional Commits)
```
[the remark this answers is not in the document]
crstrn13: that makes perfect sense. I moved it in there, thanks for the suggestion! 👍

.claude/commands/pr-description.md
```diff
@@ -2,6 +2,44 @@
 description: Generate a comprehensive PR description based on git diff and commit history
 ---
 
+## PR Title Format (Conventional Commits)
+
+Since PRs are squash-merged, the PR title becomes the commit message. All PR titles **must** follow conventional commit format:
```
averevki: We don't always squash PRs in team. I'd like an option to have PR that consists of multiple commits, each of separate type. But I guess PR title still can represent a short description of everything done in a this single PR, even if there are multiple commits

crstrn13: the point is to end up with conventional commits on the main branch. This would work only if enforcing squashing and merge. For the most part, I can see that most of the PRs are like this. Moreover squash & merge enforcement can be done in github to fit better with this behavior.

crstrn13: what I was referring to was, was to squash and merge all the commits from a PR into the PR title. This repository currently allows 3 options, but the best way this PR title convention would work is if it [squashed and merged](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) all of your commits into the PR title (which is conventional) and this will be the only commit message you will end up seeing in the main branch.

Another solution would be to do that at commit level, so every commit must follow the convention, but we kinda encourage to squash commits on every PR anyway so I think the first option is more feasible and easier to adopt.
Anyway, this can always be changed if we decide it is not good enough.

crstrn13: but here is a very good example: https://github.com/vuejs/core/commits/main/

averevki: Yeah I get it, does it matter if a single merged commit then will consist of multiple commits from the PR (commits can be from different scopes) and will be larger? Why not just merge these commits to main, no squashing? I just don't want squashing to be enforced if having the right naming is the only reason.

> Or does the content of the post-merged commits not matter?

At the end, we can have both options as we have had until now and have our PRs and commits both named by the convention, no?

emmaaroche (approved):
LGTM 👍 

Just one tiny nitpick (feel free to ignore 🙈): add a closing bracket to comment: `# Scope requirement: currently optional (set to true to make mandatory)`

PR #822, opened 2025-12-01
feat(dns): Add smarter wait for changing dns zone asserts

azgabur:
## Description
Work on https://github.com/Kuadrant/testsuite/issues/660

To introduce smarter waits for DNS related waiting. Not sure I covered all places in the codebase that would benefit from this util function. To benefit from https://github.com/Kuadrant/helm-charts-olm/pull/60 this refactor was needed.

## Changes
- Added `wait_for_dns` util method which does dns query with retries and expected result.
- Refactored two tests which had `sleep(300)` previously

## Verification
Just two tests affected.
```sh
make testsuite/tests/multicluster/load_balanced/test_change_strategy.py testsuite/tests/multicluster/load_balanced/test_change_default_geo.py
```

### DIFF
--- a/testsuite/tests/multicluster/load_balanced/test_change_default_geo.py
+++ b/testsuite/tests/multicluster/load_balanced/test_change_default_geo.py
@@ -1,10 +1,9 @@
 """Test for modification of default geolocation in DNSPolicy"""
 
-from time import sleep
-
 import pytest
 import dns.resolver
 from testsuite.config import settings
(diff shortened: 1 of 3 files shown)


fabikova (commented):
Nice, LGTM. I added just a single comment about TTL logic. Could you please explain if this is a valid concern? I'm not 100% sure if the backoff handles this.

PR #910, opened 2026-03-26
fix: add retry logic to kuadrant metrics tests

silvi-t:
## Description
  - Fix intermittent failures observed in nightly test pipeline for `test_metric_kuadrant_policies_enforced` where `kuadrant_policies_enforced` metric shows `0.0` for `RateLimitPolicy` and `TokenRateLimitPolicy`
  - Fix intermittent failures observed in nightly test pipeline for `test_metric_policy_lifecycle` where metrics don't reflect policy deletion in time
 - Possible root cause: the kuadrant-operator updates policy metrics from an in-memory topology snapshot that may lag one reconciliation cycle behind the actual policy state in etcd, causing a race condition between policy
  enforcement and metric updates. Policies committed last (`RateLimitPolicy`, `TokenRateLimitPolicy`) are most affected as they have less time for the operator to complete the follow-up reconciliation cycle

### New Features
  - Add `wait_for_metric` method to `Prometheus` class that polls a metric value with backoff retries, supporting custom comparisons via `operator` module (e.g. `operator.eq`, `operator.ge`, `operator.lt`)

  ### Bug Fixes
  - Replace single-shot metric assertions with `prometheus.wait_for_metric` in `test_metric_kuadrant_policies_enforced` to wait for Prometheus to reflect enforced status
  - Replace `wait_for_scrape` with `prometheus.wait_for_metric` in `test_metric_policy_lifecycle` for both policy creation and deletion metric checks
  - Add `request.addfinalizer(policy.delete)` to `test_metric_policy_lifecycle` to prevent policy resource leaks when assertions fail before `policy.delete()` is reached

  ### Refactoring
  - Move `service_monitor` fixture to `conftest.py` as `kuadrant_service_monitor` for reuse across test modules
  - Add `wait_for_scrape` to `kuadrant_operator_metrics` fixture in `conftest.py` to ensure fresh metrics after policy commit

## Summary by CodeRabbit

* **Tests**
  * Added a polling-based Prometheus wait that retries with configurable comparison, handles absent or multiple series, and surfaces latest observed values on failure.
  * Tests now explicitly wait for the operator service scrape readiness before validating metrics and use the new polling approach for lifecycle assertions.
  * Improved teardown sequencing to ensure resources are deleted reliably.

* **Refactor**
  * Consolidated and centralised metric label construction for consistent observability queries.

### DIFF
--- a/testsuite/prometheus.py
+++ b/testsuite/prometheus.py
@@ -1,5 +1,6 @@
 """Simple client for the Prometheus metrics"""
 
+import operator
 from datetime import datetime, timezone
 from typing import Callable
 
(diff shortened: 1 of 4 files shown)


crstrn13 (approved):
LGTM 🎉 :shipit:

PR #921, opened 2026-04-06
feat: add Prometheus observability to local-setup

crstrn13:
## Summary

Implements #906 - Adds Prometheus monitoring stack to `make local-setup` for local observability testing and debugging.

### Infrastructure
- **Prometheus Stack**: Deploys full `kube-prometheus-stack` Helm chart (v67.4.0) instead of just CRDs
  - Includes Prometheus, Grafana, and all monitoring components
  - Exposed via MetalLB LoadBalancer for local access
  - Enabled by default with `INSTALL_PROMETHEUS=true`
  
- **Kuadrant Observability**: Enables observability in Kuadrant CR when Prometheus is installed
  - Sets `spec.observability.enabled: true`
  - Enables ServiceMonitor/PodMonitor scraping

### Configuration
- **Prometheus Fixture Refactored**: testsuite/tests/conftest.py:131-186
  - Config-based connection with OpenShift auto-discovery fallback
  - Two helper functions for clean separation of concerns:
    - `_try_configured_prometheus(testconfig)`: Uses config URL if provided
    - `_try_openshift_prometheus(cluster)`: Auto-discovers thanos-querier route
  - Specific exception handling (no broad catch-all)
  
- **Settings**: Added prometheus.url configuration
  - Kind: Manual LoadBalancer URL via `make get-prometheus-url`
  - OpenShift: Auto-discovery (no config needed)

### Documentation
- Updated CLAUDE.md and README.md with:
  - Prometheus setup instructions
  - How to get LoadBalancer URL for Kind
  - Configuration examples for both platforms

## Test Results

**Observability tests (126 total):**
- 83 passed - All Prometheus metrics tests working
- 3 failed - TokenRateLimitPolicy metrics (not deployed in test setup)
- 40 skipped - Tracing tests (require Jaeger/Tempo, not included by default)

**Passing test suites:**
- Kuadrant operator health metrics (16 tests)
- Kuadrant policy metrics (12 tests)
- ServiceMonitor/PodMonitor scraping (18 tests)
- Authorino metrics (21 tests)
- Limitador metrics (9 tests)
- Telemetry policy (6 tests)

## Acceptance Criteria

- [x] `make local-setup` deploys Prometheus stack
- [x] Kuadrant CR has `observability.enabled: true` when Prometheus is installed
- [x] Metrics accessible via Prometheus LoadBalancer UI
- [x] Observability tests pass: 83/86 prometheus tests passing (3 failures unrelated to Prometheus)
- [x] Documentation updated with setup instructions

## Usage

```bash
# Default: Prometheus enabled
make local-setup

# Get Prometheus URL for Kind
make get-prometheus-url
# Output: http://172.18.255.200:9090

# Configure in config/settings.local.yaml
default:
  prometheus:
    url: "http://172.18.255.200:9090"

# Run observability tests
make observability
```

## Notes

- Prometheus is enabled by default (`INSTALL_PROMETHEUS=true`)
- Can be disabled with `INSTALL_PROMETHEUS=false make local-setup`
- OpenShift auto-discovers Prometheus (no manual config needed)
- Kind requires manual LoadBalancer URL configuration

Generated with Claude Code

## Summary by CodeRabbit

* **New Features**
  * Prometheus monitoring stack is enabled by default; observability can be auto-enabled when installed.

* **Documentation**
  * Local and OpenShift setup guidance updated to reflect Prometheus-on-by-default, access instructions, and how to override the local Prometheus URL.

* **Chores**
  * Added a CLI helper to discover/display the Prometheus web URL and new variables/flags to control Prometheus installation.

* **Tests**
  * Test fixtures updated to prefer configured Prometheus endpoints, then fall back to service discovery with clearer skip/fail behaviour.


README.md
```diff
@@ -55,8 +55,8 @@ make local-setup
 # Optional: Apply additional manifests (e.g., DNS provider credentials, secrets, etc.)
 ADDITIONAL_MANIFESTS=./my-secrets.yaml make local-setup
 
-# Optional: Install Prometheus CRDs for observability testing
-INSTALL_PROMETHEUS=true make local-setup
+# Prometheus is enabled by default. To disable:
+INSTALL_PROMETHEUS=false make local-setup
```
emmaaroche, line 59: Nitpick comment but setting `INSTALL_PROMETHEUS=false` still installs the Prometheus CRDs, just not the full stack (from what I see in the `else` block in `make/dependencies.mk`). Maybe we could clarify the wording in the README, something like:

```
# Prometheus is enabled by default. To disable the full stack (installs CRDs only):
INSTALL_PROMETHEUS=false make local-setup
```
That way it's more clear that `false` doesn't skip Prometheus entirely.

CLAUDE.md
```diff
@@ -71,6 +72,34 @@ make mypy                # Run type checking
 make clean               # Delete all test-created resources (uses USER env var)
 ```
 
+### Accessing Prometheus
+
+**OpenShift:** Prometheus auto-discovers the thanos-querier route. No configuration needed.
+
+**Kind/Local Setup:** When Prometheus is enabled (default), configure the LoadBalancer URL:
+
+1. **Get the Prometheus URL:**
```
emmaaroche, line 81: This is helpful reference material but I'm concerned about the `CLAUDE.md` growing too large if we were to update it like this for every PR. The docs recommend keeping it under 200 lines and splitting content into rule files if growing large, as longer files consume more context and reduce adherence, see https://code.claude.com/docs/en/memory#write-effective-instructions.  

Maybe we could move the "Accessing Prometheus" section to a `.claude/rules/prometheus-setup.md` file, wdyt?

testsuite/tests/conftest.py
```diff
@@ -128,31 +128,60 @@ def testconfig():
...
-    # Check if metrics are enabled
+def _try_configured_prometheus(testconfig):
+    """Try to create Prometheus client from configured URL"""
+    try:
+        testconfig.validators.validate(only=["prometheus"])
+        url = testconfig["prometheus"]["url"]
+        client = KuadrantClient(base_url=url, verify=False)
+        return Prometheus(client)
+    except (KeyError, ValidationError):
+        return None
```
silvi-t: The `prometheus.url` config key referenced here is not documented in `config/settings.local.yaml.tpl` or `config/settings.yaml`, so users have no way to know they need to set it. On kind, even though `make local-setup`       installs the Prometheus stack, all observability tests skip because neither `_try_configured_prometheus` (no config) nor `_try_openshift_prometheus` (no OpenShift) can find it.                                                 
                                                                                                                                                                                                                                   
  Two options:
  1. Add `prometheus.url` to the settings template so users know to configure it manually                                                                                                                                          
  2. Add a kind auto-discovery fallback (similar to the OpenShift one) that looks up the Prometheus LoadBalancer service in the `monitoring` namespace at runtime — this would make `make local-setup` + `make observability` work out of the box without any manual config step

README.md
crstrn13: yeah, I decided on that becaue `make clean` is trying to delete some resources that were not installed if the prom crds were not installed beforehand. I agree that requires some explanation.

CLAUDE.md
crstrn13: sure, fair point.

testsuite/tests/conftest.py
crstrn13: I documented the change in the template now. Thanks for pointing it out.

.claude/rules/prometheus-setup.md
```diff
@@ -0,0 +1,35 @@
+# Accessing Prometheus
+
+## OpenShift
+
+Prometheus auto-discovers the thanos-querier route. No configuration needed.
+
+## Kind/Local Setup
+
+When Prometheus is enabled (default), the testsuite auto-discovers the Prometheus LoadBalancer service in the `monitoring` namespace. No manual configuration is needed if `make local-setup` was used.
```
silvi-t: Looking at the `prometheus` fixture in `testsuite/tests/conftest.py`, there are only two discovery paths:
                                                                                                                                                                                                                                   
  1. `_try_configured_prometheus` — reads `prometheus.url` from config                                                                                                                                                             
  2. `_try_openshift_prometheus` — auto-discovers thanos-querier on OpenShift                                                                                                                                                      
                                                                                                                                                                                                                                   
  There's no kind/LoadBalancer auto-discovery implemented. On a local kind cluster after `make local-setup`, observability tests will skip unless the user manually sets `prometheus.url` in `config/settings.local.yaml`. You could add a similar auto-discovery method for kind, like `_try_kind_prometheus`, following the same approach as `_try_openshift_prometheus`.

.claude/rules/prometheus-setup.md
```diff
@@ -0,0 +1,35 @@
...
+## Running Observability Tests
+
+```bash
+make observability
+```
+
+## Prometheus UI
+
+- OpenShift: Access thanos-querier route in browser
+- Kind: Open the LoadBalancer URL in your browser to explore metrics
```
silvi-t: Missing trailing newline

testsuite/tests/conftest.py
silvi-t: I left a comment about the kind auto-discovery approach in `.claude/rules/promethues-setup.md` file.

.claude/rules/prometheus-setup.md
crstrn13: I could add it, but none of the other rule files have a trailing new line.

.claude/rules/prometheus-setup.md
crstrn13: I refactored that part a bit. Moved the discovery logic under the exposer interface. I kept the configured prometheus option in case we need to test in an environment where prometheus is deployed differently.

make/dependencies.mk
```diff
@@ -62,12 +62,45 @@ create-cluster-issuer: ## Create self-signed ClusterIssuer for TLS testing
...
+	@echo ""
+	@echo "Waiting for LoadBalancer IP assignment..."
+	@kubectl wait --for=jsonpath='{.status.loadBalancer.ingress[0].ip}' \
+		svc/prometheus-kube-prometheus-prometheus -n $(PROMETHEUS_NAMESPACE) --timeout=60s 2>/dev/null || true
+	@PROM_URL=$$(kubectl get svc -n $(PROMETHEUS_NAMESPACE) prometheus-kube-prometheus-prometheus \
+		-o jsonpath='http://{.status.loadBalancer.ingress[0].ip}:9090' 2>/dev/null); \
+	if [ -n "$$PROM_URL" ]; then \
+		echo "Prometheus accessible at: $$PROM_URL"; \
+		echo ""; \
+		echo "To run observability tests, add to config/settings.local.yaml:"; \
```
silvi-t: Now that Prometheus is auto-discovered via `exposer.prometheus_client()`, this manual config instruction is no longer needed and can be removed.

config/settings.local.yaml.tpl
```diff
@@ -31,6 +31,8 @@
 #    backend: "jaeger"                          # Tracing backend
 #    collector_url: "rpc://jaeger-collector.com:4317"  # Tracing collector URL (may be internal)
 #    query_url: "http://jaeger-query.com"       # Tracing query URL
+#  prometheus:
+#    url: "http://172.18.255.200:9090"            # Overrides pre-defined auto-discovery (OpenShift/Kind), use with care
```
averevki, line 35: Does dynoconf validator that fetches IP from the monitoring service automatically, same as other service url's that can be fetched by default, make sense to also have here? We have setting validators configured inside of `testsuite/config/__init__.py`, there should be some helper functions to cast service name to its loadbalancer IP already

make/dependencies.mk
crstrn13: done :+1:

config/settings.local.yaml.tpl
crstrn13: good idea, added :+1:

testsuite/tests/singlecluster/authorino/metrics/test_metrics.py
```diff
@@ -39,41 +40,49 @@
...
 @pytest.fixture(scope="module")
-def metrics_keys(authorino, service_monitor, prometheus, client, auth):
-    """
-    Send a simple get request so that a few metrics can appear and
-    return all metrics defined for the Authorino metrics service
-    """
+def metrics_labels(authorino, service_monitor, prometheus, client, auth):
+    """Send a request to generate metrics and return Prometheus query labels"""
+    prometheus.wait_for_scrape(service_monitor, "/metrics")
+    prometheus.wait_for_scrape(service_monitor, "/server-metrics")
```
crstrn13: changing the order here because on kind, the metric won't get registered. Thus, I am waiting for a scrape first, because I reworked the `metrics_keys` to retry until it finds the metrics. Tested on both kind and ocp and had satisfactory results.

testsuite/config/tools.py
```diff
@@ -52,18 +52,18 @@ def _fetcher(settings, _):
...
 
-def fetch_service_ip(name, port: int, protocol: str = "http"):
-    """Fetched load balanced ip for LoadBalancer service"""
+def fetch_service_ip(name, port: int, protocol: str = "http", namespace: str = "tools"):
+    """Fetches load balanced IP for a LoadBalancer service in a given namespace"""
 
     def _fetcher(settings, _):
         try:
-            cluster = settings["tools"]
+            cluster = settings["control_plane"]["cluster"].change_project(namespace)
```
silvi-t: I don't understand the reason for this change. The previous approach using `settings["tools"]` was more flexible — the tools project/namespace was configurable via `config/settings.yaml` without touching any code. With this  
  change, the namespace is hardcoded as a default parameter (`namespace="tools"`), so if someone needs a different tools namespace they'd have to change it in the code rather than just updating the settings file.               
                                                                                                                                                                                                                                  
Was there a specific reason for switching from `settings["tools"]` to `settings["control_plane"]["cluster"].change_project(namespace)`?

testsuite/gateway/exposers.py
```diff
@@ -78,6 +104,17 @@ def expose_hostname(self, name, gateway: Gateway) -> Hostname:
...
         return "test.com"
 
+    def prometheus_client(self):
+        """Discover Prometheus via LoadBalancer service in monitoring namespace"""
+        try:
+            monitoring = self.cluster.change_project(self.PROMETHEUS_NAMESPACE)
+            with monitoring.context:
+                svc = selector(f"service/{self.PROMETHEUS_SERVICE}").object(cls=Service)
+                port = svc.model.spec.ports[0].port
+                return KuadrantClient(base_url=f"http://{svc.external_ip}:{port}", verify=False)
```
silvi-t: With the `DefaultValueValidator` auto-discovery in `config/__init__.py`, this `prometheus_client()` method and the `PROMETHEUS_SERVICE`/`PROMETHEUS_NAMESPACE` constants are now redundant and can be removed.

.claude/rules/prometheus-setup.md
```diff
@@ -0,0 +1,30 @@
+# Accessing Prometheus
+
+## Auto-Discovery
+
+Prometheus is auto-discovered via the `Exposer.prometheus_client()` method:
+
+- **OpenShift (`OpenShiftExposer`)**: Looks up the `thanos-querier` route in `openshift-monitoring`. No configuration needed.
+- **Kind/Kubernetes (`LoadBalancerServiceExposer`)**: Looks up the `prometheus-kube-prometheus-prometheus` LoadBalancer service in the `monitoring` namespace. Works out of the box if `make local-setup` was used.
```
silvi-t: The Kind auto-discovery description here doesn't match the actual flow. Would  be good to update this to reflect the current discovery path.

testsuite/tests/conftest.py
```diff
@@ -226,29 +224,27 @@ def testconfig():
...
-    Return an instance of Thanos metrics client
-    Skip tests if query route is not properly configured
-    """
-    openshift_monitoring = cluster.change_project("openshift-monitoring")
-    # Check if metrics are enabled
+def prometheus(cluster, testconfig, skip_or_fail):
+    """Prometheus metrics client, auto-discovered via DefaultValueValidator or configured explicitly"""
+    prometheus_conf = testconfig.get("prometheus", {})
+    project = prometheus_conf.get("project", "unknown")
+    service = prometheus_conf.get("service", "unknown")
```
averevki: You can also set dynaconf validators on project and service keys, to verify they are present for the test

testsuite/gateway/exposers.py
```diff
@@ -78,6 +98,14 @@ def expose_hostname(self, name, gateway: Gateway) -> Hostname:
     def base_domain(self) -> str:
         return "test.com"
 
+    def prometheus_url(self, project, service_name):
```
averevki: I feel like this new method doesn't fit the existing exposer classes objective. I'd keep prometheus url processing inside the prometheus fixture.

testsuite/config/tools.py
```diff
@@ -69,6 +69,23 @@ def _fetcher(settings, _):
...
 
+def fetch_prometheus_url():
+    """Auto-discovers Prometheus URL using the exposer's platform-specific discovery"""
+
+    def _fetcher(settings, _):
+        try:
+            prometheus = settings["prometheus"]
+            cluster = settings["control_plane"]["cluster"]
+            exposer = settings["default_exposer"](cluster)
+            return exposer.prometheus_url(prometheus["project"], prometheus["service"])
```
averevki: Exposers are not source of truth for the cluster type testsuite running with, and they can be overwritten for the test after the settings validation.

testsuite/tests/conftest.py
crstrn13: I guess those can also be None and that way you skip the observability tests. Didn't add validator but treat the case if they are None.

testsuite/gateway/exposers.py
crstrn13: ok, I can understand why. Though, I'd like to be able to tell somehow on what distribution you are.

testsuite/config/tools.py
crstrn13: right, I moved everything in the validation, then. Though, I liked the abstraction better. Anyways I guess we can play around with that a little more.

testsuite/config/__init__.py
```diff
@@ -79,7 +79,11 @@ def __init__(self, name, default, **kwargs) -> None:
             "custom_metrics_apiserver.url",
             default=fetch_service_ip("custom-metrics-apiserver", protocol="http", port=8080),
         ),
+        DefaultValueValidator(
+            "prometheus.url",
+            default=fetch_prometheus_url(),
+        ),
     ],
-    validate_only=["authorino", "kuadrant", "default_exposer", "control_plane"],
+    validate_only=["authorino", "kuadrant", "default_exposer", "control_plane", "prometheus"],
```
silvi-t: The other optional services (keycloak, mockserver, tracing) are not in `validate_only`. Since the prometheus fixture in `testsuite/tests/conftest.py` already does this, maybe remove "prometheus" from `validate_only` to keep it consistent with the rest?

testsuite/tests/conftest.py
```diff
@@ -226,29 +224,20 @@ def testconfig():
...
-            assert yaml.safe_load(cm["config.yaml"])["enableUserWorkload"]
-    except Exception:  # pylint: disable=broad-exception-caught
-        pytest.skip("User workload monitoring is disabled")
-
-    # find thanos-querier route in the openshift-monitoring project
-    # this route allows to query metrics
+        testconfig.validators.validate(only=["prometheus"])
+        url = testconfig["prometheus"]["url"]
+    except (KeyError, ValidationError):
+        skip_or_fail("Prometheus not available. Set prometheus.url in config or ensure monitoring is deployed.")
```
silvi-t: When running on Kind without `prometheus.project` and `prometheus.service` set in `settings.local.yaml`, the skip message suggests setting `prometheus.url`, but the real fix is setting `prometheus.project: "monitoring"` and `prometheus.service: "prometheus-kube-prometheus-prometheus"` so auto-discovery works. The message could mention that, e.g. "Set prometheus.project and prometheus.service for your environment, or override with prometheus.url in config."

testsuite/config/__init__.py
crstrn13: done.

testsuite/tests/conftest.py
crstrn13: added.

silvi-t (dismissed):
LGTM

testsuite/config/tools.py
```diff
@@ -69,6 +71,60 @@ def _fetcher(settings, _):
...
+            with cluster.context:
+                svc = selector(f"service/{service_name}").object(cls=Service)
+                ip = svc.external_ip
+                if ip:
+                    port = svc.model.spec.ports[0].port
+                    return f"http://{ip}:{port}"
+        except OpenShiftPythonException:
+            logger.info("Service '%s' not found in '%s', trying OpenShift route", service_name, project)
+        except (AttributeError, IndexError) as exc:
+            logger.warning("Service '%s' in '%s' has no external IP or ports: %s", service_name, project, exc)
```
averevki, line 120: I think it still should prioritize the openshift environment first here

testsuite/tests/singlecluster/authorino/metrics/test_metrics.py
```diff
@@ -39,41 +40,49 @@
...
+@pytest.fixture(scope="module")
+def metrics_keys(prometheus, metrics_labels):
+    """Return a function that checks if a metric exists, retrying for propagation delays"""
+
+    @backoff.on_predicate(backoff.constant, interval=10, jitter=None, max_tries=6)
+    def _has_metric(metric):
+        return metric in prometheus.get_metrics(labels=metrics_labels).names
 
-    return prometheus.get_metrics(labels={"service": authorino.metrics_service.name()}).names
+    return _has_metric
```
averevki: This is changing how metrics are acquired for the test. Instead of sending a single request for metrics only once for the whole module, it now queries prometheus for each test. This test otherwise is stable on ocp, wait_for_scrape is something that insures that metrics appear. I don't like this tradeoff in favor of this test runnable on Kind envionments

testsuite/tests/singlecluster/authorino/metrics/test_metrics.py
```diff
@@ -39,41 +40,49 @@
...
 
 
 @pytest.mark.parametrize("metric", METRICS)
 def test_metrics(metric, metrics_keys):
     """Test for metrics that Authorino export at the /metrics endpoint"""
-    assert metric in metrics_keys
+    assert metrics_keys(metric), f"Metric '{metric}' not found"
```
averevki: arguably, a less verbose assert than `metric in metrics_keys`. Although new assert message rescues it a bit

testsuite/gateway/exposers.py
averevki: Yes we need some reliable way to determine the cluster distribution, just not the exposer class

testsuite/config/tools.py
crstrn13: switched order.

testsuite/tests/singlecluster/authorino/metrics/test_metrics.py
crstrn13: reverted changes.

testsuite/tests/singlecluster/authorino/metrics/test_metrics.py
crstrn13: reverted changes.

testsuite/tests/singlecluster/authorino/metrics/test_metrics.py
```diff
@@ -44,12 +44,12 @@ def metrics_keys(authorino, service_monitor, prometheus, client, auth):
...
     """
-    response = client.get("/get", auth=auth)
-    assert response.status_code == 200
-
     prometheus.wait_for_scrape(service_monitor, "/metrics")
     prometheus.wait_for_scrape(service_monitor, "/server-metrics")
 
+    response = client.get("/get", auth=auth)
+    assert response.status_code == 200
+
```
crstrn13, line 52: reverted order, as there might be a small window where the scraping is not ready and the requests goes in. This way, the request will only go if the metric is scraped.

averevki: FYI this single request here is for some metrics to appear if they aren't by default. I'm not really sure if it is still needed for this test, but the order plays this role in deep_metrics tests, as far as I remember. Should be good if it still works after the change

averevki (approved):

PR #923, opened 2026-04-07
test(tracing): add control plane sources attributes tests

crstrn13:
## Description

Adds test coverage for control plane tracing `sources` attributes (issue #919). These attributes link policy reconciliation spans to the policies that triggered them. Also fixes timing-related flakiness in existing lifecycle tracing tests when running in fast environments like kind.

Closes #919

### New Tests (`test_control_plane_sources.py`)
- **`test_authconfig_span_attributes`**: Validates authconfig spans have `sources`, `name`, and `namespace` attributes with correct AuthPolicy references
- **`test_limitador_span_attributes`**: Validates limitador spans have `sources`, `name`, and `namespace` attributes with correct RateLimitPolicy references
- **`test_authconfig_span_is_child_of_reconciler`**: Validates authconfig spans are children of `reconciler.auth_configs` spans
- **`test_authconfig_sources_contains_multiple_policies`**: Creates a second AuthPolicy targeting the same route and verifies policy references appear in sources
- **`test_limitador_sources_contains_multiple_policies`**: Creates a second RateLimitPolicy targeting the same route and verifies policy references appear in sources

### Tracing Client Improvements
- Added `start_time` parameter to `TracingClient.get_traces()`, `JaegerClient.get_traces()`, and `RemoteTempoClient.get_traces()` for timestamp-bounded trace queries (microseconds)

### Lifecycle Test Fixes (`test_control_plane_lifecycle.py`)
- Fixed race conditions in `test_policy_update_generates_new_reconciliation_trace` and `test_policy_target_change_traced` by capturing precise timestamps before mutations and using `start_time` to scope trace queries, instead of relying on snapshot comparison alone

### Documentation
- Added Jaeger tracing access instructions to `CLAUDE.md`

## Verification

```bash
poetry run pytest -vv -n4 testsuite/tests/singlecluster/tracing/control_plane/
```


CLAUDE.md
```diff
@@ -72,6 +72,45 @@ make mypy                # Run type checking
 make clean               # Delete all test-created resources (uses USER env var)
 ```
 
+### Accessing Jaeger Tracing
```
silvi-t: The "Accessing Jaeger Tracing" section added to CLAUDE.md duplicates the content already in `.claude/rules/tracing-setup.md`, which is auto-loaded into Claude Code's context. Let's remove this section from CLAUDE.md to avoid duplication and keep `.claude/rules/tracing-setup.md` as the single source of truth.

testsuite/tests/singlecluster/tracing/control_plane/test_control_plane_sources.py
```diff
@@ -0,0 +1,253 @@
...
+
+from testsuite.kuadrant.policy.authorization.auth_policy import AuthPolicy
+from testsuite.kuadrant.policy.rate_limit import Limit, RateLimitPolicy
+from testsuite.kubernetes import Selector
+
+pytestmark = [pytest.mark.observability, pytest.mark.limitador, pytest.mark.authorino, pytest.mark.kuadrant_only]
+
+
+@pytest.fixture(scope="module")
+def authconfig_trace(auth_traces, skip_or_fail):
```
silvi-t: The `authconfig_trace` and `limitador_trace` fixtures use `skip_or_fail` when no matching trace is found, but tracing is already confirmed enabled by `require_tracing_enabled` and the policies are committed. If no trace with `sources` appears, that's a real failure. A plain `assert False, "No trace with authconfig span containing 'sources' attribute found"` would be sufficient here.

testsuite/tests/singlecluster/tracing/control_plane/test_control_plane_sources.py
```diff
@@ -0,0 +1,253 @@
...
+    for trace in auth_traces:
+        spans = trace.filter_spans(lambda s: s.operation_name == "authconfig" and s.has_tag("sources"))
+        if spans:
+            return trace
+
+    return skip_or_fail("No trace with authconfig span containing 'sources' attribute found")
+
+
+@pytest.fixture(scope="module")
+def limitador_trace(rl_traces, skip_or_fail):
```
silvi-t: Same as for `authconfig_trace` fixture.

testsuite/tests/singlecluster/tracing/control_plane/test_control_plane_sources.py
```diff
@@ -0,0 +1,253 @@
...
+    for trace in traces:
+        spans = trace.filter_spans(lambda s: s.operation_name == "authconfig" and s.has_tag("sources"))
+        for span in spans:
+            sources = span.get_tag("sources")
+            if sources and (first_policy_ref in sources or second_policy_ref in sources):
+                return trace
+
+    return skip_or_fail(
+        f"No trace with authconfig span found with either policy. "
+        f"Looking for {first_policy_ref} or {second_policy_ref} in sources"
```
silvi-t: Again I would leave this as plain assert since missing that trace is real issue.

testsuite/tests/singlecluster/tracing/control_plane/test_control_plane_sources.py
```diff
@@ -0,0 +1,253 @@
...
+    )
+
+
+def test_authconfig_sources_contains_multiple_policies(
```
silvi-t: The `authconfig_trace_multiple_policies` fixture queries traces with `start_time=create_time` (after the second policy was created), so the first policy's reconciliation traces won't be in the result set. This means
  `first_policy_ref` can never be found — the `or` check is effectively dead code.

The same applies to `limitador_trace_multiple_policies` and both corresponding tests.

  Either:
  1. Simplify to only check for `second_policy_ref` (since that's the only one that can appear in the time-scoped traces), or
  2. Remove the `start_time` filter to capture traces from both policies — but that risks picking up unrelated older traces.

silvi-t (approved):
LGTM

PR #927, opened 2026-04-08
test(egress): add tests for egress credentials injection 

averevki:
## Summary

  Add E2E tests for egress credential injection via AuthPolicy, based on the investigation from
  https://github.com/Kuadrant/architecture/issues/148.

  - **Credential injection pattern:** AuthPolicy uses `metadata.http` to fetch a K8s Secret via the Kubernetes API
   (authenticated with Authorino's SA token), then injects it as a request header via `responses.success.headers`
  - **Destination-based selection:** Two HTTPRoutes (`/service1`, `/service2`) with separate AuthPolicies inject
  different credentials per destination

  ## Changes

  - Added `replace_prefix_match` to `URLRewriteFilter` for path rewriting (`testsuite/gateway/__init__.py`)
  - Added `credentials` and `shared_secret_ref` parameters to `MetadataSection.add_http()`
  (`testsuite/kuadrant/policy/authorization/sections.py`)
  - Moved `test_egress.py` into `egress/` directory, extracted shared egress fixtures to `egress/conftest.py`
  - `credentials_injection/conftest.py` — MockServer backend, SA token secret, service API key secrets,
  credential-injecting AuthPolicy

## New tests 
  - `test_credential_injection.py` — single-route credential injection test
  - `test_credential_injection_by_destination.py` — destination-based credential injection test
  
 ## Prerequisites
 Add cluster CA to authorino pod trust store - https://github.com/Kuadrant/helm-charts-olm/pull/79

  ## Verification

  ```bash
  make testsuite/tests/singlecluster/egress/
```
  
  Closes #913 

## Summary by CodeRabbit

* **New Features**
  * Enhanced URL rewriting capabilities with prefix replacement support for gateway routes.
  * Added credential injection functionality for egress gateway traffic.
  * Extended authorization and rate limiting policy enforcement to egress gateway scenarios.

* **Tests**
  * Expanded test coverage for egress gateway credential injection and policy enforcement.

* **Chores**
  * Updated test execution paths for egress gateway test suite.

### CODE REVIEW

File: testsuite/tests/singlecluster/egress/test_egress.py [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+LIMIT = Limit(3, "5s")
+
+
+@pytest.fixture(scope="module")
+def hostname(request, backend, blame, cluster) -> Hostname:
+    """Expose the backend service directly via an OpenShift Route"""
+    route = OpenshiftRoute.create_instance(cluster, blame("backend"), backend.name, "http", tls=True)
+    request.addfinalizer(route.delete)
+    route.commit()
+    return route
```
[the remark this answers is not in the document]
averevki: That's not valid, OpenshiftRoute doesn't have a wait_for_ready() method, and it isn't needed there

testsuite/tests/singlecluster/egress/test_egress.py
```diff
@@ -0,0 +1,85 @@
...
+    return rate_limit
+
+
+@pytest.fixture(scope="module")
+def auth(oidc_provider):
+    """Returns OIDC authentication object for HTTPX"""
+    return HttpxOidcClientAuth(oidc_provider.get_token, "authorization")
+
+
+def test_egress_authorization(client, auth):
```
zkraus, line 60: Should we introduce a new test category / mark `egress` ? or this would be too granular at the moment?

averevki: There will be only 4 egress tests for the expected future but I think we still could have it. Those are only the helping targets for now, not used by any automation. I've added the `egress-gateway` target, thank you

averevki:
rebased on main^ after egress gateway base test got merged

testsuite/tests/singlecluster/egress/credentials_injection/test_credential_injection.py
```diff
@@ -0,0 +1,50 @@
...
+    """Commit the AuthPolicy and wait for it to be enforced"""
+    request.addfinalizer(authorization.delete)
+    authorization.commit()
+    authorization.wait_for_ready()
+
+
+def test_egress_credential_injection(client, mockserver_expectation):
+    """Test that the correct credential is accepted (200) and that no injected credential results in rejection (401)"""
+    response = client.get(mockserver_expectation)
+    assert response.status_code == 200
```
maksymvavilov, line 47: If we don't trust the cluster CA, this will fail, right? 
Is it worth having a "preflight check" to make sure we do trust it (check for the ConfigMap created by helm-charts-olm CA trust hook or something like that, and skip the test if we don't)? 
Or the first thing that you do when it fails is look into the logs, and it will be in plain text there?

maksymvavilov (dismissed):
It tests what should be tested. 
This is built on how I proposed to deal with injection, but turns out the MCP Gateway already invented this wheel. The same mechanism, but using Vault instead of k8s secret. But it should be fine since the mechanism is identical. 
Just one question about the test (it's a question rather than a suggestion).

testsuite/tests/singlecluster/egress/credentials_injection/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+@pytest.fixture(scope="module")
+def mockserver_expectation(mockserver_client, module_label):
+    """MockServer expectation: 200 if Authorization: Bearer <service1-key>, 401 otherwise"""
+    template = """
+    {
+      "statusCode": #if($request.headers['authorization']
+        && $request.headers['authorization'].contains("Bearer %s")) 200 #else 401 #end,
+      "headers": { "Content-Type": ["application/json"] }
+    }
+    """ % SERVICE1_API_KEY
```
azgabur, line 52: Can you use f-string?

averevki: No because this template contains symbols that break the f-string

testsuite/tests/singlecluster/egress/credentials_injection/test_credential_injection.py
averevki: No logs will just say that it failed on 401 status code, because metadata collection from the cluster would fail and mockserver backend would return 401 in case no credentials are passed. That's actually a good point, I've added following check for the volume to the fixture:

https://github.com/Kuadrant/testsuite/blob/a1f607736715e6b6b543a0ddd1b18e211f39fe89/testsuite/tests/singlecluster/egress/credentials_injection/conftest.py#L19-L24

azgabur (approved):
LGTM

PR #934, opened 2026-04-14
fix(dnscli): change secret-generation command name to add-cluster-secret

averevki:
## Description

CoreDNS secret-generation command name was changed to `add-cluster-secret`

## Tests

* Updated test invocations to reflect the command name change from `secret-generation` to `add-cluster-secret` in cluster secret generation operations.


silvi-t (approved):
Since the change is small, I only did an eye review, but LGTM.

PR #948, opened 2026-04-22
feat(testsuite): introduce pytest-rerunfailures plugin

silvi-t:
> [!IMPORTANT]

  ## Description
  - Introduce `pytest-rerunfailures` plugin to automatically retry failed tests (default `--reruns 3`), reducing flaky test noise in CI
  - Annotate 60+ tests with `@pytest.mark.flaky` to either disable reruns (`reruns=0`) for tests that modify module-scoped fixtures, or add appropriate `reruns_delay` for rate limit tests
  - Record rerun metadata (attempt count, failure messages, captured output) as JUnit XML user properties for Report Portal integration
  - Make `KubernetesObject.commit()` resilient to reruns by falling back to `apply()` when a resource already exists

## Per-test rerun markers

  `pytest-rerunfailures` only reruns the **test function**, not module-scoped fixtures. Some tests need special handling:                                                                                                          
   
  - **`@pytest.mark.flaky(reruns=0)`** — disables reruns for tests that delete or modify module-scoped resources (e.g., delete a policy, change a targetRef, rollout a deployment). The fixture is not recreated between reruns, so
   the test cannot recover.
                                                                                                                                                                                                                                   
  - **`@pytest.mark.flaky(reruns=3, reruns_delay=X)`** — adds a delay between reruns for rate limit tests. When a test exhausts its counter and fails, the counter is still active in Limitador. A rerun that starts immediately   
  would hit `429` on the very first request. The delay is set to `window + 5s` so the counter fully resets before the next attempt.
  ## Changes

  ### New Plugin & Configuration
  - Added `pytest-rerunfailures` dependency in `pyproject.toml`
  - Added `--reruns 3` default to the `PYTEST` command in `Makefile`

  ### Core Infrastructure
  - Modified `KubernetesObject.commit()` in `testsuite/kubernetes/__init__.py` to catch `AlreadyExists` and fall back to `apply()`, so module-scoped resources survive reruns
  - Added `_format_failure_message()` helper in `testsuite/tests/conftest.py` that produces clean tracebacks by filtering out `site-packages` frames and extracting stderr from `OpenShiftPythonException`
  - Added `_collect_rerun_attempt()` and `_write_rerun_properties()` in `testsuite/tests/conftest.py` to record per-attempt failure messages and isolated captured output as `__rp_rerun_*` user properties at teardown
  - Extended `pytest_runtest_makereport` hook to call the new rerun helpers on `"call"/"setup"` failures and at `"teardown"`

### Bug Fix                                                                                                                                                                                                                      
  - Fixed image digest stripping in `testsuite/component_metadata.py` — `container.image.split("@", 1)[0]` now correctly removes the `@sha256:...` suffix                                                                          
                                                                                                                                                                                                                                   
  ### Documentation
  - Updated `CLAUDE.md` with guidelines for `@pytest.mark.flaky` usage

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Default automatic test reruns enabled (3 attempts) and test reports now include per-attempt messages.

* **Documentation**
  * Added guidance for controlling reruns, recommended delays for rate-limit tests, and when to disable retries.

* **Tests**
  * Applied flaky/retry markers across the test suite with tailored rerun counts/delays; several tests explicitly disable retries where unsafe.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/.claude/rules/test-reruns.md
+++ b/.claude/rules/test-reruns.md
@@ -1,0 +1,34 @@
+# Test Reruns (`@pytest.mark.flaky`)
+
+The testsuite uses `pytest-rerunfailures` to automatically retry failed tests (default: `--reruns 3` in the Makefile). The `@pytest.mark.flaky` marker controls rerun behavior per test.
+
+**Disable reruns** for tests that cannot recover from a failure, because they delete or modify module-scoped fixtures during the test. Since `pytest-rerunfailures` only reruns the test function (not module-scoped fixtures), the deleted/modified resource will not be recreated:
+
+```python
+@pytest.mark.flaky(reruns=0)
+def test_policy_deletion(policy):
+    """This test deletes a module-scoped fixture, so rerunning would fail"""
+    policy.delete()
+    assert not policy.exists()
+```
+
+**Add a delay** for rate limit tests that exhaust a counter and assert `429`. On rerun, the counter from the previous attempt may still be active. Set `reruns_delay` to the rate limit window + 5 seconds:
+
+```python
+@pytest.mark.flaky(reruns=3, reruns_delay=15)  # Limit window is 10s, wait 15s
+def test_rate_limit(client):
+    responses = client.get_many("/get", 5)
+    responses.assert_all(status_code=200)
+    assert client.get("/get").status_code == 429
+```
+
+**When to use which:**
+
+| Situation | Marker |
+|-----------|--------|
+| Test deletes or modifies a module-scoped fixture | `@pytest.mark.flaky(reruns=0)` |
+| Test creates and deletes resources within the test body (e.g., UI tests) | `@pytest.mark.flaky(reruns=0)` |
+| Rate limit test with 5s window | `@pytest.mark.flaky(reruns=3, reruns_delay=10)` |
+| Rate limit test with 10s window | `@pytest.mark.flaky(reruns=3, reruns_delay=15)` |
+| Rate limit test with 60s window | `@pytest.mark.flaky(reruns=3, reruns_delay=65)` |
+| Normal test (no side effects) | No marker needed (uses global `--reruns 3`) |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,6 +171,10 @@ Tests use markers to categorize functionality. All available markers are defined
 - `@pytest.mark.issue("https://github.com/...")` - Linked to GitHub/Jira issues
 - `@pytest.mark.min_ocp_version((4, 20))` - Minimum OpenShift version requirement
 
+### Test Reruns (`@pytest.mark.flaky`)
+
+See `.claude/rules/test-reruns.md` for detailed guidance on using `@pytest.mark.flaky` to control rerun behavior per test.
+
 ## Coding Conventions
 
 ### Docstrings
--- a/testsuite/tests/singlecluster/egress/test_egress.py
+++ b/testsuite/tests/singlecluster/egress/test_egress.py
@@ -65,6 +65,7 @@ def test_egress_authorization(client, auth):
     assert response.status_code == 200
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=10)
 def test_egress_ratelimit(client, auth):
     """Test that RateLimitPolicy is enforced on egress gateway traffic"""
     sleep(5 + 1)  # make sure request limit quota is reset before starting the test
--- a/testsuite/tests/singlecluster/extensions/plan_policy/test_plan_policy.py
+++ b/testsuite/tests/singlecluster/extensions/plan_policy/test_plan_policy.py
@@ -82,11 +82,15 @@ def commit(request, plan_policy, authorization):
 @pytest.mark.parametrize(
     "user_with_plan, allowed_requests",
     [
-        (tier, plan.limits["custom"][0]["limit"] if "custom" in plan.limits else list(plan.limits.values())[0])
-        for tier, plan in PLANS.items()
+        pytest.param("gold", 5, id="gold"),
+        pytest.param("silver", 3, id="silver"),
+        pytest.param("bronze", 2, id="bronze", marks=pytest.mark.flaky(reruns=0)),
     ],
     indirect=["user_with_plan"],
 )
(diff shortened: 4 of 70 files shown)


CLAUDE.md
```diff
@@ -171,6 +171,41 @@ Tests use markers to categorize functionality. All available markers are defined
 - `@pytest.mark.issue("https://github.com/...")` - Linked to GitHub/Jira issues
 - `@pytest.mark.min_ocp_version((4, 20))` - Minimum OpenShift version requirement
 
+### Test Reruns (`@pytest.mark.flaky`)
+
+The testsuite uses `pytest-rerunfailures` to automatically retry failed tests (default: `--reruns 3` in the Makefile). The `@pytest.mark.flaky` marker controls rerun behavior per test.
+
+**Disable reruns** for tests that cannot recover from a failure, because they delete or modify module-scoped fixtures during the test. Since `pytest-rerunfailures` only reruns the test function (not module-scoped fixtures), the deleted/modified resource will not be recreated:
+
+```python
```
emmaaroche: This is helpful reference material but I'm concerned about the `CLAUDE.md` growing too large if we were to update it like this for every PR. The docs recommend keeping it under 200 lines and splitting content into rule files if growing large, as longer files consume more context and reduce adherence, see https://code.claude.com/docs/en/memory#write-effective-instructions.

Maybe we could move the "Test Reruns" section to a .claude/rules/test-reruns.md file, wdyt?

testsuite/tests/singlecluster/extensions/plan_policy/test_plan_policy.py
```diff
@@ -87,6 +87,7 @@ def commit(request, plan_policy, authorization):
     ],
     indirect=["user_with_plan"],
 )
+@pytest.mark.flaky(reruns=3, reruns_delay=25)
 def test_plan_policy(client, user_with_plan, allowed_requests):
     """
```
emmaaroche, line 95: This test uses `reruns_delay=25`, but the bronze plan has `limits={"daily": 2}` which is a 24-hour window. If the test exhausts its 2 request counter and fails, waiting 25 seconds won't be long enough for the counter to reset in this case.

Could we maybe disable reruns for the bronze parametrization specifically?

CLAUDE.md
silvi-t: Moved the "Test Reruns" section from CLAUDE.md to .claude/rules/test-reruns.md

testsuite/tests/singlecluster/extensions/plan_policy/test_plan_policy.py
silvi-t: Good catch! The bronze parametrization now has `marks=pytest.mark.flaky(reruns=0)` which overrides the test-level `reruns=3`, so gold/silver will retry with the 25s delay (sufficient for their 10s/20s windows) while bronze (daily limit) won't retry at all.

.claude/rules/test-reruns.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Disable reruns** for tests that cannot recover from a failure, because they delete or modify module-scoped fixtures during the test. Since `pytest-rerunfailures` only reruns the test function (not module-scoped fixtures), the deleted/modified resource will not be recreated:
+
+```python
+@pytest.mark.flaky(reruns=0)
+def test_policy_deletion(policy):
+    """This test deletes a module-scoped fixture, so rerunning would fail"""
+    policy.delete()
+    assert not policy.exists()
+```
```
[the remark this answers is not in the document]
silvi-t: This is only example for Claude code how to add this mark @pytest.mark.flaky`(reruns=0)

emmaaroche:
I was scanning through tests that meet the criteria for needing a flaky marker and noticed the test in `testsuite/tests/singlecluster/limitador/section/test_multiple_rules.py` doesn't have one, just checking if that is intentional?

silvi-t:
> I was scanning through tests that meet the criteria for needing a flaky marker and noticed the test in `testsuite/tests/singlecluster/limitador/section/test_multiple_rules.py` doesn't have one, just checking if that is intentional?

Good catch! I had Claude scan through the rate limit tests to add the flaky markers, and it missed this one. Added `@pytest.mark.flaky(reruns=3, reruns_delay=10)` for the 5s window.

testsuite/tests/singlecluster/egress/test_egress.py
```diff
@@ -65,6 +65,7 @@ def test_egress_authorization(client, auth):
     assert response.status_code == 200
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=10)
```
azgabur, line 68: `reruns=3` is already applied from the pytest arg in Makefile, what about only specifying the delay? This applies to all marks not just this one.

silvi-t: `reruns=3` is not redundant here. When a `@pytest.mark.flaky` marker is present, `pytest-rerunfailures` uses the marker as priority over the global `--reruns` setting. If `reruns` is not specified in the marker, it defaults to `1`, not to the global value from the Makefile. So `@pytest.mark.flaky(reruns_delay=10)` would only rerun the test once instead of three times.

emmaaroche (dismissed):
LGTM!

testsuite/tests/singlecluster/extensions/plan_policy/test_plan_policy.py
zkraus: Thank you for explaining the behavior. Consider adding it as a comment here, so it is documented in place.

zkraus (dismissed):
One optional recommendation. Otherwise LGTM.

testsuite/tests/singlecluster/extensions/plan_policy/test_plan_policy.py
silvi-t: I added a comment with an explanation.

emmaaroche (dismissed):

emmaaroche (approved):

zkraus (approved):

PR #970, opened 2026-05-06
fix: recursive depth on nested oc selector.

crstrn13:
## Description

Introduced bug with the local kind prometheus due to faulty indentation.

## Release Notes

* **New Features**
  * Improved Prometheus URL auto-discovery now supports route-based endpoint detection for better reliability in cluster monitoring configurations, with automatic fallback to existing discovery mechanisms when routes are unavailable.


silvi-t (approved):
LGTM

PR #971, opened 2026-05-07
fix(metrics): scrape server-metrics after sending a request.

crstrn13:
## Description

Was scraping the server metrics before sending the request. This was introduced by commit trying to address that sending the request before the scrape will result in the same error on kind as it is way too fast for the scraping to be effective. Now we assert for the metrics that do not require request first and scrape only after the first request.

## Summary by CodeRabbit

* **Tests**
  * Updated metrics validation tests to ensure proper sequencing of metrics data collection, improving test reliability for authenticated traffic scenarios.


silvi-t (approved):
LGTM

PR #972, opened 2026-05-07
feat: assert denied requests are not received by upstream on data plane tests

### LINKED ISSUE #947: Assert denied requests are not received by the upstream on every data plane test
Parent: #945

Verify that requests denied by authorization (401/403) or rate limiting (429) do not reach the upstream service. This check should run automatically on every data plane test.

crstrn13:
> [!IMPORTANT]

Closes https://github.com/Kuadrant/testsuite/issues/947

## Description

Adds automatic detection of upstream request leaks — when the gateway denies a request (401/403/429), the testsuite now verifies that the request never reached the backend. This catches bugs where denied requests bypass the gateway and leak to the upstream service.

### Upstream Leak Detection
- `KuadrantClient` injects an `X-Testsuite-Tracking` header (UUID) on every request and records tracking IDs for denied responses (401/403/429)
- A `pytest_runtest_makereport` hook in the singlecluster conftest queries MockServer after each passing test to verify none of the denied requests leaked to the upstream
- `Mockserver.retrieve_requests_by_header()` supports querying by header value for leak checks

### Exposable Abstraction
- Extracted `Exposable` base class from `Gateway` (`testsuite/gateway/__init__.py`) so that both `Gateway` and `Backend` can be exposed by an `Exposer`
- `Backend` now implements `Exposable` with an `expose()` method that creates an admin service and hostname for external access
- `LoadBalancerServiceExposer` gains a `backend_service_type` property returning `"LoadBalancer"` so backends get externally accessible services on Kind clusters
- All exposers (`OpenShiftExposer`, `LoadBalancerServiceExposer`, `DNSPolicyExposer`) updated to accept `Exposable` instead of `Gateway`

### Backend Changes
- `MockserverBackend` no longer accepts a `service_type` parameter — service type is now determined by the exposer
- The session-scoped `backend` fixture in `singlecluster/conftest.py` now exposes the backend via the default exposer for admin access

## Verification

```bash
make kuadrant
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added automatic request tracking header injection in HTTP client.
  * Added denial-leak verification to detect if denied requests reach upstream backends.
  * Added method to query mock server by request header.
  * Improved backend exposure with flexible service strategies.

* **Refactor**
  * Decoupled exposure APIs from gateway-specific implementation.

* **Chores**
  * Added configuration flag for denial verification control.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/testsuite/pull/972?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 TB ?= short
 LOGLEVEL ?= INFO
-VERIFY_DENIALS ?= true # Set to false to skip verification of expected denials (e.g., 403 responses) in tests marked with `verify_denials` (e.g., `authorino` tests). This can be useful when debugging test failures that may be related to unexpected denials, but should generally be left enabled to ensure proper test validation.
+VERIFY_DENIALS ?= true # Set to false to skip the --verify-denials check that ensures denied requests (401/403/429) did not leak to the upstream backend.
 
 ifdef WORKSPACE  # Yes, this is for jenkins
 resultsdir = $(WORKSPACE)
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,9 @@ disconnected: poetry-no-dev  ## Run tests compatible with disconnected clusters
 egress-gateway: poetry-no-dev  ## Run egress gateway tests
 	$(PYTEST) -n4 -m 'egress_gateway' --dist loadfile --enforce $(flags) testsuite/tests/singlecluster/egress/
 
+data-plane data_plane: poetry-no-dev  ## Run data plane leak detection tests
+	$(PYTEST) -n4 -m 'data_plane' --dist loadfile $(flags) testsuite/tests/singlecluster
+
 kuadrantctl: poetry-no-dev  ## Run Kuadrantctl tests
 	$(PYTEST) -n4 --dist loadfile --enforce $(flags) testsuite/tests/kuadrantctl/
 
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ disconnected: poetry-no-dev  ## Run tests compatible with disconnected clusters
 egress-gateway: poetry-no-dev  ## Run egress gateway tests
 	$(PYTEST) -n4 -m 'egress_gateway' --dist loadfile --enforce $(flags) testsuite/tests/singlecluster/egress/
 
-data-plane data_plane: poetry-no-dev  ## Run data plane leak detection tests
+data-plane: poetry-no-dev  ## Run data plane leak detection tests
 	$(PYTEST) -n4 -m 'data_plane' --dist loadfile $(flags) testsuite/tests/singlecluster
 
 kuadrantctl: poetry-no-dev  ## Run Kuadrantctl tests
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,6 @@ disconnected: poetry-no-dev  ## Run tests compatible with disconnected clusters
 egress-gateway: poetry-no-dev  ## Run egress gateway tests
 	$(PYTEST) -n4 -m 'egress_gateway' --dist loadfile --enforce $(flags) testsuite/tests/singlecluster/egress/
 
-
 kuadrantctl: poetry-no-dev  ## Run Kuadrantctl tests
 	$(PYTEST) -n4 --dist loadfile --enforce $(flags) testsuite/tests/kuadrantctl/
 
--- a/Makefile
+++ b/Makefile
@@ -75,8 +75,6 @@ disconnected: poetry-no-dev  ## Run tests compatible with disconnected clusters
 egress-gateway: poetry-no-dev  ## Run egress gateway tests
 	$(PYTEST) -n4 -m 'egress_gateway' --dist loadfile --enforce $(flags) testsuite/tests/singlecluster/egress/
 
-data-plane: poetry-no-dev  ## Run data plane leak detection tests
-	$(PYTEST) -n4 -m 'data_plane' --dist loadfile $(flags) testsuite/tests/singlecluster
 
 kuadrantctl: poetry-no-dev  ## Run Kuadrantctl tests
 	$(PYTEST) -n4 --dist loadfile --enforce $(flags) testsuite/tests/kuadrantctl/
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 TB ?= short
 LOGLEVEL ?= INFO
+VERIFY_DENIALS ?= true # Set to false to skip verification of expected denials (e.g., 403 responses) in tests marked with `verify_denials` (e.g., `authorino` tests). This can be useful when debugging test failures that may be related to unexpected denials, but should generally be left enabled to ensure proper test validation.
 
 ifdef WORKSPACE  # Yes, this is for jenkins
 resultsdir = $(WORKSPACE)
@@ -15,6 +16,10 @@ RUNSCRIPT = poetry run ./scripts/
 
 ifdef junit
 PYTEST += --junitxml=$(resultsdir)/junit-$(@F).xml -o junit_suite_name=$(@F)
+endif
+
+ifdef VERIFY_DENIALS # Only add the flag if VERIFY_DENIALS is set to a non-empty value (default is "true")
+PYTEST += --verify-denials=$(VERIFY_DENIALS)
 endif
 
 # Collector PYTEST Override
--- a/testsuite/backend/__init__.py
+++ b/testsuite/backend/__init__.py
@@ -58,7 +58,7 @@ class Backend(LifecycleObject, Referencable, Exposable):
 
     def expose(self, exposer: "Exposer", name: str):
         """Creates external access for admin APIs via the exposer"""
-        self._admin_hostname = exposer.expose_backend(name, self)
+        self._admin_hostname = exposer.expose_hostname(name, self)
 
     @property
     def url(self):
--- a/testsuite/backend/__init__.py
+++ b/testsuite/backend/__init__.py
@@ -43,7 +43,13 @@ class Backend(LifecycleObject, Referencable, Exposable):
 
     @property
     def service_name(self) -> str:
+        """Returns the Kubernetes Service name for this backend"""
         return self.name
+
+    @property
+    def port_name(self) -> str:
+        """Service port name used when creating external routes"""
+        return "http"
 
     @property
     def match_labels(self) -> dict[str, str]:
--- a/testsuite/backend/__init__.py
+++ b/testsuite/backend/__init__.py
@@ -2,23 +2,34 @@
 
 from abc import abstractmethod
 from functools import cached_property
+from typing import TYPE_CHECKING, Optional
 
-from testsuite.gateway import Referencable
+from testsuite.gateway import Referencable, Exposable
 from testsuite.lifecycle import LifecycleObject
 from testsuite.kubernetes.client import KubernetesClient
 from testsuite.utils.constants import HTTP_API_PORT
 
+if TYPE_CHECKING:
+    from testsuite.gateway import Hostname, Exposer
 
-class Backend(LifecycleObject, Referencable):
+
+class Backend(LifecycleObject, Referencable, Exposable):
     """Backend (workload) deployed in Kubernetes"""
 
     def __init__(self, cluster: KubernetesClient, name: str, label: str):
-        self.cluster = cluster
+        self._cluster = cluster
         self.name = name
         self.label = label
 
         self.deployment = None
         self.service = None
+        self._admin_hostname: Optional["Hostname"] = None
+        self._admin_service = None
+
+    @property
+    def cluster(self) -> KubernetesClient:
+        """Returns KubernetesClient"""
+        return self._cluster
 
     @property
     def reference(self):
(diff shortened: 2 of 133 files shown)


testsuite/tests/singlecluster/conftest.py
```diff
@@ -99,14 +103,22 @@ def mockserver_config(cluster, blame, label):
 
 
 @pytest.fixture(scope="session")
-def backend(request, cluster, blame, label, mockserver_config):
+def backend_exposer(request, testconfig, cluster):
+    """Dedicated session-scoped exposer for the backend, avoids ScopeMismatch with module-scoped exposer overrides"""
+    exp = testconfig["default_exposer"](cluster)
```
emmaaroche, line 109: Missing testconfig validation. Add `testconfig.validators.validate(only="default_exposer")` before accessing `testconfig["default_exposer"]` so you get clear error if config missing

testsuite/tests/singlecluster/conftest.py
```diff
@@ -166,6 +178,51 @@ def client(route, hostname):  # pylint: disable=unused-argument
...
+    tracking_header = KuadrantClient.TRACKING_HEADER
+
+    leaked = []
+    try:
+        for denied_id in denied_ids:
+            if ms.retrieve_requests_by_header(tracking_header, denied_id):
+                leaked.append(denied_id)
+    except (httpx.RequestError, httpx.HTTPStatusError):
+        logging.warning("Failed to check for upstream leaks via MockServer", exc_info=True)
+        return
```
emmaaroche, line 226: This hook silently logs MockServer errors and lets test pass. What happens if MockServer admin is unreachable - should test fail or at least be skipped (leak check couldn't run, can't guarantee no leak) or pass (don't break tests when admin unavailable)? Current choice might hide real leaks if MockServer down.

testsuite/tests/singlecluster/authorino/conftest.py
```diff
@@ -35,10 +35,10 @@ def authorino(kuadrant, cluster, blame, request, testconfig, label):
 
 
 @pytest.fixture(scope="module")
-def authorization(authorization, oidc_provider, route, authorization_name, cluster, label) -> AuthConfig:
+def authorization(authorization, oidc_provider, gateway, authorization_name, cluster, label) -> AuthConfig:
     """In case of Authorino, AuthConfig used for authorization"""
     if authorization is None:
-        authorization = AuthConfig.create_instance(cluster, authorization_name, route, labels={"testRun": label})
+        authorization = AuthConfig.create_instance(cluster, authorization_name, gateway, labels={"testRun": label})
```
silvi-t: The change from `route` to `gateway` breaks all standalone Authorino tests.
 
`AuthConfig.create_instance` calls `target.add_auth_config()` on its third argument (`testsuite/kuadrant/policy/authorization/auth_config.py:57`), but `add_auth_config` is only defined on `EnvoyVirtualRoute` (`testsuite/gateway/envoy/route.py:61`), not on `Envoy/TLSEnvoy`. In standalone mode the gateway is an Envoy instance, so this raises `AttributeError: 'Envoy' object has no attribute 'add_auth_config'.`
This needs to be reverted back to `route`

testsuite/tests/singlecluster/conftest.py
crstrn13: Good point, I guess the mockserver should never be down as it is the sole backend in this case. I wasn't too sure myself either which would be the best course of action here. I guess I wanted to see it work and to not fail as this is more of a supplementary check. Thinking of wrapping it under a feature flag so to do the extra check only when instructed to. I plan on adding the feature flag, but not sure if I should the fail the test instead of warning. What do you reckon?

testsuite/tests/singlecluster/authorino/conftest.py
crstrn13: oh my, thanks a lot for finding this one. It slipped me completely

testsuite/tests/singlecluster/conftest.py
emmaaroche: The feature flag approach sounds good 👍 Thinking about it more, I'd go with warning instead of failing the test as that way you'd still see when the check didn't run without blocking tests that are otherwise passing.

emmaaroche: If we're going with warning, maybe it could be enabled by default instead of when instructed to? Since it won't fail tests, just flags potential issues in output.

crstrn13: yes, I think this is what I had in mind as well. It will just warn, the flag will turn on/off the hook at the end.

testsuite/backend/__init__.py
```diff
@@ -2,23 +2,35 @@
...
-        self.cluster = cluster
+        self._cluster = cluster
+
         self.name = name
         self.label = label
 
         self.deployment = None
         self.service = None
+        self._admin_hostname: Optional["Hostname"] = None
+        self._admin_service = None
```
silvi-t: `_admin_service` is initialized here and cleaned up in `delete()` but never assigned to an actual value anywhere — it's always `None`. If not needed, consider removing it and the cleanup code in `delete()`.

testsuite/gateway/__init__.py
```diff
@@ -276,12 +293,17 @@ def __init__(self, cluster):
...
+    def expose_hostname(self, name, exposable: Exposable) -> Hostname:
         """
         Exposes hostname, so it is accessible from outside
         Actual hostname is generated from "name" and is returned in a form of a Hostname object
         """
 
+    @property
+    def backend_service_type(self) -> Optional[Literal["ClusterIP", "LoadBalancer", "NodePort", "ExternalName"]]:
+        """Service type to create when exposing a backend directly. None means no extra service needed."""
+        return None
```
silvi-t: This property is defined here and in the `Exposer` base class but never read anywhere in the codebase. Is this intended for future use, or was it meant to be wired into ?

testsuite/tests/multicluster/conftest.py
```diff
@@ -82,7 +82,7 @@ def backends(request, cluster, cluster2, blame, label):
             data={"echo_expectation.json": echo_json},
         )
         config.commit()
-        mockserver = MockserverBackend(cluster_client, name, label, service_type="ClusterIP", config=config)
+        mockserver = MockserverBackend(cluster_client, name, label, config=config)
```
silvi-t: Removing `service_type="ClusterIP"` here silently changes the backend service from ClusterIP to LoadBalancer (the default in `MockserverBackend.__init__`). Was this intentional? The PR description says the `service_type` parameter was removed, but the constructor still accepts it.

testsuite/tests/singlecluster/gateway/dnspolicy/health_check/test_additional_headers.py
```diff
@@ -40,12 +39,13 @@ def gateway(request, cluster, blame, base_domain, module_label, subdomain):
...
 
 @pytest.fixture(scope="module")
-def backend(request, cluster, blame, label):
+def backend(request, cluster, blame, label, exposer):
     """Use mockserver as backend for health check requests to verify additional headers"""
     mockserver = MockserverBackend(cluster, blame("mocksrv"), label)
     request.addfinalizer(mockserver.delete)
     mockserver.commit()
     mockserver.wait_for_ready()
+    mockserver.expose(exposer, blame("mocksrv-admin"))
```
silvi-t: This fails on OCP with `nodename nor servname provided, or not known`. The `exposer` here is a `DNSPolicyExposer`, which just generates a hostname and expects a DNSPolicy to create the DNS record — but there's no DNSPolicy for the backend, so the admin hostname never resolves.

The old code worked because it accessed MockServer directly via the LoadBalancer service IP: `Mockserver(KuadrantClient(base_url=f"http://{backend.service.refresh().external_ip}:8080"))`

The `Exposable` abstraction works for `OpenShiftExposer` (creates a Route) and `LoadBalancerServiceExposer` (uses direct IP), but breaks in any test directory that overrides `exposer` to `DNSPolicyExposer`.

testsuite/tests/conftest.py
```diff
@@ -31,6 +31,9 @@ def pytest_addoption(parser):
         "--enforce", action="store_true", default=False, help="Fails tests instead of skip, if capabilities are missing"
     )
     parser.addoption("--standalone", action="store_true", default=False, help="Runs testsuite in standalone mode")
+    parser.addoption(
+        "--verify-denials", default="false", help="Verifies that denied requests did not leak to the upstream backend"
```
silvi-t: Makefile defaults `VERIFY_DENIALS` to `true`, but the pytest option defaults to `"false"`. This means leak detection is on via `make` targets but off via `poetry run pytest`. Should these be aligned?

testsuite/backend/__init__.py
```diff
@@ -30,6 +41,24 @@ def reference(self):
             "namespace": self.cluster.project,
         }
 
+    @property
+    def service_name(self) -> str:
```
silvi-t, line 45: Missing docstring

testsuite/backend/__init__.py
```diff
@@ -30,6 +41,24 @@ def reference(self):
...
+        return self.name
+
+    @property
+    def match_labels(self) -> dict[str, str]:
+        """Pod selector labels used by this backend's deployment"""
+        return {"app": self.label, "deployment": self.name}
```
silvi-t, line 57: This was added but never used. Consider either using `self.match_labels` in subclasses' `commit()` methods, or removing the property until it's actually needed.

Makefile
```diff
@@ -2,6 +2,7 @@
 
 TB ?= short
 LOGLEVEL ?= INFO
+VERIFY_DENIALS ?= true # Set to false to skip verification of expected denials (e.g., 403 responses) in tests marked with `verify_denials` (e.g., `authorino` tests). This can be useful when debugging test failures that may be related to unexpected denials, but should generally be left enabled to ensure proper test validation.
```
silvi-t: The comment references "tests marked with verify_denials" but there's no such pytest marker — `--verify-denials` is a global CLI flag, not a per-test marker. Consider updating the comment to reflect that.

testsuite/gateway/__init__.py
```diff
@@ -142,21 +142,38 @@ class GRPCRouteMatch:
...
+        """Service name to route traffic to"""
+
+    def external_ip(self) -> str:
+        """Returns external IP and port for external access"""
+        raise NotImplementedError(f"{type(self).__name__} does not support external_ip")
+
+    @property
+    def port_name(self) -> str:
+        """Service port name used when creating external routes"""
+        return "api"
```
silvi-t, line 165: The test fails on OCP with persistent 503s because `Exposable.port_name` defaults to "api", but the OpenShift Route resolves `targetPort` against the Endpoints port name, which inherits from the Service port name — "http"
`(ServicePort(name="http", ...))` in MockserverBackend. `port_name` should return "http".

silvi-t (commented):
Was the migration of `testsuite/tests/singlecluster/egress/credentials_injection/conftest.py:34` to the new `expose()/admin_hostname` pattern missed? It still uses `backend.service.refresh().external_ip` directly.

testsuite/tests/conftest.py
crstrn13: good catch, should be fixed now.

testsuite/backend/__init__.py
crstrn13: added.

testsuite/backend/__init__.py
crstrn13: good suggestion :+1:

Makefile
crstrn13: good catch :+1:

testsuite/gateway/__init__.py
crstrn13: should be overwritten now.

testsuite/tests/singlecluster/conftest.py
```diff
@@ -166,6 +179,54 @@ def client(route, hostname):  # pylint: disable=unused-argument
...
+    finally:
+        backend_client.close()
+
+    if leaked:
+        report.outcome = "failed"
+        report.longrepr = (
+            f"{len(leaked)} denied request(s) leaked to the upstream backend. "
```
silvi-t, line 234: The `pytest_runtest_makereport` hook sets `report.outcome = "failed"` when a leak is detected, but this triggers `pytest-rerunfailures` to retry the test (default `--reruns 3`). Since a genuine leak will fail every time, this
  causes 3 unnecessary reruns — plus teardown and recreation of module-scoped fixtures each time.

  **Fix:** set `report.wasxfail` to prevent rerunfailures from retrying:

  ```python
  if leaked:
      report.outcome = "failed"
      report.wasxfail = "denied request leaked to upstream"
      report.longrepr = (
          f"{len(leaked)} denied request(s) leaked to the upstream backend. "
          f"The gateway returned a denial status (401/403/429) but the request still reached MockServer."
      )

  Verified locally by simulating a fake leak and running with --reruns 3 — without the fix, the test reruns 3 times with full fixture teardown/recreation before finally failing.
  ```

crstrn13: amazing, didn't take that into consideration :pray:

emmaaroche (dismissed):
No other comments to add from me. LGTM 👍

silvi-t (approved):
LGTM 👍 The leak detection mechanism works as expected. Appreciate the effort put into this PR.

emmaaroche (approved):

PR #973, opened 2026-05-11
feat(local-setup): use cloud-provider-kind for local-setup

### LINKED ISSUE #957: Replace MetalLB with cloud-provider-kind for local-setup LoadBalancer support
## Problem

The current `make local-setup` uses [MetalLB](https://metallb.org/) to provide LoadBalancer service support in the kind cluster. MetalLB works by advertising L2 addresses on the container network, which requires the host to have direct L2 connectivity to the kind container network.

**This breaks on macOS** (and Windows) because kind containers run inside a VM (Docker Desktop / Podman Machine). The host has no direct route to the container network IPs assigned by MetalLB, making LoadBalancer services unreachable from the host. This is a blocker for local development and test execution on macOS.

Additionally, the current setup has a **hardcoded IP pool** (`172.18.255.200-172.18.255.250`) in `make/dependencies.mk` which assumes the Docker default `kind` network CIDR, and there is **no podman support** — `kind create cluster` doesn't pass `KIND_EXPERIMENTAL_PROVIDER`, so podman users can't use the local setup at all.

## Proposed Solution

Replace MetalLB with [cloud-provider-kind](https://github.com/kubernetes-sigs/cloud-provider-kind), a purpose-built cloud provider for kind clusters that handles LoadBalancer services by creating tunnels (port-forwards) from the host to the container. This works transparently regardless of whether the host can route to the container network, enabling near-native container networking on **all platforms** (Linux, macOS, Windows).

### What cloud-provider-kind does

- Watches for `Service` resources of type `LoadBalancer`
- Creates tunnels from the host to the kind node, exposing the service on `localhost` (or the node IP)
- Populates the service's `.status.loadBalancer.ingress` field, just like a real cloud provider would
- Works with both Docker and Podman as container runtimes

### 1. Add podman support to kind cluster creation
- Add a `CONTAINER_ENGINE` variable (default: `docker`) to `make/vars.mk`
- Pass it to kind via `KIND_EXPERIMENTAL_PROVIDER=$(CONTAINER_ENGINE)` in `make/kind.mk`

### 2. Remove MetalLB
- Remove the `install-metallb` target from `make/dependencies.mk`
- Remove `METALLB_VERSION` and `METALLB_TIMEOUT` from `make/vars.mk`
- Remove `install-metallb` from the `local-setup` dependency chain in `make/local-setup.mk`

### 3. Add cloud-provider-kind
- Add a new make target (e.g., `install-cloud-provider`) that runs `cloud-provider-kind` as a background process
- `cloud-provider-kind` can be installed via `go install sigs.k8s.io/cloud-provider-kind@latest` or downloaded from [GitHub releases](https://github.com/kubernetes-sigs/cloud-provider-kind/releases)
- It needs to run alongside the kind cluster (watches the kubeconfig's current context), typically started as a background process: `cloud-provider-kind &`
- Pass `CONTAINER_ENGINE` through so cloud-provider-kind uses the correct runtime

### 4. Update local-setup flow
- Replace `install-metallb` with the new cloud-provider-kind target in `make/local-setup.mk`
- Ensure cloud-provider-kind is running before any services of type LoadBalancer are created (gateway, tools, etc.)

### 5. Verify testsuite compatibility
- The testsuite's auto-fetch helpers (`testsuite/config/tools.py`) discover service IPs from LoadBalancer `.status.loadBalancer.ingress` — this should work unchanged since cloud-provider-kind populates the same field
- Validate that the tunneled addresses are reachable from the test runner

## References

- [cloud-provider-kind repo](https://github.com/kubernetes-sigs/cloud-provider-kind)
- [kind LoadBalancer docs](https://kind.sigs.k8s.io/docs/user/loadbalancer/)
- [MetalLB limitation with Docker Desktop](https://metallb.universe.tf/installation/clouds/) — no L2 on VM-based setups

silvi-t:
> [!IMPORTANT]

Closes #957 

  ## Description
  - Replace MetalLB with `cloud-provider-kind` for LoadBalancer IP assignment in local Kind clusters
  - Support Podman for the local setup
  - Fix Envoy gateway test reliability by waiting for LoadBalancer service readiness

  ## Changes

  ### Local Setup
  - Replace `install-metallb` target with `start-cloud-provider` and `stop-cloud-provider` targets in `make/dependencies.mk`
  - Add `CONTAINER_ENGINE` variable (default: `docker`) to `make/vars.mk`
  - Pass `KIND_EXPERIMENTAL_PROVIDER` to `kind create/delete cluster` commands in `make/kind.mk`
  - Stop `cloud-provider-kind` on `local-cleanup` in `make/local-setup.mk`
  - Remove `METALLB_VERSION` and `METALLB_TIMEOUT` variables

  ### Bug Fix
  - Add `self.service.wait_for_ready()` after Envoy LoadBalancer service creation in `testsuite/gateway/envoy/__init__.py`. Unlike MetalLB which assigns IPs synchronously, `cloud-provider-kind` assigns them asynchronously, so
  tests need to wait for the external IP before sending requests

  ### Documentation
  - Update `README.md` prerequisites to list `cloud-provider-kind` instead of MetalLB
  - Add troubleshooting note for Podman users with another Docker-compatible runtime (Docker Desktop, OrbStack, etc.)
  - Replace macOS limitation warning with Docker/Podman usage instructions

  ## Verification steps
  ```bash
  # Full local setup with Docker (default)
  make local-setup

  # Verify LoadBalancer IPs are assigned
  kubectl get svc -A | grep LoadBalancer

  # Run tests
  make smoke

  # Cleanup
  make local-cleanup
```

## Summary by CodeRabbit

* **Documentation**
  * Updated Quick Start for local Kind clusters with Docker vs Podman guidance, container-engine configuration and macOS compatibility notes.

* **New Features**
  * Replaced MetalLB-based local LoadBalancer setup with cloud-provider-kind and added metrics-server to the startup flow.

* **Chores**
  * Added start/stop cloud-provider automation and integrated it into local setup and cleanup flows.

* **Bug Fixes**
  * Ensure Envoy-related LoadBalancer services are awaited until ready after deployment.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/testsuite/pull/973?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### CODE REVIEW

File: make/dependencies.mk [hunk 1]
```diff
@@ -8,31 +8,25 @@ install-metrics-server: ## Install metrics-server
...
+	@if ! command -v cloud-provider-kind >/dev/null 2>&1; then \
+		echo "ERROR: cloud-provider-kind not found."; \
+		echo "Install it with: go install sigs.k8s.io/cloud-provider-kind@latest"; \
+		echo "  or on macOS: brew install cloud-provider-kind"; \
+		exit 1; \
+	fi
+	@sudo pkill -f cloud-provider-kind 2>/dev/null || true
+	@echo "Starting cloud-provider-kind..."
+	@sudo cloud-provider-kind > /tmp/cloud-provider-kind.log 2>&1 &
+	@sleep 2
```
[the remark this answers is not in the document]
silvi-t: cloud-provider-kind is a developer-only tool — there won't be unrelated instances running on a dev machine.

testsuite/gateway/envoy/__init__.py
```diff
@@ -98,6 +98,7 @@ def commit(self):
             service_type="LoadBalancer",
         )
         self.service.commit()
+        self.service.wait_for_ready()
```
emmaaroche: I know this PR is focused on Kind clusters, but this Envoy code is also used on OpenShift/remote clusters where LoadBalancers can be slow. Other LoadBalancer services (`mockserver.py`, `gateway_api/gateway.py`) pass `timeout` and `settings["control_plane"]["slow_loadbalancers"]`. Maybe we could use the same pattern here for consistency?

silvi-t: Added

README.md
```diff
@@ -38,11 +38,13 @@ For local development and testing, you can set up a complete Kuadrant environmen
...
 * [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation)
+* [cloud-provider-kind](https://github.com/kubernetes-sigs/cloud-provider-kind) (LoadBalancer support)
 * [Helm](https://helm.sh/docs/intro/install/)
 * [jq](https://jqlang.github.io/jq/download/) (JSON processor)
 
-> **⚠️ macOS Limitation:**
-> MetalLB LoadBalancer services have limited functionality on macOS due to Docker Desktop's VM isolation. While MetalLB will work inside the cluster, LoadBalancer IPs won't be accessible from your Mac host machine. **For macOS users, we recommend running tests in containers** (see [From a Container](#from-a-container) section above) in addition to the local Kind setup.
+> **Docker users:** Set `CONTAINER_ENGINE=docker` when running `make local-setup`. Podman is used by default.
+
+> **macOS container runtime:** This setup has been tested with [Podman Desktop](https://podman-desktop.io/). Other Docker-compatible runtimes (OrbStack, Docker Desktop, Rancher Desktop) may not work because `cloud-provider-kind` uses `docker ps` internally to discover Kind containers — if the Docker CLI points to a different runtime, it won't find the Podman-based Kind cluster.
```
emmaaroche: The reason OrbStack doesn't work for me is slightly different than what is noted here. In my testing, `cloud-provider-kind` did find the Kind containers fine and assigned LoadBalancer IPs successfully, but services weren't accessible via those IPs (connection timeouts). The Envoy containers were using Docker port mapping instead of binding to the LoadBalancer IPs directly.

So the container discovery works with OrbStack, but the LoadBalancer networking doesn't. I didn't test Docker Desktop or Rancher Desktop so not sure if they have the same issue however to comment on those.

silvi-t: Thanks for your findings. I've updated the note to be more general. Let me know if that looks okay.

emmaaroche (dismissed):
LGTM 👍

make/dependencies.mk
```diff
@@ -8,31 +8,25 @@ install-metrics-server: ## Install metrics-server
...
+start-cloud-provider: ## Start cloud-provider-kind for LoadBalancer services
+	@if ! command -v cloud-provider-kind >/dev/null 2>&1; then \
+		echo "ERROR: cloud-provider-kind not found."; \
+		echo "Install it with: go install sigs.k8s.io/cloud-provider-kind@latest"; \
+		echo "  or on macOS: brew install cloud-provider-kind"; \
+		exit 1; \
+	fi
+	@sudo pkill -f cloud-provider-kind 2>/dev/null || true
+	@echo "Starting cloud-provider-kind..."
+	@sudo cloud-provider-kind > /tmp/cloud-provider-kind.log 2>&1 &
```
crstrn13: sudo won't work on linux. I think we should do something that detects your OS and does sudo if needed.

make/vars.mk
```diff
@@ -3,6 +3,7 @@
 
 # Kind cluster configuration
 KIND_CLUSTER_NAME ?= kuadrant-local
+CONTAINER_ENGINE ?= podman
```
crstrn13: I've had issues with podman on linux. There is one way to make it work, by setting the `DOCKER_HOST` env var. I am not sure if this is just my setup, but I'd keep the docker default instead as I think it is more spread throughout the team and in general.

make/dependencies.mk
```diff
@@ -8,31 +8,25 @@ install-metrics-server: ## Install metrics-server
...
-	@echo "MetalLB installed with IP pool 172.18.255.200-172.18.255.250"
+.PHONY: start-cloud-provider
+start-cloud-provider: ## Start cloud-provider-kind for LoadBalancer services
+	@if ! command -v cloud-provider-kind >/dev/null 2>&1; then \
+		echo "ERROR: cloud-provider-kind not found."; \
+		echo "Install it with: go install sigs.k8s.io/cloud-provider-kind@latest"; \
+		echo "  or on macOS: brew install cloud-provider-kind"; \
+		exit 1; \
+	fi
+	@sudo pkill -f cloud-provider-kind 2>/dev/null || true
```
crstrn13: When make runs this, the shell's own command line is:

  /bin/sh -c 'sudo pkill -f cloud-provider-kind 2>/dev/null || true'

  pkill -f matches against the full command line of every process. That shell process's command line contains the string cloud-provider-kind — so pkill matches and kills its own parent shell. Make sees the shell get terminated by a signal and reports Terminated.

  The bracket trick '[c]loud-provider-kind' works because:
  - As a regex, [c]loud-provider-kind matches the string cloud-provider-kind (the target process)
  - As a literal string in the shell's command line, it reads [c]loud-provider-kind, which does not match the regex — so pkill skips itself and its parent shell
  
  It's a standard Unix idiom — you see the same thing with grep (ps aux | grep '[n]ginx' to avoid matching the grep itself).

make/vars.mk
silvi-t: Done

make/dependencies.mk
silvi-t: I added system check and different commands for linux/macOS. Please test the new approach on linux and let me know how it goes.

make/dependencies.mk
```diff
@@ -8,31 +8,37 @@ install-metrics-server: ## Install metrics-server
...
-	@echo "MetalLB installed with IP pool 172.18.255.200-172.18.255.250"
+.PHONY: start-cloud-provider
+start-cloud-provider: ## Start cloud-provider-kind for LoadBalancer services
+	@if ! command -v cloud-provider-kind >/dev/null 2>&1; then \
+		echo "ERROR: cloud-provider-kind not found."; \
+		echo "Install it with: go install sigs.k8s.io/cloud-provider-kind@latest"; \
+		echo "  or on macOS: brew install cloud-provider-kind"; \
+		exit 1; \
+	fi
+	@if pgrep -f '[c]loud-provider-kind' >/dev/null 2>&1; then \
```
crstrn13:
```suggestion
	@if pgrep -x cloud-provider-kind >/dev/null 2>&1; then \
```

crstrn13: Wasn't able to make it work without this correction

crstrn13 (changes requested):

make/dependencies.mk
silvi-t: Done

crstrn13 (approved):
Tested on Fedora 43, works like a charm! 🎉 :shipit:

PR #979, opened 2026-05-14
feat(commands): update rerun command to support rptool

silvi-t:
## Description
  - Update the `/rerun-report-portal-tests` command to support fetching failed tests directly from Report Portal via `rptool`
  - Accept test names in pytest path format (preferred) alongside legacy dot-notation with auto-detection and conversion
  - Make RP project optional in rptool queries, falling back to the user's default configuration

## Verification steps
  - No test files changed — this is a Claude Code command definition update
  - Verify by running `/rerun-report-portal-tests` with:
    - rptool arguments (RP project + launch name + test target)


averevki (approved):
Looks good 👍 I think this command could benefit further from being a Claude Skill instead now

PR #983, opened 2026-05-18
fix: collect launch properties when some cluster lacks Kuadrant

silvi-t:
## Description
  - During nightly runs, when the first cluster was missing Kuadrant, none of the Report Portal launch properties were recorded — the info_collector tests were silently skipped by the global `has_kuadrant()` check, and the
  `openshift_version` fixture cascaded a failure through the `cluster` fixture (which requires the `kuadrant` namespace)
  - Made `make collect` / info_collector more resilient: properties are now scraped from the first cluster that has the required components, so we still get records from the remaining clusters
  - Clusters missing Kuadrant now appear in the launch description with `not installed` instead of being omitted

  ## Changes

  ### Bug Fixes
  - Bypass the `has_kuadrant()` check in `pytest_runtest_setup` for `info_collector.py` — it handles missing components gracefully via try/except
  - Decouple the `openshift_version` fixture from the `cluster` fixture by connecting to the `default` namespace instead, so OCP version detection works regardless of the `kuadrant` namespace

  ### Enhancements
  - Add `_first_connected()` helper in `info_collector.py` that iterates all configured clusters and returns the first `(cluster_client, project)` pair with a connected namespace
  - Update `gather_cluster_versions()`, `test_kube_context`, `test_kuadrant_properties`, and `test_istio_properties` to fall back to secondary clusters when the first lacks `kuadrant-system` or `istio-system`
  - Update `_collect_single_cluster` in `component_metadata.py` to fall back to the `default` namespace and report `kuadrant_image: not installed` for clusters missing Kuadrant

## Verification steps
  
  ```bash
  # 1. Run info_collector against a setup where the first cluster lacks Kuadrant
  make collect
  
  # 2. Import results into Report Portal and verify the launch description
  rptool write --launch-name <name> junit-00-collect.xml
  ```
  - In Report Portal, verify that the launch description shows all clusters, with `kuadrant_image: not installed` for clusters missing Kuadrant

## Summary by CodeRabbit

* **New Features**
  * Improved cluster reachability checks and direct OpenShift version detection per cluster.
  * Component image reporting now includes images without explicit tags (tag shown as empty).

* **Bug Fixes**
  * Skip unreachable clusters early; record kuadrant as "not installed" when system project is disconnected.
  * Multi-cluster collection now records per-cluster properties consistently and avoids failing on missing namespaces.

### DIFF
--- a/testsuite/component_metadata.py
+++ b/testsuite/component_metadata.py
@@ -21,13 +21,14 @@ class ReportPortalMetadataCollector:
 
     def collect_all_clusters(self):
         """Collect metadata from all configured clusters."""
-        clusters_config = self._get_cluster_configurations()
+        clusters_config = self.get_cluster_configurations()
         for cluster_name, cluster_client in clusters_config:
             metadata = self._collect_single_cluster(cluster_client)
             if metadata:
                 self.all_cluster_metadata[cluster_name] = metadata
 
-    def _get_cluster_configurations(self):
+    @staticmethod
+    def get_cluster_configurations():
         """Get cluster configurations from settings."""
         clusters_config = [("cluster1", settings["control_plane"]["cluster"])]
         if cluster2 := settings["control_plane"].get("cluster2"):
@@ -38,14 +39,16 @@ class ReportPortalMetadataCollector:
 
     def _collect_single_cluster(self, cluster_client):
         """Collect metadata for a single cluster."""
-        project = cluster_client.change_project(settings["service_protection"]["system_project"])
-        if not project.connected:
+        if not cluster_client.is_reachable:
             return None
 
+        project = cluster_client.change_project(settings["service_protection"]["system_project"])
+        metadata = self._get_kuadrant_metadata(project) if project.connected else {"kuadrant_image": "not installed"}
+
         return {
-            "metadata": self._get_kuadrant_metadata(project),
+            "metadata": metadata,
             "console_url": self._get_console_url(cluster_client.api_url),
-            "ocp_version": self.get_ocp_version(project),
+            "ocp_version": cluster_client.ocp_version,
         }
 
     @staticmethod
@@ -77,23 +80,6 @@ class ReportPortalMetadataCollector:
             console_hostname = hostname.replace("api.", "console-openshift-console.apps.", 1)
             return f"https://{console_hostname}"
         return api_url
-
-    @staticmethod
-    def get_ocp_version(project) -> Optional[str]:
-        """Retrieve and format OCP version from cluster."""
-        try:
-            with project.context:
-                version_result = oc.selector("clusterversion").objects()
-                if version_result:
-                    ocp_version = version_result[0].model.status.history[0].version
-                    if ocp_version:
-                        parts = ocp_version.split(".")
-                        if len(parts) >= 2:
-                            return f"{parts[0]}.{parts[1]}"
-        except (oc.OpenShiftPythonException, AttributeError, KeyError, IndexError, ValueError) as e:
-            logger.warning("Failed to get OCP version: %s", e)
-
-        return None
 
     @staticmethod
     def get_kubernetes_version(project) -> Optional[str]:
--- a/testsuite/component_metadata.py
+++ b/testsuite/component_metadata.py
@@ -116,13 +116,12 @@ class ReportPortalMetadataCollector:
                     if normalised_image in seen:
                         continue
                     seen.add(normalised_image)
-                    print(f"{image=}")
                     image_name = normalised_image.split("/")[-1]
                     if ":" in image_name:
                         name, tag = image_name.rsplit(":", 1)
                         images.append((name, tag, image))
                     else:
-                        logger.debug("Skipping image without tag: %s", image)
+                        images.append((image_name, None, image))
         except (oc.OpenShiftPythonException, AttributeError, KeyError, IndexError, ValueError) as e:
             logger.warning("Failed to get images from %s: %s", project, e)
 
--- a/testsuite/kubernetes/client.py
+++ b/testsuite/kubernetes/client.py
@@ -114,7 +114,7 @@ class KubernetesClient:
             return False
         return True
 
-    @property
+    @cached_property
     def ocp_version(self) -> Optional[str]:
         """Returns the OpenShift version (major.minor) or None if not available."""
         result = self.do_action(
--- a/testsuite/kubernetes/client.py
+++ b/testsuite/kubernetes/client.py
@@ -1,6 +1,8 @@
 """This module implements an KubernetesCLI interface using oc/kubectl binary commands."""
 
+import logging
 from functools import cached_property
+from typing import Optional
 from urllib.parse import urlparse
 import tempfile
 import yaml
@@ -13,6 +15,8 @@ from testsuite.kubernetes.service import Service
 from .service_account import ServiceAccount
 from .deployment import Deployment
 from .secret import Secret
+
+logger = logging.getLogger(__name__)
 
 
 class KubernetesClient:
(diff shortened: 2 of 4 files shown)


testsuite/tests/conftest.py
```diff
@@ -422,9 +426,12 @@ def dns_provider_secret(testconfig):
 
 
 @pytest.fixture(scope="session")
-def openshift_version(cluster):
+def openshift_version(testconfig):
     """Get OpenShift cluster version"""
-    result = cluster.do_action(
+    project = testconfig["control_plane"]["cluster"].change_project("default")
+    if not project.connected:
+        return None
```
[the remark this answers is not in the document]
silvi-t: The `default` namespace exists on every Kubernetes cluster and cannot be deleted. If it's not reachable, the cluster itself is unreachable and no tests would run regardless.

testsuite/tests/info_collector.py
```diff
@@ -30,16 +30,33 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _first_connected(namespace):
```
averevki: Can you please remind me why this collector wasn't collecting information from every cluster, and is using only the first one it finds available?

testsuite/tests/info_collector.py
```diff
@@ -79,17 +96,21 @@ def test_cluster_properties(record_testsuite_property):
...
     if kube_context:
         record_testsuite_property("kube_context", kube_context)
 
 
 def test_kuadrant_properties(record_testsuite_property):
-    """Record kuadrant related properties"""
-    cluster_client = settings["control_plane"]["cluster"]
-    project = cluster_client.change_project("kuadrant-system")
+    """Record kuadrant related properties from the first cluster with kuadrant-system."""
+    _, project = _first_connected("kuadrant-system")
```
averevki: hardcoded namespace

testsuite/tests/info_collector.py
```diff
@@ -30,16 +30,33 @@
...
 logger = logging.getLogger(__name__)
 
+
+def _first_connected(namespace):
+    """Return the first (cluster_client, project) connected to the given namespace, or (None, None)."""
+    clusters = [settings["control_plane"]["cluster"]]
+    if cluster2 := settings["control_plane"].get("cluster2"):
+        clusters.append(cluster2)
+    if cluster3 := settings["control_plane"].get("cluster3"):
+        clusters.append(cluster3)
```
averevki: I feel like `_get_cluster_configurations` collector method is already doing similar thing. You think it can be reused?

testsuite/tests/conftest.py
averevki: I would prefer `oc whoami` or `oc api-versions`, and a dedicated openshift client method, if you only need to verify cluster connectivity by switching to default namespace.

testsuite/component_metadata.py
```diff
@@ -39,13 +39,21 @@ def _get_cluster_configurations(self):
...
+                "ocp_version": self.get_ocp_version(project),
+            }
+
+        fallback = cluster_client.change_project("default")
+        if not fallback.connected:
             return None
 
         return {
-            "metadata": self._get_kuadrant_metadata(project),
+            "metadata": {"kuadrant_image": "not installed"},
```
averevki: This can be handled inside the method. 

You should try to reuse the same dict here and only substitute the values.

testsuite/component_metadata.py
```diff
@@ -39,13 +39,21 @@ def _get_cluster_configurations(self):
...
-        if not project.connected:
+        if project.connected:
+            return {
+                "metadata": self._get_kuadrant_metadata(project),
+                "console_url": self._get_console_url(cluster_client.api_url),
+                "ocp_version": self.get_ocp_version(project),
+            }
+
+        fallback = cluster_client.change_project("default")
+        if not fallback.connected:
```
averevki: same for default namespace

testsuite/tests/info_collector.py
silvi-t: In our multicluster pipeline, all clusters share the same configuration. They're deployed with the same kuadrant-operator image, so the component images and versions are identical across clusters. The launch attributes this collector gathers (mostly Kuadrant component images extracted from the operator) would be duplicated if we collected from every cluster. Using only one cluster avoids that redundancy while still capturing all the relevant information.
This approach can change in the future if we need cluster-specific properties, but right now collecting from multiple clusters would only result in duplicate attributes.

averevki: I would still prefer collecting the real information from each cluster, even if collect task would take 3x time.

I don't like that this already confusing process is getting refactored with the new confusing algorithm, I need to also be aware of now.

This might be the moment to change this, and assign `None` values to missing configurations instead of blindly trusting the install pipelines.

silvi-t: I'd like to pull @zkraus  into this conversation to get his opinion as well. My concern is that collecting attributes from every cluster would result in unnecessary duplication (since all clusters share the same operator image, the attributes would be identical). We'd either have duplicate keys or need cluster-prefixed names (e.g. `cluster1_kuadrant-operator`, `cluster2_kuadrant-operator`), which makes the launch attributes harder to read and filter on in Report Portal without adding useful information.

testsuite/tests/info_collector.py
silvi-t: Done

testsuite/tests/info_collector.py
silvi-t: Done

testsuite/component_metadata.py
silvi-t: Done

testsuite/component_metadata.py
silvi-t: Done

testsuite/tests/conftest.py
silvi-t: Done

testsuite/tests/info_collector.py
zkraus: I confirm, that the reason was to collect only one, because of assumption of unified environment.
But there should be no harm of collecting it. It will be preparation, if/when we start testing interoperability.

If the versions would be equal, I would not set duplicated attributes (if that is even possible), so maybe checking if there is already equal key and value. This should be simple enough, and we can change that if necessary.

On the other hand, as we will be getting the version and any data, from each cluster -- which is definitely beneficial -- please do use logger, and log the information collected per cluster in the collection test. Log as much as you think would be valuable. In the log, it is accessible, readable, and will not hurt anyone. will not overuse attributes. I think that is nearly ideal place to put a large amount of information, in case we need it later.  -- Next upgrade, would be to put this data as an attachment in json/yaml format (next time).

TL;DR: I agree, collect info from all clusters, deduplicate attributes, log everything.

silvi-t:
@averevki , @zkraus 
 ## Brief summary of changes

  **`testsuite/kubernetes/client.py`**
  - Added `is_reachable` property — checks cluster connectivity using `oc api-versions` (works on both OCP and Kind)
  - Added `ocp_version` property — returns OpenShift version (major.minor) using `status.history[0].version`

  **`testsuite/component_metadata.py`**
  - Made `_get_cluster_configurations` a public `@staticmethod` (`get_cluster_configurations`) for reuse
  - Refactored `_collect_single_cluster` to use `is_reachable` instead of `change_project("default").connected`
  - Removed `get_ocp_version` — replaced by `KubernetesClient.ocp_version`
  - Removed fallback to `default` namespace — single dict structure with `"not installed"` for missing Kuadrant

  **`testsuite/tests/info_collector.py`**
  - Replaced `_first_connected` with `_all_cluster_projects` — iterates all clusters, yields `None` for missing namespaces
  - Added `_print_cluster_data` helper — standardized per-cluster logging
  - Added `_record_unique` helper — deduplicates attributes before recording to Report Portal
  - All property tests now collect from every cluster, log per-cluster data, and only record unique attributes
  - Replaced hardcoded `"kuadrant-system"` with `settings["service_protection"]["system_project"]`
  - Removed `test_kube_context`

  **`testsuite/tests/conftest.py`**
  - `openshift_version` fixture now uses `cluster.ocp_version` and `status.history[0].version`

testsuite/tests/info_collector.py
```diff
@@ -70,39 +87,62 @@ def test_launch_description(record_testsuite_property):
...
+        versions = {
+            "kubernetes": ReportPortalMetadataCollector.get_kubernetes_version(project),
+            "openshift": cluster.ocp_version,
+        }
+        cluster_data[cluster_name] = [f"{k}:{v}" for k, v in versions.items()]
+        for key, value in versions.items():
+            properties.append((key, value))
+
+    _print_cluster_data(cluster_data)
+    _record_unique(record_testsuite_property, properties)
```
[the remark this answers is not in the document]
silvi-t: I intentionally removed that test.

testsuite/kubernetes/client.py
```diff
@@ -100,6 +104,30 @@ def connected(self):
...
+    def is_reachable(self):
+        """Returns True if the cluster is reachable, without depending on any specific namespace."""
+        try:
+            self.do_action("api-versions")
+        except OpenShiftPythonException as e:
+            logger.warning("Cluster is not reachable: %s", e)
+            return False
+        return True
+
+    @property
```
averevki:
```suggestion
    @cached_property
```

silvi-t: done

averevki (approved):

Issue #984, opened 2026-05-18
Add wait_for_ready() to singlecluster route fixture and EnvoyVirtualRoute


averevki:
> [!NOTE]
> **Missing route waits on multicluster tests routes require additional investigation and issues opened**

The default `route` fixture in `testsuite/tests/singlecluster/conftest.py` does not call `wait_for_ready()` after `commit()`. This was missed when the method was added to `HTTPRoute`.

Adding the call directly causes some tests to fail and needs investigation.

### Changes needed

**1. Add no-op `wait_for_ready()` to `EnvoyVirtualRoute`**

`EnvoyVirtualRoute` lacks a `wait_for_ready()` method, so calling it unconditionally on the route fixture would break standalone Authorino tests. Add the following placeholder matching the existing `commit()`/`delete()` pattern:

```python
# testsuite/gateway/envoy/route.py
    class HTTPRoute:

    def wait_for_ready(self):
        return
```

**2. Add `wait_for_ready()` to the singlecluster route fixture**

```python
# testsuite/tests/singlecluster/conftest.py
    def route():
    ...
    route.commit()
    route.wait_for_ready()
    return route
```

**3. Investigate and fix test failures triggered by the wait**
Some tests fail when the wait is added. These need to be investigated and fixed before merging.

PR #985, opened 2026-05-18
fix: add missing waits for httproute

averevki:
## Summary

Add missing waits for httproute. Adding the missing wait on the route fixture in singlecluster's conftest triggered a few sporadic testsuite failures, so I made a follow-up issue for it: #984

## Summary by CodeRabbit

* **Tests**
  * Added explicit readiness checks for newly created routes across multiple test suites (gRPC, HTTP and UI tests). Tests now wait for routes to be fully ready before proceeding, improving stability and reducing intermittent failures caused by race conditions during test setup.

### DIFF
--- a/testsuite/tests/singlecluster/grpc/conftest.py
+++ b/testsuite/tests/singlecluster/grpc/conftest.py
@@ -26,6 +26,7 @@ def route(request, gateway, blame, hostname, backend, module_label):
     route.add_backend(backend)
     request.addfinalizer(route.delete)
     route.commit()
+    route.wait_for_ready()
     return route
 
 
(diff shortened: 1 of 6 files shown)


emmaaroche (dismissed):
Should this PR also cover the multicluster route fixture? The `routes` fixture in [testsuite/tests/multicluster/conftest.py:94](https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/multicluster/conftest.py#L94-L107) commits without `wait_for_ready()`.

Just a question / not a blocker though, still approving!

emmaaroche:
To follow up on my last comment these httproute fixtures also do not currently have a `wait_for_ready`:

- [testsuite/tests/singlecluster/ui/console_plugin/overview/test_overview.py#L96](https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/singlecluster/ui/console_plugin/overview/test_overview.py#L96) - HTTPRoute and GRPCRoute in test function
- [testsuite/tests/singlecluster/tracing/control_plane/test_control_plane_errors.py#L20](https://github.com/Kuadrant/testsuite/blob/main/testsuite/tests/singlecluster/tracing/control_plane/test_control_plane_errors.py#L20) - `orphan_test_route` fixture

averevki:
@emmaaroche Thanks for pointing this out, I've added missing waits you mentioned except the ones for multicluster tests. I think it was intended for them not to use wait_for_ready on some objects as it needs to wait until the same object is created on a second cluster. However, I think it still might be possible to add missing waits on multicluster objects, but it would require a bit deeper investigation. I have added multicluster route waits to the follow up issue I created.

emmaaroche (approved):

PR #995, opened 2026-06-08
fix(planpolicy): make keycloak attribute name with planID unique

averevki:
## Description

Fix the `Attribute configuration already exists with 'name':'planID'` happening on planpolicy tests in pipelines, by making the keycloak attribute name unique

## Summary by CodeRabbit

* **Tests**
  * Refactored plan policy test suites to use dynamic fixture-based configuration instead of static definitions, improving test flexibility and maintainability. This update applies to both single-cluster and gRPC test implementations, enabling more robust validation of plan tier configurations and rate-limiting policies.


silvi-t (approved):
LGTM

PR #1006, opened 2026-06-22
fix: increase mockserver CPU limit

silvi-t:
> [!IMPORTANT]

  ## Description
  - Increase `MockserverBackend` CPU limit from `100m` to `500m` to prevent JVM cold start timeouts
  - With only `100m` CPU, the mockserver JVM takes 20-30s to start processing requests, but the test client times out after 5s and retries automatically
  - Each retry still counts against the rate limit at the gateway level, so by the time mockserver is ready, the rate limit is already exhausted and the test fails
  - Confirmed with `curl` that the gateway and rate limiting work correctly — the issue is purely the slow mockserver + client retries
  - The issue was only caught when running a single test file (`test_rate_limit_authz.py`) in isolation, not during the full `make kuadrant` suite where mockserver is already warmed up by earlier tests

  ## Changes
  ### Bug Fix
  - Increased `limits_cpu` from `100m` to `500m` in `testsuite/backend/mockserver.py` to give the JVM enough CPU for cold start while still preventing cluster resource starvation that #990 was solving

  ## Verification steps
  ```sh
  poetry run pytest -vv testsuite/tests/singlecluster/test_rate_limit_authz.py
  ```

## Summary by CodeRabbit

* **Chores**
  * Optimised backend test infrastructure resource allocation.

### DIFF
--- a/testsuite/backend/mockserver.py
+++ b/testsuite/backend/mockserver.py
@@ -80,7 +80,7 @@ class MockserverBackend(Backend):
             selector=Selector(matchLabels=match_labels),
             labels={"app": self.label},
             resources=ContainerResources(
-                limits_cpu="100m", requests_cpu="10m", limits_memory="300Mi", requests_memory="200Mi"
+                limits_cpu="500m", requests_cpu="10m", limits_memory="300Mi", requests_memory="200Mi"
             ),
             lifecycle={"postStart": {"exec": {"command": ["/bin/sh", "init-mockserver"]}}},
             volumes=self.config.volumes if self.config else None,


testsuite/backend/mockserver.py
```diff
@@ -80,7 +80,7 @@ def commit(self):
             selector=Selector(matchLabels=match_labels),
             labels={"app": self.label},
             resources=ContainerResources(
-                limits_cpu="100m", requests_cpu="10m", limits_memory="300Mi", requests_memory="200Mi"
+                limits_cpu="500m", requests_cpu="10m", limits_memory="300Mi", requests_memory="200Mi"
```
averevki, line 83: 500 might be a bit high, as even the kuadrant operators use around 200-300. However I haven't seen how mockserver backend behaves during the peak loads now, and setting CPU limit a bit higher than that shouldn't be a huge problem. LGTM

averevki (approved):

PR #1009, opened 2026-06-25
feat: patch kuadrant operator image via KUADRANT_OPERATOR_IMAGE

crstrn13:
## Summary
- Wire up the existing `KUADRANT_OPERATOR_IMAGE` variable (defined but unused in `vars.mk`) to actually patch the operator deployment image
- Add a `patch-kuadrant-operator-image` target that uses `kubectl set image` to swap the manager container image when the variable is set
- The patch runs after operator deployment (both components and helm modes), so you get the correct RBAC/CRDs/config from the gitref but run a custom-built image

## Usage
```bash
KUADRANT_OPERATOR_IMAGE=quay.io/kuadrant/kuadrant-operator:my-tag make local-setup
```

Or standalone on an existing cluster:
```bash
KUADRANT_OPERATOR_IMAGE=quay.io/kuadrant/kuadrant-operator:my-branch make patch-kuadrant-operator-image
```

## Test plan
- [ ] Run `make local-setup` without `KUADRANT_OPERATOR_IMAGE` — operator deploys as before, skip message printed
- [ ] Run `make local-setup` with `KUADRANT_OPERATOR_IMAGE=<custom-image>` — operator image is patched after deployment
- [ ] Run `make patch-kuadrant-operator-image` standalone on an existing cluster with the variable set

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * The Kuadrant Operator deployment now automatically applies a configured container image after installation.
  * When a custom operator image is set, the deployment waits for the updated rollout to complete.
  * If no custom image is provided, the deployment now skips the image update gracefully with a clear message.

### DIFF
--- a/make/kuadrant.mk
+++ b/make/kuadrant.mk
@@ -30,6 +30,7 @@ endif
 	$(MAKE) patch-kuadrant-operator-env
 	@echo "Kuadrant Operator $(KUADRANT_OPERATOR_VERSION) installed from Helm"
 endif
+	$(MAKE) patch-kuadrant-operator-image
 ifeq ($(INSTALL_TRACING),true)
 	@echo "Configuring OTEL tracing on limitador-operator..."
 	@kubectl set env deployment/limitador-operator-controller-manager \
@@ -38,6 +39,19 @@ ifeq ($(INSTALL_TRACING),true)
 		OTEL_EXPORTER_OTLP_INSECURE=true
 	@kubectl rollout status deployment/limitador-operator-controller-manager \
 		-n $(KUADRANT_NAMESPACE) --timeout=$(KUBECTL_TIMEOUT)
+endif
+
+.PHONY: patch-kuadrant-operator-image
+patch-kuadrant-operator-image: ## Patch Kuadrant Operator deployment with custom image
+ifneq ($(KUADRANT_OPERATOR_IMAGE),)
+	@echo "Patching Kuadrant Operator image to $(KUADRANT_OPERATOR_IMAGE)..."
+	kubectl set image deployment/kuadrant-operator-controller-manager \
+		-n $(KUADRANT_NAMESPACE) \
+		manager=$(KUADRANT_OPERATOR_IMAGE)
+	kubectl -n $(KUADRANT_NAMESPACE) rollout status deployment/kuadrant-operator-controller-manager --timeout=$(KUBECTL_TIMEOUT)
+	@echo "Kuadrant Operator image patched"
+else
+	@echo "No custom operator image specified (KUADRANT_OPERATOR_IMAGE not set), skipping"
 endif
 
 .PHONY: patch-kuadrant-operator-env


make/kuadrant.mk
```diff
@@ -30,6 +30,7 @@ endif
 	$(MAKE) patch-kuadrant-operator-env
 	@echo "Kuadrant Operator $(KUADRANT_OPERATOR_VERSION) installed from Helm"
 endif
+	$(MAKE) patch-kuadrant-operator-image
```
averevki, line 33: You think setting a custom kuadrant-operator image could be useful to expose for configuration in the Kuadrant's helm chart values yaml? Here, I suppose, we just use custom images for testing, but could it be also useful for other Kuadrant users?

make/kuadrant.mk
```diff
@@ -40,6 +41,19 @@ ifeq ($(INSTALL_TRACING),true)
...
 endif
 
+.PHONY: patch-kuadrant-operator-image
+patch-kuadrant-operator-image: ## Patch Kuadrant Operator deployment with custom image
+ifneq ($(KUADRANT_OPERATOR_IMAGE),)
+	@echo "Patching Kuadrant Operator image to $(KUADRANT_OPERATOR_IMAGE)..."
+	kubectl set image deployment/kuadrant-operator-controller-manager \
+		-n $(KUADRANT_NAMESPACE) \
+		manager=$(KUADRANT_OPERATOR_IMAGE)
+	kubectl -n $(KUADRANT_NAMESPACE) rollout status deployment/kuadrant-operator-controller-manager --timeout=$(KUBECTL_TIMEOUT)
```
averevki, line 51: Would this stop the make target execution if fails?

crstrn13: yes, the rollout would fail on timeout if the pod isn't able to start. Maybe it isn't able to pull the image or stuff like that.

make/kuadrant.mk
crstrn13: Good point, don't think a regular user should want to do this. The main reason I added this is because you can no longer change the wasm-shim image easily, since it is _embedded_ in the kuadrant operator for now. But our releases should be the only versions a regular user should care about. Does that answer your question?

azgabur (commented):
What is the advantage for patching kuadrant opeator image instead of deploying the needed version from released catalogSource? I can see only disadvantage of disconnecting operator version mentioned in catalogSource used for helm install and actual used kuadrant operator image. Such patch can be very hidden on long running cluster.

Why is this needed?

Patching relatedImages field in Subscription would be more persitent as well.

crstrn13:
@azgabur this is only related to local deployment. Sometimes you might need a quick check during the day that encompasses version x of kuadrant-operator and version y of the wasm-shim. The local setup doesn't even use olm to begin with.

make/kuadrant.mk
averevki: Should this failure be handled so the script can continue its execution, or at least this wait to go to the very end of the script? I see there is still some stuff being configured after this deployment patch, like tracing and Kuadrant's CR deployment

crstrn13: I think it would be more convenient to fail if it is not being able to pull the image, so you can check if it is accessible etc and thus you can make the changes needed so it will be able to pull the image, rather than giving the false impression that the env is in a working state.

averevki (approved):

PR #1011, opened 2026-07-02
refactor(test): restructure OIDC policy tests and add redirect bug coverage

crstrn13:
## Description

Restructures the OIDC policy test suite for clarity, reduces duplication, and adds new test coverage for the OIDCPolicy bug fixes from [kuadrant-operator#2032](https://github.com/Kuadrant/kuadrant-operator/pull/2032).

## Changes

**Test restructuring:**
- Removed the `OIDCTestClient` wrapper class (`testsuite/oidc/test_client.py`) — tests now create Keycloak clients directly via `ClientConfig` + `KeycloakOpenID`, removing an unnecessary abstraction layer
- Moved `ClientConfig` dataclass to `testsuite/oidc/keycloak/objects.py` alongside other Keycloak object wrappers
- Extracted the shared `auth` fixture to the OIDC conftest (was duplicated across 4 test files)
- Split monolithic test functions into focused, single-concern tests across dedicated modules: confidential client, public client, service client, JWT validation, and redirect handling

**New test coverage (kuadrant-operator#2032):**
- `test_oidc_policy_redirect.py` — validates fixes for:
  - Redirect URI includes non-standard listener port (tests with port 8001)
  - Target cookie preserves query string parameters after OIDC auth redirect
  - Authenticated requests with query parameters succeed
- `test_oidc_policy_custom_redirect.py` — validates custom `redirectURI`:
  - When `provider.redirectURI` is set, the authorize redirect uses the custom callback URL
  - The custom callback path replaces the default `/auth/callback`

**Infrastructure:**
- `LoadBalancerServiceExposer` hostname now includes `cluster.project` to avoid collisions when multiple namespaces share a domain

## Known gaps

- OPA cookie parser fix for `=` in JWT base64 values (operator Bug 2) — this is only exercisable via the full OIDC authorization code flow (callback route), which requires programmatic Keycloak login. The operator has unit tests for this fix.

## Verification

- Ran full OIDC test suite against a Kind cluster with Kuadrant deployed
- All linting checks pass (`pylint`, `mypy`, `black`)


testsuite/gateway/exposers.py
```diff
@@ -69,7 +69,7 @@ class LoadBalancerServiceExposer(Exposer):
     """Exposer using Load Balancer service for Gateway"""
 
     def expose_hostname(self, name, gateway: Gateway) -> Hostname:
-        hostname = f"{name}.{self.base_domain}"
+        hostname = f"{name}-{self.cluster.project}.{self.base_domain}"
```
crstrn13, line 72: OpenShiftExposer creates Routes and the resulting hostname is `{name}.{cluster.apps_url}`, which naturally includes the project context since Routes are namespace-scoped. On Kind, the `LoadBalancerServiceExposer` was just doing `{name}.test.com` — no project/namespace in the hostname at all. Adding `cluster.project` makes the Kind hostname format mirror what OpenShift produces, so tests don't break when switching between the two environments.

PR #1010, opened 2026-07-01
chore: Add readiness probe to mockserver backend

azgabur:
## Description
Added readiness probe and wait_for_ready on MockserverBackend as the startup takes some time (Java app).

Values taken from https://github.com/mock-server/mockserver-monorepo/blob/master/helm/mockserver/templates/deployment.yaml#L104

## Changes
Added readiness probe, wait for ready on deployment

## Verification
```sh
make smoke
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mock server startup reliability by waiting for readiness before continuing setup.
  * Added readiness probe configuration to the mock server deployment to reduce timing-related test failures.  
* **Chores**
  * Introduced dedicated readiness timing constants for the mock server test environment.

PR #1012, opened 2026-07-03
revert: Mockserver readiness probe

azgabur:
## Description
Previous PR https://github.com/Kuadrant/testsuite/pull/1010 added readiness probe for MockserverBackend but did not set configuration option for enabling it. `/mockserver/ready` endpoint is only valid in newer versions of mockserver, but we are using older version without this endpoint. This went unnoticed due to test expectation which made every endpoint valid, even `/mockserver` prefixed ones. Test `testsuite/tests/singlecluster/gateway/dnspolicy/health_check/test_additional_headers.py` does not use this test expectation so the readiness check never finished. 

After careful consideration and checking possibility of using newer Mockserver version 7.4.0 with the endpoint `/mockserver/ready` the job of readiness check is done by `init-mockserver` script https://github.com/3scale-qe/mockserver-container/blob/main/init-mockserver . The script wont finish until new expectations are confirmed to be set by API.

>The Container's status is not set to RUNNING until the postStart handler completes.
https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion

Only time the Mockserver behaves strangely with 503 responses is when CPU throttled. PR https://github.com/Kuadrant/testsuite/pull/1006 fixed that. 

## Changes
Reverted #1010

## Verification
```sh
make smoke
```
```sh
make testsuite/tests/singlecluster/gateway/dnspolicy/health_check/test_additional_headers.py
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container environment handling so existing settings are preserved and applied together.
  * Simplified service startup checks, avoiding redundant readiness handling during deployment.

* **Chores**
  * Updated timeout settings used by the test suite for DNS propagation and readiness-related waits.

### DIFF
--- a/testsuite/backend/mockserver.py
+++ b/testsuite/backend/mockserver.py
@@ -6,13 +6,7 @@ from testsuite.kubernetes import Selector
 from testsuite.kubernetes.config_map import ConfigMap
 from testsuite.kubernetes.deployment import Deployment, ContainerResources, ConfigMapVolume, VolumeMount
 from testsuite.kubernetes.service import Service, ServicePort
-from testsuite.utils.constants import (
-    MOCKSERVER_INTERNAL_PORT,
-    HTTP_API_PORT,
-    SERVICE_READY_TIMEOUT,
-    MOCKSERVER_READINESS_INITIAL_DELAY,
-    MOCKSERVER_READINESS_PERIOD,
-)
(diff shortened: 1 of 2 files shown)


silvi-t (approved):
LGTM

PR #1016, opened 2026-07-14
fix(egress): fix reruns on credential injection tests

averevki:
## Summary 

Fix the reruns configuration, added in https://github.com/Kuadrant/testsuite/pull/1015, on the egress credential injection tests.

Also added an assert for a response object, in case of a request timeout.

## Summary by CodeRabbit

* **Tests**
  * Improved credential-injection test validation with clearer checks for HTTP responses and authorisation headers.
  * Added automatic retries for assertion-related failures to improve test reliability.
  * Expanded coverage for authorised, unauthorised, invalid-token, missing-token, and destination-specific requests.


silvi-t (approved):
LGTM

PR #1017, opened 2026-07-14
chore: Remove HPA test for gateway and metrics apiserver

azgabur:
## Description
Work on https://redhat.atlassian.net/browse/CONNLINK-1301
To make it easier to use multiple tools deployment on same cluster as well as tools on remote cluster this test must be removed. It is a duplicate of `testsuite/tests/singlecluster/gateway/scaling/test_manual_scale_gateway.py` only using HPA instead of manual scaling. In my opinion both test test the same behavior in which we are interested. Second test is testing more Kubernetes HPA than the scaled Gateway behavior. The first test is sufficient in that.

## Changes
- Removal of custom api server code and test using which uses the custom HPA.
- Unifiing test case between `test_manual_scale_gateway.py` and removed `test_auto_scale_gateway.py`
- Better coupling of variables for LIMIT constant

## Verification
Needs single reviewer, preferably @crstrn13 who added this test
```sh
make testsuite/tests/singlecluster/gateway/scaling
```


crstrn13 (approved):
LGTM! 🎉 :shipit:

PR #1029, opened 2026-08-03
fix(oidc): oidc hostname requires namespace when used on non-OCP clusters

averevki:
### Summary

On OpenShift, Routes get a hostname in the `{name}-{namespace}.{apps_url}` format, so passing just the domain_name to the exposer produces the correct match. On EKS/Kind clusters, the LoadBalancer exposer doesn't add namespace name to the hostname, so we need to set it manually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated gateway hostname handling to use the correct project namespace.
  * Added provider-specific hostname support for OpenShift, EKS and Kind environments.
  * Improved OIDC policy test setup across supported cluster configurations.

### DIFF
--- a/testsuite/tests/singlecluster/extensions/oidc_policy/conftest.py
+++ b/testsuite/tests/singlecluster/extensions/oidc_policy/conftest.py
@@ -8,7 +8,8 @@ in their respective test files.
 from contextlib import contextmanager
 import pytest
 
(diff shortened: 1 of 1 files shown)


silvi-t (approved):
LGTM. I only ran the OIDC policy tests on an OCP cluster since I don't have an AWS one on hand, but it didn't break anything on OCP.

PR #1031, opened 2026-08-04
fix(extensions): add 5s wait for PipelinePolicy

silvi-t:
## Mitigate PipelinePolicy test flakiness in nightly runs
  
  PipelinePolicy tests intermittently fail in nightly CI with `500 Internal Server Error` instead of the expected status code. The root cause is that the `Enforced` status condition is reported prematurely — the kuadrant control plane marks the policy as enforced without waiting for the WASM shim to confirm that the EnvoyFilter has actually been applied. This means the gateway may not yet be ready to handle requests through the pipeline when tests start sending traffic.

  Add a 5-second wait after `wait_for_ready()` in `PipelinePolicy` to give the EnvoyFilter time to be actually enforced on the data plane.

## Summary by CodeRabbit

* **Tests**
  * Improved extension policy readiness checks by allowing time for enforcement changes to propagate before validation.
  * Increased reliability of automated tests that verify policy behaviour on the data plane.

### DIFF
--- a/testsuite/kuadrant/extensions/pipeline_policy.py
+++ b/testsuite/kuadrant/extensions/pipeline_policy.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import time
 from functools import cached_property
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


trepel (approved):
Haven't actually tried but it looks reasonable. I think that there are more flaky tests in our test suite failing due to same/similar reasons.

LGTM!

PR #1036, opened 2026-08-07
fix: exclude extensions from make kuadrant

silvi-t:
## Summary
  - Exclude extension tests from the `make kuadrant` target by adding `not extensions` marker filter
  - Extension tests remain runnable via the dedicated `make extensions` target
  - Temporary measure until the extension framework is stabilized
  
- Added `extensions` marker to `trlp_metrics` tests since they depend on `TelemetryPolicy` from the extension framework

## Summary by CodeRabbit

* **Tests**
  * Updated the standard test run to exclude extension-specific tests, alongside standalone, disruptive, and UI tests.
  * Existing parallel test execution and test distribution remain unchanged.


trepel (approved):
LGTM

